PR #234, opened 2022-03-11
Pre-pack asset configs (again)

jflatow:
Saves ~1.5KB contract size, which we're likely going to need


jflatow:
I think we're past this, so closing, but we can always find it in history if we want to come back

PR #251, opened 2022-03-18
Cap utilization at 1 for the supply rate scalar

jflatow:
This makes the supply rate linear after util = 1, which guarantees it won't surpass borrow rate and eat into reserves, or eat into reserve rate at all, which the quadratic form does.


kevincheng96:
Closing since this is no longer applicable (we don't need to cap utilization at 100% and are also now using a different interest rate model)

PR #344, opened 2022-05-11
Split out supply and borrow rate models into separate kink models

kevincheng96:
This PR implements a new interest rate model where the supply IR curve is no longer dependent on the borrow IR curve. This new IR model also no longer has a fixed reserve rate.

I'll only merge this after we are done with our interest rate research.


contracts/Comet.sol
```diff
@@ -163,8 +173,6 @@ contract Comet is CometMainInterface {
         if (config.assetConfigs.length > MAX_ASSETS) revert TooManyAssets();
         if (config.baseMinForRewards == 0) revert BadMinimum();
         if (AggregatorV3Interface(config.baseTokenPriceFeed).decimals() != PRICE_FEED_DECIMALS) revert BadDecimals();
-        if (config.reserveRate > FACTOR_SCALE) revert BadReserveRate();
-        if (config.kink > FACTOR_SCALE) revert BadKink();
```
kevincheng96, line 167: I removed this check because we theoretically have an IR model with a kink above 100%.

test/interest-rate-test.ts
```diff
@@ -157,11 +109,166 @@ describe('interest rates', function () {
...
     assertInterestRatesMatch(0, supplyRate.mul(SECONDS_PER_YEAR));
     // interestRateBase + interestRateSlopeLow * utilization
-    // = 0.005 + 0.1 * 0 = 0.005
-    assertInterestRatesMatch(exp(5, 15), borrowRate.mul(SECONDS_PER_YEAR));
+    // = 0.01 + 0.05 * 0 = 0.01
+    assertInterestRatesMatch(exp(0.01, 18), borrowRate.mul(SECONDS_PER_YEAR));
   });
 });
+
+// === Tests for old kink model with fixed reserve rate (v2 model) ===
```
jflatow, line 121: We want to keep this here?

jflatow (commented):
So many params now, but lgtm

test/interest-rate-test.ts
kevincheng96: I'm wondering if we should store both versions of Comet/Configurator in the repo, in case we ever want to switch between the two rate models. Though, it does seem difficult managing multiple versions of Comet/Configurator.

jflatow: We can always use git?

jflatow: Maybe a good time to drop a tag

jflatow: So we have a few impl contracts lying around and the one Comet extends from at any given time is the one we use? Wonder if this would get hard to manage and we should always just think about variant Comets using git commits/history, since they can vary in other ways besides IR models too

kevincheng96: Actually, the dependencies are the other way around. I was thinking something like this:

```
Contracts in repo:
CometOldModel.sol: CometOldModel extends CometBase
CometNewModel.sol: CometNewModel extends CometBase
CometXModel.sol: CometXModel extends CometBase
CometBase.sol: Basically an abstract Comet.sol but with the interest rate functions defined as abstract 
```

That way, the diffs among these contracts are only the interest model implementations. It'll make it easier to maintain different types of interest rate models IMO. But yeah, just thinking out loud here 🧠

kevincheng96: Removed the commented-out tests for the old IR model.

jflatow (approved):
Great work on this and all the analysis around the potential models! LGTM

PR #345, opened 2022-05-12
Rebased version of packing asset configs

kevincheng96:


jflatow:
Attaching to #234 and closing

PR #419, opened 2022-06-17
[L02] Address small gas inefficiencies

toninorair:
I measured gas differences for all of the suggested fixes. The only gas difference is added by an `unchecked` clause for the loop index. Everything else like moving items' length into a separate variable and using `++i` vs `i++` doesn't affect the number of gas units. My assumption is the compiler is smart enough to optimize for those cases.

### DIFF
--- a/contracts/Comet.sol
+++ b/contracts/Comet.sol
@@ -1115,7 +1115,6 @@ contract Comet is CometMainInterface {
 
         uint startGas = gasleft();
         accrueInternal();
-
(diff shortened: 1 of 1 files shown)


avalan4e57:
A nitpick from me concerning ++i vs. i++. Since you’re already modifying this part of the code, why not make it explicitly correct rather than relying on the compiler’s implicit optimization? This impacts readability because you can’t be certain the optimization happens just by looking at the code. Plus, we can’t guarantee that a different compiler would handle it the same way.

PR #450, opened 2022-06-29
Configurator check for baseToken instead of governor

jflatow:
As proposed by OpenZeppelin, since `governor` is mutable in configuration. For correctness, we should either do this or update docs that `baseToken` and `trackingIndexScale` are technically mutable. If we check for `governor`, then governance could reset `governor` to `address(0)` and then overwrite the entire config, which would include `baseToken` and `trackingIndexScale`.


kevincheng96 (approved):

kevincheng96 (commented):
Technically, `baseToken` is still mutable if it was initially set as `address(0)` 😛

PR #451, opened 2022-06-30
Basic simulation for interest rate model equilibrium

kevincheng96:
This PR implements a simple and bare-bones simulation (detailed in [RFC 036](https://docs.google.com/document/d/1tguCRfd2EfEdezg7TgCivARCUe_FaxhAeLhC2vb3tDE/edit#heading=h.madcppce2zyp)) for our interest rate models. The goal is to use some basic heuristics to determine the market equilibrium utilization, interest rates, and total supply/borrow size for an interest rate model. The simulation will also output scatterplots to show the equilibrium states of each model.

This simulation branch is currently rebased on top of #344, which introduces a new interest model with independently set borrow and supply curves.

The simulation can be run via a script using the command `yarn simulation`. 

**Scatterplots for simulation results:**
Rob's model with supply rate slope low adjusted:
Supply Rate Slope Low Adjustments

Rob's model with borrow rate slope low adjusted:
Borrow Rate Slope Low Adjustments

Custom models I've been messing with (Note: `0.035 S/0.045 B` stands for 0.035 supply slope low and 0.045 borrow slope low):
Custom Models

### DIFF
--- a/simulation/Chart.ts
+++ b/simulation/Chart.ts
@@ -9,11 +9,11 @@ const GREEN = 'rgba(194, 213, 100, 0.85)';
 
 export interface SimulationResults {
   [label: string]: {
-    utilization: number,
-    totalSupply: number,
-    annualProfit: number,
-  }
-};
+    utilization: number;
+    totalSupply: number;
+    annualProfit: number;
+  };
+}
 
 function getProfitibilityColor(profitability: number, neutralRange: number) {
   if (profitability > neutralRange) {
@@ -28,7 +28,7 @@ function getProfitibilityColor(profitability: number, neutralRange: number) {
 function createData(simResults: SimulationResults) {
   const labels = [];
   const dataObject = { data: [], pointRadius: 5, backgroundColor: [] };
-  Object.keys(simResults).map((label, i) => {
+  Object.keys(simResults).map(label => {
     labels.push(label);
     dataObject.data.push({
       x: simResults[label].totalSupply,
@@ -63,8 +63,7 @@ export async function createChart(simResults: SimulationResults, title?: string)
         legend: {
           position: 'right',
           labels: {
-            generateLabels: function (chart) {
-              var data = chart.data;
+            generateLabels: function (_chart) {
               return [{
                 text: 'Profitable',
                 fillStyle: GREEN,
@@ -80,7 +79,7 @@ export async function createChart(simResults: SimulationResults, title?: string)
         },
         datalabels: {
           align: 'top',
-          formatter: function (value, context) {
+          formatter: function (_value, context) {
             return context.chart.data.labels[context.dataIndex];
           },
           font: {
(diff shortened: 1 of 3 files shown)


jflatow (commented):
This is cool! Will look more closely in a bit

test/simulations/run-simulation.ts
```diff
@@ -0,0 +1,70 @@
+import { exp, makeProtocol, wait } from '../helpers';
+import { Supplier } from './actors/Supplier';
+import { Actor } from './actors/Actor';
+import { Market } from './Market';
+import { simulate } from './Simulate.ts';
+import { World } from './World';
+import { Borrower } from './actors/Borrower';
+
+// XXX doesn't have to be under the unit test directory. can be a standalone script
```
jflatow, line 9: Yeah I guess that probably makes more sense, and we could still import helpers

jflatow (commented):
Awesome 🙌  Gonna try to fit in the frequency / expected rate thing I was describing

test/simulations/run-simulation.ts
kevincheng96: Moved this to be its own script/directory.

simulation/Chart.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+async function fileExists(path: string): Promise<boolean> {
+  try {
+    await fs.stat(path);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
```
jflatow, line 107: Very cool! Want to upload a png? Not sure about introducing this dep though, its failing to install for me :(

jflatow: Maybe add to `optionalDependencies`?

kevincheng96: Sounds good. I can make the dependencies optional.

Sure, I'll upload the pngs (they are excluded by `.gitignore` right now). It's still a WIP because I want to also visualize how profitable (or at least if profitible) each of these points are.

kevincheng96:
Still working on visualizations for this PR so going to make it a draft again.

simulation/Chart.ts
kevincheng96: You should be able to view the charts now.

jflatow: Hm I can't run yarn anymore, which doesn't seem to be related to this branch necessarily (something to do with `hardhat-cover` actually, so I'll debug that separately). I still can't install these optionalDependencies (out of the box at least), but you are saying I can view them bc you added them to the repo? I guess that's cool (and thanks), but I was just thinking upload them to the PR. You think we should store them in the repo? Might not be a bad idea if not everyone can run these sims easily but also large-ish binary files and could get outdated

kevincheng96: Oh yeah that's smarter 🤦.  I wasn't planning on merging the images, but committed them temporarily so you can view them. Let me just upload them directly to the PR instead.

jflatow: Ok fixed the other issue using `npm i`, but still can't compile with these optional dependencies since the typescript compilation fails on `import { ChartConfiguration } from 'chart.js';`. So hesitant to merge, maybe we can find a different way of plotting that doesn't require native deps to be setup correctly so this can be portable

kevincheng96:
I've added some scatterplots to the original post of this PR.

PR #452, opened 2022-06-30
Let ERC20 interface use principal instead of present value

jflatow:


jflatow:
These (this and #453) seem like they'll increase confusion for users (due to the ERC-20 functions using principal and other functions using present value [which is more natural for thinking about the base asset]), although might be worse for composability. For instance, we wouldn't be able to use 📈ETH as collateral in 📈USDC without a wrapper, which feels weird, but also doesn't make a ton of sense as a real use case. OTOH a wrapper to make constant balances (e.g. like wstETH) and add EIP-4626 support could achieve this pretty easily, although that would mean an additional contract deployment per instance. Would love to get feedback from @transmissions11 on this, but closing for now since activity/interest seems to have died down.

sambacha:
> These (this and #453) seem like they'll increase confusion for users (due to the ERC-20 functions using principal and other functions using present value [which is more natural for thinking about the base asset]), although might be worse for composability. For instance, we wouldn't be able to use 📈ETH as collateral in 📈USDC without a wrapper, which feels weird, but also doesn't make a ton of sense as a real use case. OTOH a wrapper to make constant balances (e.g. like wstETH) and add EIP-4626 support could achieve this pretty easily, although that would mean an additional contract deployment per instance. Would love to get feedback from @transmissions11 on this, but closing for now since activity/interest seems to have died down.

Would a new Wrapped Ether contract be of any benefit for the case of using ETH in a wrapped form?

jflatow:
> > These (this and #453) seem like they'll increase confusion for users (due to the ERC-20 functions using principal and other functions using present value [which is more natural for thinking about the base asset]), although might be worse for composability. For instance, we wouldn't be able to use 📈ETH as collateral in 📈USDC without a wrapper, which feels weird, but also doesn't make a ton of sense as a real use case. OTOH a wrapper to make constant balances (e.g. like wstETH) and add EIP-4626 support could achieve this pretty easily, although that would mean an additional contract deployment per instance. Would love to get feedback from @transmissions11 on this, but closing for now since activity/interest seems to have died down.
> 
> Would a new Wrapped Ether contract be of any benefit for the case of using ETH in a wrapped form?

What did you have in mind? This is not about the WETH collateral but about the Comet ERC20 itself

sambacha:
> > > These (this and #453) seem like they'll increase confusion for users (due to the ERC-20 functions using principal and other functions using present value [which is more natural for thinking about the base asset]), although might be worse for composability. For instance, we wouldn't be able to use 📈ETH as collateral in 📈USDC without a wrapper, which feels weird, but also doesn't make a ton of sense as a real use case. OTOH a wrapper to make constant balances (e.g. like wstETH) and add EIP-4626 support could achieve this pretty easily, although that would mean an additional contract deployment per instance. Would love to get feedback from @transmissions11 on this, but closing for now since activity/interest seems to have died down.
> > 
> > 
> > Would a new Wrapped Ether contract be of any benefit for the case of using ETH in a wrapped form?
> 
> What did you have in mind? This is not about the WETH collateral but about the Comet ERC20 itself

Still reading this codebase ser, will get back to you now that I know that is really the matter, thx

PR #453, opened 2022-06-30
Alternative version of ERC20 interface using principal

jflatow:


jflatow:
See #452

jflatow:
We are reconsidering using principal for ERC20 functions, and tending towards preferring this alternative at this point. Re-opening so we can further develop this branch.

contracts/Comet.sol
```diff
@@ -882,23 +882,26 @@ contract Comet is CometMainInterface {
...
      * @return true
      */
-    function transferFrom(address src, address dst, uint amount) override external returns (bool) {
-        transferInternal(msg.sender, src, dst, baseToken, amount);
+    function transferFrom(address src, address dst, uint amount) override public returns (bool) {
+        if (isTransferPaused()) revert Paused();
+        if (!hasPermission(src, msg.sender)) revert Unauthorized();
+        if (src == dst) revert NoSelfTransfer();
+
+        transferBase(src, dst, safe104(amount));
```
kevincheng96, line 904: We need to convert `amount` to present value before using it as an argument here.

kevincheng96, line 904: I think we can avoid duplicating these checks by calling `transferInternal` directly.

contracts/Comet.sol
```diff
@@ -1267,31 +1270,50 @@ contract Comet is CometMainInterface {
...
     function totalSupply() override external view returns (uint256) {
         (uint64 baseSupplyIndex_, ) = accruedInterestIndices(getNowInternal() - lastAccrualTime);
         return presentValueSupply(baseSupplyIndex_, totalSupplyBase);
     }
 
     /**
-     * @notice Get the total amount of debt
+     * @notice Get the total present value of base in circulation
+     * @dev Note: uses updated interest indices to calculate
+     * @return The present value of tokens
```
kevincheng96, line 1283: nit: `present value of supply`?

kevincheng96 (commented):
Nice! Left some comments related to the correctness of the ERC20 `transferFrom` implementation. 

Otherwise, I think the interfaces look good. It is slightly weird that the ERC20 transfer and Comet transfer functions use different units for `amount`, but I don't think it's really that bad since we can just think of the ERC20 functions as a completely separate interface.

scott-silver:
Looking good!

Looks like there are some tests that need to be updated. Might be worth adding some additional tests to confirm that we're getting the values we'd expect.

contracts/Comet.sol
jflatow: Hm weird, I think I did it this way to specifically handle the `amount` being different, but its clearly wrong

contracts/Comet.sol
jflatow: Adding `supply` to the comment, but wdyt about the one for `totalSupply` then?

contracts/Comet.sol
jflatow: I guess I was thinking `amount` in `transferBase` was principal for some reason, since its not, agree its probably better to go through `transferInternal` even if its a little gassier. Also slightly weird that you can specify a varying amount of principal to end up at `max`, but I suppose harmless

jflatow:
This also isn't handling `Transfer` events appropriately...this change starts to look pretty crazy and creates a bunch of inconsistencies

jflatow:
This also creates another inconsistency in the amounts being specified to transfer, which should technically not be mapped to present supply value, and so we should technically add a separate `transferBasePrincipal` function, but which breaks other core design assumptions (including around the `Transfer` event mentioned above)

jflatow:
@transmissions11 there's a bunch of issues we see here trying to take this further, sorry to say I think we want to abandon it. Happy to discuss possible solutions, but at this point I think we suggest a wrapper for non-rebalancing ERC20 / EIP-4626.

PR #455, opened 2022-07-01
Fix overflow bug that occasionally occurs when supplying 0 base

kevincheng96:
This PR fixes a math overflow bug that occasionally occurs when supplying 0 base. I also added a unit test to verify the fix, as well as another unit test to show a weird quirk that can occur when withdrawing 0 base. 

**Bug details:**
In `supplyBase`, we calculate `dstPrincipalNew` using `dstPrincipal` (old principal value before amount is supplied). The formula is:

`int104 dstPrincipalNew = principalValue(presentValue(dstPrincipal) + signed104(amount))`

When `amount=0`, the formula condenses to:

`dstPrincipalNew = principalValue(presentValue(dstPrincipal))`

Due to the fact that both `principalValue` and `presentValue` round down in favor of the protocol, we can actually end up with a value of `dstPrincipalNew < dstPrincipal`. 

This breaks our assumption in `repayAndSupplyAmount` that `newPrincipal >= oldPrincipal` MUST be true. In the old code, this would cause `supplyAmount` returned from `repayAndSupplyAmount` to be an extremely large number (`uint104(-1)` to be precise), which would later cause an overflow during an addition operation. The new code now explicitly checks this assumption and sets both `repayAmount` and `supplyAmount` to 0 if the assumption is violated. We apply a similar check in `withdrawAndBorrowAmount` as well.

### DIFF
--- a/contracts/Comet.sol
+++ b/contracts/Comet.sol
@@ -591,7 +591,7 @@ contract Comet is CometMainInterface {
      * @dev The change in principal broken into repay and supply amounts
      */
     function repayAndSupplyAmount(int104 oldPrincipal, int104 newPrincipal) internal pure returns (uint104, uint104) {
-        // If the new principal is less than the old principal, then no amount has been repaid or supplied.
+        // If the new principal is less than the old principal, then no amount has been repaid or supplied
         if (newPrincipal < oldPrincipal) return (0, 0);
 
         if (newPrincipal <= 0) {
@@ -607,7 +607,7 @@ contract Comet is CometMainInterface {
      * @dev The change in principal broken into withdraw and borrow amounts
      */
     function withdrawAndBorrowAmount(int104 oldPrincipal, int104 newPrincipal) internal pure returns (uint104, uint104) {
-        // If the new principal is greater than the old principal, then no amount has been withdrawn or borrowed.
+        // If the new principal is greater than the old principal, then no amount has been withdrawn or borrowed
         if (newPrincipal > oldPrincipal) return (0, 0);
 
         if (newPrincipal >= 0) {
--- a/contracts/Comet.sol
+++ b/contracts/Comet.sol
@@ -592,6 +592,9 @@ contract Comet is CometMainInterface {
      * @dev Note: The assumption `newPrincipal >= oldPrincipal` MUST be true
      */
     function repayAndSupplyAmount(int104 oldPrincipal, int104 newPrincipal) internal pure returns (uint104, uint104) {
+        // If the new principal is less than the old principal, then no amount has been repaid or supplied.
+        if (newPrincipal < oldPrincipal) return (0, 0);
+
         if (newPrincipal <= 0) {
             return (uint104(newPrincipal - oldPrincipal), 0);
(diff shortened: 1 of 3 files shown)


test/withdraw-test.ts
```diff
@@ -193,6 +193,46 @@ describe('withdrawTo', function () {
...
+    });
+    // Weird quirk of round down behavior where `withdrawAmount` is 1 even though
+    // `amount` is 0. So no base leaves Comet (which is expected)
+    expect(event(s0, 2)).to.be.deep.equal({
+      Transfer: {
+        from: alice.address,
+        to: ethers.constants.AddressZero,
+        amount: 1n,
+      }
+    });
```
kevincheng96, line 233: Just wanted to highlight this weird interaction that can occasionally occur when withdrawing 0 base.

contracts/Comet.sol
```diff
@@ -592,6 +592,9 @@ contract Comet is CometMainInterface {
      * @dev Note: The assumption `newPrincipal >= oldPrincipal` MUST be true
```
jflatow, line 592: I think we can remove this now, or move the comment below to a dev note?

contracts/Comet.sol
```diff
@@ -606,6 +609,9 @@ contract Comet is CometMainInterface {
      * @dev Note: The assumption `oldPrincipal >= newPrincipal` MUST be true
```
jflatow, line 609: Same as above

contracts/Comet.sol
kevincheng96: Done.

contracts/Comet.sol
kevincheng96: Done.

contracts/Comet.sol
```diff
@@ -589,9 +589,11 @@ contract Comet is CometMainInterface {
 
     /**
      * @dev The change in principal broken into repay and supply amounts
-     * @dev Note: The assumption `newPrincipal >= oldPrincipal` MUST be true
      */
     function repayAndSupplyAmount(int104 oldPrincipal, int104 newPrincipal) internal pure returns (uint104, uint104) {
+        // If the new principal is less than the old principal, then no amount has been repaid or supplied.
```
jflatow, line 594: unbelievable nit but are these the only full-sentence comments in the contract?

jflatow (approved):
💪 🥇 🏆 🎣

contracts/Comet.sol
jflatow: actually they aren't. I'll clean it up in a separate PR (if at all)

kevincheng96: Sure thing, I removed the `.` here

scott-silver (commented):
🙌

test/supply-test.ts
```diff
@@ -193,6 +193,45 @@ describe('supplyTo', function () {
     });
   });
 
+  // This is an edge-case that can occur when a user supplies 0 base.
```
scott-silver, line 196: really good explanation

antonina-cherednichenko (approved):
great work

PR #456, opened 2022-07-07
Liquidation Bot

scott-silver:
Opening up our primary Liquidation Bot branch as a draft PR, so we can start opening PRs against this branch and start tracking feedback.

TODO:

- [x] move liquidation-bot vendor contracts into a sub directory
- [x] use Vendoza to manage vendor contracts?
- [x] add documentation to README
- [x] ~~write a deploy_liquidator script for Fuji~~ (Uniswap isn't on Fuji)
- [x] ~~add assets and pool fees to Kovan deploy_liquidator migration~~ (there won't be any Uniswap pools for our Comet assets
- [x] ~~re-run the Kovan deploy_liquidator migration; update roots.json with new address~~
- [x] update `run-liquidation-bot` script to work ~~across different networks~~ on mainnet
- [x] update Liquidator.sol license to MIT


contracts/Comet.sol
```diff
@@ -1323,4 +1323,4 @@ contract Comet is CometMainInterface {
             default { return(0, returndatasize()) }
         }
     }
-}
+}
```
jflatow, line 1326: This is the change to `Comet.sol`?

contracts/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,206 @@
...
+        address[] memory assets = decoded.assets;
+
+        uint256 totalAmountOut = 0;
+        for (uint i = 0; i < assets.length; i++) {
+            address asset = assets[i];
+            uint256 baseAmount = decoded.baseAmounts[i];
+
+            if (baseAmount == 0) continue;
+
+            // XXX approve everything all at once?
```
jflatow, line 86: Why not?

contracts/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,206 @@
...
+            TransferHelper.safeApprove(comet.baseToken(), address(comet), baseAmount);
+
+            uint24 poolFee = getPoolFee(asset);
+
+            // XXX Replace 0 with more meaningful value here
```
jflatow, line 91: Should we just pass this in too?

contracts/Comet.sol
scott-silver: I think there was a vestigial `import hardhat/console.log;` in this file at one point

contracts/liquidator/Liquidator.sol
scott-silver: it's probably not necessary, since we're always going to revert if the transaction is not profitable.

contracts/liquidator/Liquidator.sol
scott-silver: no reason not to

test/liquidation/comp-abi.ts
```diff
@@ -0,0 +1,2 @@
+export default [];
```
scott-silver, line 1: delete this?

scripts/liquidation_bot/index.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  CometInterface,
+  Liquidator
+} from '../../build/types';
+import liquidateUnderwaterBorrowers from './liquidateUnderwaterBorrowers';
+
+const loopDelay = 5000;
+
+async function main() {
+  console.log('main called');
```
scott-silver, line 12: delete

test/liquidation/makeLiquidatableProtocol.ts
```diff
@@ -0,0 +1,205 @@
...
+    method: 'hardhat_impersonateAccount',
+    params: [USDC_WHALE],
+  });
+  let usdcWhaleSigner = await ethers.getSigner(USDC_WHALE);
+
+  // transfer DAI to Comet, so the protocol can sell them
+  await mockDai.connect(daiWhaleSigner).transfer(comet.address, 200000000000000000000n);
+  // transfer USDC to comet, so it has money to pay out withdraw to underwater user
+  await mockUSDC.connect(usdcWhaleSigner).transfer(comet.address, 300000000n); // 300e6
+  // transfer USDC to signer, so it has money to purchase collateral (is this still necessary?)
```
scott-silver, line 163: - [ ]  determine if this is necessary

contracts/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,206 @@
...
+    /// @param params The parameters necessary for flash and the callback, passed in as FlashParams
+    /// @notice Calls the pools flash function with data needed in `uniswapV3FlashCallback`
+    function initFlash(FlashParams memory params) external {
+        // Absorb Comet underwater accounts
+        comet.absorb(address(this), params.accounts);
+
+        (uint256 totalBaseAmount, uint256[] memory assetBaseAmounts, address[] memory cometAssets) = calculateTotalBaseAmount();
+
+        address poolToken0 = params.reversedPair ? comet.baseToken(): params.pairToken;
+        address poolToken1 = params.reversedPair ? params.pairToken : comet.baseToken();
```
kevincheng96, line 177: Feels a bit brittle for the client to pass in `reversedPair`. I took a look at the `PoolAddress` library and looks like we can just use [this logic](https://github.com/Uniswap/v3-periphery/blob/main/contracts/libraries/PoolAddress.sol#L20-L27) to create the `PoolKey`. You can also probably derive the value of `reversedPair` using the same logic (e.g. `if (pairToken > baseToken) reversedPair = true;`).

contracts/liquidator/Liquidator.sol
kevincheng96: Do you mean approve `uint256.MAX` instead of `baseAmount`?

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -0,0 +1,60 @@
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
+import {
+  CometInterface,
+  Liquidator
+} from '../../build/types';
+
+// XXX this is Kovan-specific
+const daiPool = {
+  tokenAddress: '0x6B175474E89094C44Da98b954EedeAC495271d0F',
+  poolFee: 500,
```
kevincheng96, line 10: Why not 100? Seems like that's the most liquid pool: https://info.uniswap.org/#/tokens/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -0,0 +1,60 @@
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
+import {
+  CometInterface,
+  Liquidator
+} from '../../build/types';
+
+// XXX this is Kovan-specific
```
kevincheng96, line 7: The token address seems to be DAI on mainnet.

kevincheng96 (commented):
This looks great overall!

contracts/liquidator/Liquidator.sol
jflatow: Yeah token order should be reliable

contracts/liquidator/Liquidator.sol
jflatow: Oh is that what you meant? I guess its a bit more fragile if we rely on all the tokens having the approve max behavior

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
jflatow: I think they've tried to use the same deployer/nonce across networks as much as possible

contracts/liquidator/Liquidator.sol
kevincheng96: Wouldn't setting the approve amount to `type(uint256).max` pass for all ERC20 tokens? It could also save some extra gas.

test/liquidation/comp-abi.ts
scott-silver: nvm, being fleshed-out in #461

contracts/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,206 @@
...
+        uint256 amount;
+        address payer;
+        PoolAddress.PoolKey poolKey;
+        address[] assets;
+        uint256[] baseAmounts;
+        bool reversedPair;
+    }
+
+    using LowGasSafeMath for uint256;
+    using LowGasSafeMath for int256;
```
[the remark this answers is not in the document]
scott-silver: #463

contracts/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,206 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity 0.8.15;
+pragma abicoder v2;
```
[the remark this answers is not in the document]
scott-silver: good call; removed

contracts/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,206 @@
...
+                quoteCollateral = amount of DAI you get for 1 USDC
+                collateralBalance = Comet's balance of DAI
+                price = amount of USDC required to buy all DAI
+
+                1 / quotePrice = x / collateralBalance
+                (1 / quotePrice) * collateralBalance = x
+            */
+            // uint256 assetBaseAmount = comet.collateralBalanceOf(address(comet), asset) * quotePrice / 1e30; // PRICE_SCALE + 1e12
+            // console.log("assetBaseAmount: %s", assetBaseAmount);
+            uint256 assetBaseAmount = ((1e6 * 1e18 / quotePrice) * collateralBalance) / 1e18;
```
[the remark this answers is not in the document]
antonina-cherednichenko: Fixed in one of PRs

contracts/liquidator/Liquidator.sol
antonina-cherednichenko: Interesting point, I will try it out.

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
antonina-cherednichenko: because here we added support for DAI base asset and basically we want to borrow from one DAI pool with a fee of 500 and then swap to USDC using the second pool with a 100 fee. Still good point. Now thinking about it, I will swap those 2 pools because Comet won't support DAI.

contracts/liquidator/Liquidator.sol
antonina-cherednichenko: I added it in the latest merged pr

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
antonina-cherednichenko: Changed to 100 and swap asset pool to 500 instead

contracts/liquidator/Liquidator.sol
```diff
@@ -1,4 +1,4 @@
-//SPDX-License-Identifier: Unlicense
+// SPDX-License-Identifier: BUSL-1.1
```
jflatow, line 1: Hmm, I wonder if we actually want this to be the same license as everything else 🤔

contracts/liquidator/Liquidator.sol
```diff
@@ -12,7 +12,12 @@ import "./vendor/@uniswap/v3-periphery/contracts/interfaces/ISwapRouter.sol";
 import "../CometInterface.sol";
 import "../ERC20.sol";
 
+/**
+ * @title Compound's Liquidator Contract
+ * @author Compound
+ */
```
jflatow, line 18: Maybe "Compound's Example Liquidator Contract" and possibly a note that this is meant for others to build upon?

contracts/liquidator/Liquidator.sol
jflatow: 🤖 suggesting MIT

scott-silver: updated

contracts/liquidator/Liquidator.sol
```diff
@@ -28,114 +35,212 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
+        for (uint8 i = 0; i < poolConfigs.length; i++) {
+            UniswapPoolConfig memory poolConfig = poolConfigs[i];
+            if (poolConfig.asset == asset) {
+                return poolConfig;
+            }
         }
+
+        // If asset is not found, proceed with default pool config
+        return UniswapPoolConfig({
+            fee: 500,
```
kevincheng96, line 109: I think this is where you would want to use `DEFAULT_POOL_FEE`.

kevincheng96, line 109: What happens if a 500 fee pool does not exist for this asset? Does the whole txn revert?

contracts/liquidator/Liquidator.sol
```diff
@@ -28,114 +35,212 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
+            asset: asset,
+            isLowLiquidity: false
+        });
     }
 
-    function getPoolFee(address asset) internal view returns(uint24) {
-        uint24 poolFee = poolFees[asset];
-        return poolFee == 0 ? defaultPoolFee : poolFee;
+    /**
+     * @dev Swaps the given asset to USDC(base token) using Uniswap pools
```
kevincheng96, line 116: nit: missing space between `USDC` and parenthesis

contracts/liquidator/Liquidator.sol
```diff
@@ -28,114 +35,212 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
         uint256 fee1,
         bytes calldata data
     ) external override {
+        // Verify uniswap callback, recommended security measure
         FlashCallbackData memory decoded = abi.decode(data, (FlashCallbackData));
         CallbackValidation.verifyCallback(factory, decoded.poolKey);
 
         address[] memory assets = decoded.assets;
 
+        // Allow Comet protocol to withdraw USDC(base token) for collateral purchase
```
kevincheng96, line 186: nit: missing space after `USDC`

contracts/liquidator/Liquidator.sol
```diff
@@ -28,114 +35,212 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
-                        amountOutMinimum: baseAmount,
-                        sqrtPriceLimitX96: 0
-                    })
-                );
+            uint256 amountOut = swapCollateral(asset);
             totalAmountOut += amountOut;
         }
 
-        uint256 fee = decoded.reversedPair? fee0 : fee1;
+        // We borrow only 1 asset, so one of fees will be 0
```
kevincheng96, line 202: Cool!

contracts/liquidator/Liquidator.sol
```diff
@@ -28,114 +35,212 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
+        uint256 fee = fee0 + fee1;
+        // Payback flashloan to Uniswap pool and profit to the caller
         payback(decoded.amount, fee, comet.baseToken(), totalAmountOut, decoded.payer);
     }
 
+    /**
+     * @dev Returns loan to Uniswap pool and sends USDC (base token) profit to caller
+     * @param amount The loan amount that need to be repaid
+     * @param fee The fee for taking the loan
+     * @param token The base token whch was borrowed for successful liquidation
```
kevincheng96, line 212: nit: typo

contracts/liquidator/Liquidator.sol
```diff
@@ -28,114 +35,212 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
+     * @param amountOut The total amount of base token received after liquidation
+     * @param payer The caller address of liquidation bot
+     */
     function payback(
         uint256 amount,
         uint256 fee,
         address token,
         uint256 amountOut,
         address payer
```
kevincheng96, line 221: Would it be more accurate to call this `recipient`, since it's the receiving address of the funds (paid by `address(this)`)?

contracts/liquidator/Liquidator.sol
```diff
@@ -145,37 +250,41 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
-            // console.log("assetBaseAmount: %s", assetBaseAmount);
-            uint256 assetBaseAmount = ((1e6 * 1e18 / quotePrice) * collateralBalance) / 1e18;
+
+            if (collateralBalance == 0) continue;
+
+            // Find the price in asset needed to base QUOTE_PRICE_SCALE of USDC(base token) of collateral
+            uint256 quotePrice = comet.quoteCollateral(asset, QUOTE_PRICE_SCALE * comet.baseScale());
+            uint256 assetBaseAmount = comet.baseScale() * QUOTE_PRICE_SCALE * collateralBalance / quotePrice;
+
+            // Liquidate only positions with adequate gains, no need to collect residue from protocol
```
kevincheng96, line 260: nit: this actually checks the size of the liquidation, but not the gains

kevincheng96 (commented):
💪

contracts/liquidator/Liquidator.sol
antonina-cherednichenko: Sounds good, I will update it

contracts/liquidator/Liquidator.sol
antonina-cherednichenko: great catch, I will update that.

antonina-cherednichenko: yes, it reverts. But it happens only if config for an asset is not given.

README.md
```diff
@@ -202,6 +202,8 @@ For more information, see [SCENARIO.md](./SCENARIO.md).
 9. run `1653512186_seed_rewards_with_comp`, `prepare` step
 10. run `1653512186_seed_rewards_with_comp`, `enact` step
 11. execute the proposal from step 10
+12. run `1657124918_deploy_liquidator`, `prepare` step (optional)
+13. update `deployments/kovan/root.json` with liquidator root from step 12 (optional)
```
jflatow, line 206: Hm, I think we should discuss this -- not sure about making liquidator part of normal deployment

scott-silver: going to delete these steps (and the `deploy_liquidator` migration).

it doesn't actually make sense to run the liquidator on Kovan since there are no liquidity pools for the cloned assets that the Kovan version of Comet uses

contracts/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,314 @@
...
+        uint24 fee;
+    }
+
+    /** Liquidator configuration constants **/
+
+    /// @notice The scale for asset price calculations
+    uint256 public constant QUOTE_PRICE_SCALE = 1e6;
+
+    /// @notice Uniswap standard pool fee, used if custom fee is not specified for the pool
+    uint24 public constant DEFAULT_POOL_FEE = 500;
```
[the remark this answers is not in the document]
scott-silver: updated; it's now in use on line 115

contracts/liquidator/Liquidator.sol
scott-silver: updated in #474

contracts/liquidator/Liquidator.sol
scott-silver: updated in #474

contracts/liquidator/Liquidator.sol
scott-silver: comment updated in #474

README.md
scott-silver: deleted

contracts/Comet.sol
jflatow: Should we just get rid of the diff on this one?

scott-silver: sure; added to #480

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
scott-silver: removed this comment

scripts/liquidation_bot/liquidator-abi.ts
```diff
@@ -0,0 +1,232 @@
...
+    'name': 'unwrapWETH9',
+    'outputs': [],
+    'stateMutability': 'payable',
+    'type': 'function'
+  },
+  {
+    'stateMutability': 'payable',
+    'type': 'receive'
+  }
+];
```
jflatow, line 232: Curious why we needed to include all these abis here like this and they can't be generated with the normal hardhat compilation?

jflatow (commented):
This looks like a great starting point for the community! One last (I think) question about the abis we are including

scripts/liquidation_bot/liquidator-abi.ts
scott-silver: yep, good call. we can use `getContractAt`.

#481

jflatow: what about all the other abis? I guess those aren't easy to access from here

scott-silver: `getContractAt` uses the name ("Liquidator") to grab the ABI from the hardhat artifact.

The other ABIs are for tokens (COMP, LINK, UNI, etc). We don't have the contract code for those tokens locally, so I don't think we can use `getContractAt` to grab their ABIs.

I did some experimenting with using the local ERC20 interface for the ABI. So I replaced this:

```
const mockDai = new ethers.Contract(DAI, daiAbi, signer);
```

...with this:

```
const mockDai = await ethers.getContractAt("ERC20", DAI, signer) as ERC20;
```

This has some unintended consequences, though. It changes the return types of some of the methods. And I'm assuming it means you would be unable to call any non-ERC20 method on that contract.

It seems inaccurate, too. It assumes that all of the tokens have implemented the ERC20 interface in the exact same way.

jflatow (approved):
I think we should merge, maybe @ajb413 can add the note about running on `mainnet` in the guide PR?

kevincheng96 (approved):
💪

PR #465, opened 2022-07-11
Cleanups

antonina-cherednichenko:
1. Move 'approve' base amount up vs approving for each asset
2. Additional check in swap
3. Comments


scott-silver (approved):
🧼  🧽  ✨

PR #467, opened 2022-07-12
Toni/add min check and fixes

antonina-cherednichenko:
Add pool configs structures and check for minimum supported liquidation amount (set to 10USDC) in `makeLiquidatableProtocol`

### DIFF
--- a/contracts/liquidator/Liquidator.sol
+++ b/contracts/liquidator/Liquidator.sol
@@ -48,6 +48,9 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
     /// @notice Compound Comet protocol
     CometInterface public immutable comet;
 
+    /// @notice Minimum available amount for liquidation in USDC (base token)
+    uint256 public immutable liquidationThreshold;
+
     /// @notice The address of WETH asset
     address public immutable weth;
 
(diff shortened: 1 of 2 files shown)


contracts/liquidator/Liquidator.sol
```diff
@@ -61,41 +72,41 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
+        uint256 _liquidationThreshold,
+        UniswapPoolConfig[] memory _poolConfigs
     ) PeripheryImmutableState(_factory, _WETH9) {
-        require(_assets.length == _poolFees.length, "Wrong input");
-        require(_assets.length == _lowLiquidity.length, "Wrong input");
-
         swapRouter = _swapRouter;
         comet = _comet;
         weth = _WETH9;
+        liquidationThreshold = _liquidationThreshold;
```
scott-silver, line 88: what's the advantage of setting a minimum? (as opposed to just letting the transaction proceed as long as its profitable)

antonina-cherednichenko: running complex logic for basically no gains, saving gas this way

scott-silver: personally, i'd let the caller decide whether it's worth it to them to call. but i don't feel strongly about it

scott-silver (approved):
👍

contracts/liquidator/Liquidator.sol
antonina-cherednichenko: but they cannot decide in our contract. There is some residue left in the protocol, and there will attempt to buy it. I agree, it has its pros and cons.

PR #471, opened 2022-07-13
Certora's formal verification for comet

MichaelMorami:


MichaelMorami:
@jflatow keep in mind that certora's git actions fail because the Certora key is not set as repository secret.

jflatow:
> @jflatow keep in mind that certora's git actions fail because the Certora key is not set as repository secret.

I think the secrets actually need to be set in your repo for this to succeed on the PR. Surprised that you don't have `CERTORAKEY` set yourself though, since you've been testing there?

MichaelMorami:
> > @jflatow keep in mind that certora's git actions fail because the Certora key is not set as repository secret.
> 
> I think the secrets actually need to be set in your repo for this to succeed on the PR. Surprised that you don't have `CERTORAKEY` set yourself though, since you've been testing there?

PR #476, opened 2022-07-15
Support deploying specific contracts instead of all at once

kevincheng96:
This PR improves our deployment scripts in a few ways:

- Consolidate all deployment options into `DeployCometOptionalParams` so only a single object for optional params needs to be passed into `deployComet`
- Introduce a new `ContractsToDeploy` optional param that allows for more specificity in which contracts to deploy. This should put us in the right direction for mainnet and L2 migrations that don't need certain contracts (e.g. `SimpleTimelock` and `GovernorSimple`)

`ContractsToDeploy` usage:
If `ContractsToDeploy` is not provided as an optional param, all contracts will be deployed. The interface also has an `all` property that explicitly states to deploy all contracts. The `all` property can be combined with other properties to deploy "all but some" contracts. e.g.

```
{ all: true, timelock: false }
```

will deploy all contracts except for the `Timelock`.

### DIFF
--- a/src/deploy/Network.ts
+++ b/src/deploy/Network.ts
@@ -27,7 +27,7 @@ import { ExtConfigurationStruct } from '../../build/types/CometExt';
 
 import { ContractsToDeploy, DeployedContracts, ProtocolConfiguration } from './index';
 import { getConfiguration } from './NetworkConfiguration';
-import { extractCalldata, fastGovernanceExecute } from '../utils';
+import { extractCalldata, fastGovernanceExecute, shouldDeploy } from '../utils';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
 
 export async function deployNetworkComet(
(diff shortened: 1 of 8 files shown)


jflatow (commented):
Yay, fantastic idea/initiative 👏

src/deploy/Network.ts
```diff
@@ -25,34 +25,54 @@ import {
...
 export async function deployNetworkComet(
   deploymentManager: DeploymentManager,
-  deployProxy: DeployProxyOption = { deployCometProxy: true, deployConfiguratorProxy: true },
-  configurationOverrides: ProtocolConfiguration = {},
+  contractsToDeploy?: ContractsToDeploy,
+  configurationOverrides?: ProtocolConfiguration,
   contractMapOverride?: ContractMap,
   adminSigner?: SignerWithAddress,
 ): Promise<DeployedContracts> {
+  if (contractsToDeploy == null) {
```
scott-silver, line 40: you could also accomplish these lines using default parameters:

```
export async function deployNetworkComet(
  deploymentManager: DeploymentManager,
  contractsToDeploy?: ContractsToDeploy = { all: true },
  configurationOverrides?: ProtocolConfiguration = {},
  contractMapOverride?: ContractMap,
  adminSigner?: SignerWithAddress,
)
```

PR #477, opened 2022-07-15
Migration scripts to upgrade interest rate model on testnet

kevincheng96:
This PR adds migration scripts to upgrade the interest rate model on kovan and fuji. The migration script partially re-deploys contracts (only the Configurator proxy, implementation, and CometFactory need to be re-deployed).

I opted for deploying a new Configurator proxy instead of re-using the old one so we can modify the storage layout without having to keep old storage variables around and having to append new storage variables at the end. Rob said this was okay to do since we are still on testnet.

Migrations run:

**Kovan**:
- [x] 1. Prepare migration
 - https://github.com/compound-finance/comet/actions/runs/2693702996
- [x] 2. Enact migration
 - https://github.com/compound-finance/comet/actions/runs/2693862371
- [x] 3. Execute proposal
 - https://kovan.etherscan.io/tx/0x8d858e0b31a99aac89a2081bc4057fc0194a14555be6594c0f14dc872258ed4b

**Fuji**:
- [x] 1. Prepare migration
 - https://github.com/compound-finance/comet/actions/runs/2693704894
- [x] 2. Enact migration
 - https://github.com/compound-finance/comet/actions/runs/2693786849
- [x] 3. Execute proposal
 - https://testnet.snowtrace.io/tx/0x41e189384cba4646b4c9ce86bbbcc39747435c81e0fdcccc72a46b28cfc35e9a


scott-silver:
I like the optionality you've added with the `kevin/improve-deployment-scripts` branch.

If you were deploying to testnets from scratch today, would you end up deploying the Configurator twice (once in the initial deploy migration and once with this migration)?

kevincheng96:
> If you were deploying to testnets from scratch today, would you end up deploying the Configurator twice (once in the initial deploy migration and once with this migration)?

If you were to deploy from scratch today, you would skip this deployment because the first deployment script would already deploy the newest interest rate model to testnet.

scott-silver:
ok cool, that makes sense. so this migration is really only meant to alter the existing testnet deploy (which was deployed with the previous IR model).

is that an argument for running this migration but not merging it into `main`?

jflatow:
Cool, let's review all this today though since I think there was intentional value in trying to upgrade Configurator while still on testnet, but not sure where we're at with everything esp since we already merged the IR model changes

kevincheng96:
> is that an argument for running this migration but not merging it into `main`?

Curious what your thoughts are on this since my personal opinion is that migrations should always be merged if they are run on-chain.

Let's say you have a migration on mainnet that updates the `Comet` implementation. That migration is similar to this one in that they are both ad hoc and won't be run again if we did a fresh re-deploy. I'm leaning towards merging that migration because it changes the on-chain state and would be nice to track. Otherwise, we would have a bunch of unmerged migrations, and that feels inconsistent to me.

scott-silver (approved):
sounds like we're going to spend some more time thinking about how we manage migrations.

but this looks good to me for the time-being

jflatow (approved):
Should we merge this and make our future/discussed changes off `main`?

kevincheng96:
> Should we merge this and make our future/discussed changes off `main`?

Sure, let's do that.

PR #479, opened 2022-07-18
delete deploy_liquidator migration

scott-silver:
deleting, since it doesn't actually make sense to deploy the Liquidator on Kovan (since there are no liquidity pools for our cloned assets)


antonina-cherednichenko (commented):
should were be a migration fo mainnet instead ?

PR #484, opened 2022-07-21
Add option to lazily verify contracts

kevincheng96:
One contributing factor to the flakiness of our deployment scripts is that each contract is verified right as it is deployed, rather than after all of the contracts have been deployed. Since contract verification happens over network requests to etherscan, they can sometimes fail (e.g. API throttling, normal network request failures) and require the entire deployment script to be re-ran from the start, wasting a lot of gas and time.

In an attempt to reduce the overall flakiness of our deployment scripts, this PR introduces an option to lazily verify contracts. Instead of verifying each contract the moment they are deployed, the deployment manager will cache the verification arguments and allow all contracts to be verified at the end of the deployment using the `verifyContracts` function.


deployments/fuji/migrations/1644432723_deploy_fuji.ts
```diff
@@ -82,6 +84,9 @@ migration('1644432723_deploy_fuji', {
       contracts
     );
 
+    // Verify contracts after all contracts have been deployed
+    await deploymentManager.verifyContracts();
```
jflatow, line 88: Awesome feature, but this interface feels a little awkward...you have to set the option then you call verify at the end too, which doesn't seem to check the option?

kevincheng96: Yeah I can spend more time thinking about the interface. It's a bit clunky because the deployment manager doesn't know when it should verify contracts, so we always have to manually call it at some point.

plugins/deployment_manager/Deploy.ts
```diff
@@ -21,6 +22,7 @@ export interface DeployOpts {
   overwrite?: boolean; // should we overwrite existing contract link
   connect?: Signer; // signer for the returned contract
   verify?: boolean; // verify contract on etherscan
+  lazyVerify?: boolean; // delay verification: can prevent flakiness of deployments
```
kevincheng96, line 25: I'm thinking of consolidating these two options into one with an enum value instead. Feels kind of weird that you can specify both `verify` and `lazyVerify` right now. Instead, I think it's a nicer interface if you specify either `VerifyOption.EAGER`, `VerifyOption.LAZY`, `VerifyOption.NONE`.

jflatow: Agree the 2 options are part of awkwardness and the states you can get into with it are weird now

src/deploy/Network.ts
```diff
@@ -225,7 +225,6 @@ export async function deployNetworkComet(
   }
 
   await deploymentManager.putRoots(updatedRoots);
-  await deploymentManager.spider();
```
scott-silver, line 228: 👍

deployments/fuji/migrations/1644432723_deploy_fuji.ts
kevincheng96: Added a `doThenVerify` function to `DeploymentManager` that should improve the interface a bit.

plugins/deployment_manager/Deploy.ts
kevincheng96: Combined these options into a union type `VerificationStrategy`.

plugins/deployment_manager/Deploy.ts
```diff
@@ -20,7 +21,7 @@ export interface DeployOpts {
   name?: string; // name for aliasing
   overwrite?: boolean; // should we overwrite existing contract link
   connect?: Signer; // signer for the returned contract
-  verify?: boolean; // verify contract on etherscan
+  verificationStrategy?: VerificationStrategy; // strategy for verifying contracts on etherscan
```
jflatow, line 24: I guess maybe my question now is, why do we ever need an eager strategy?

kevincheng96: Good question. Mainly just preserving the status quo, but also maybe if want to avoid writing/reading to cache for various use cases (smaller deploys, scenarios).

jflatow: Still not convinced we need anything but lazy, but this all looks very clean 🧼 I think we should merge, but also, if the complexity of supporting multiple verification strategies leads to any problems or difficulties, I think we should consider only supporting lazy

jflatow (approved):
Looks great 🤩 But left my $.02 about only supporting lazy in the future if there's any reason to do that

plugins/deployment_manager/Deploy.ts
kevincheng96: Definitely, shouldn't be too hard to only support lazy if we ever want to in the future

PR #488, opened 2022-07-25
Remove the use of fastGovernanceExecute in our deploys

kevincheng96:
This PR removes the need to use a `fastGovernanceExecute` (propose and execute changes via governance; only doable on testnets) in our deploy scripts, which makes them more compatible for mainnet deploys.


mykelp (approved):
Nice cleanup!

jflatow (approved):
👏

PR #489, opened 2022-07-26
Fault tolerant deploys

kevincheng96:
This PR makes our deploys more consistent by handling network failures using retries and timeouts (to handle hanging txns). A new function `asyncCallWithRetry` is added to the `DeploymentManager` for this purpose.

I opted to implement the retry/timeout logic at the function level. The other approach would be to wrap an entire deployment script around one retry/timeout handler, but this requires us to cache which txns were already executed, which is difficult to generically do. The trade-off with my approach is that we have to explicitly wrap ad hoc (non-deploy) txns such as `usdc.initialize()` with an `asyncCallWithRetry` function in the deployment script.

### DIFF
--- a/deployments/kovan/migrations/1644388553_deploy_kovan.ts
+++ b/deployments/kovan/migrations/1644388553_deploy_kovan.ts
@@ -9,6 +9,7 @@ import {
   ProxyAdmin__factory
 } from '../../../build/types';
 import { Contract } from 'ethers';
+import { debug } from '../../../plugins/deployment_manager/Utils';
 
 let cloneNetwork = 'mainnet';
 let cloneAddr = {
@@ -51,19 +52,25 @@ migration('1644388553_deploy_kovan', {
       cloneNetwork
     );
 
-    await wait(await usdcProxy.changeAdmin(usdcProxyAdmin.address));
+    debug(`Changing admin of USDC proxy to ${usdcProxyAdmin.address}`);
+    await deploymentManager.asyncCallWithRetry(
+      (signer_) => wait(usdcProxy.connect(signer_).changeAdmin(usdcProxyAdmin.address))
+    )
     usdc = usdcImplementation.attach(usdcProxy.address);
     // Give signer 10,000 USDC
-    await wait(
-      usdc.initialize(
-        'USD Coin',
-        'USDC',
-        'USD',
-        6,
-        signerAddress,
-        signerAddress,
-        signerAddress,
-        signerAddress
+    debug(`Initializing USDC`);
+    await deploymentManager.asyncCallWithRetry(
+      (signer_) => wait(
+        usdc.connect(signer_).initialize(
+          'USD Coin',
+          'USDC',
+          'USD',
+          6,
+          signerAddress,
+          signerAddress,
+          signerAddress,
+          signerAddress
+        )
       )
     );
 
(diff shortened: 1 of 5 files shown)


plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -240,9 +247,31 @@ export class DeploymentManager {
...
+  async asyncCallWithRetry(fn: (signer: SignerWithAddress) => Promise<any>, retries: number = 5, timeLimit?: number) {
+    const signer = await this.getSigner();
+    try {
+      return await asyncCallWithTimeout(fn(signer), timeLimit);
+    } catch (e) {
+      retries -= 1;
+      debug(`Retrying with retries left: ${retries}`);
+      debug(`Error is ${e}`);
+      if (retries === 0) throw e;
+      // XXX to be extra safe, we can also get the signer transaction count and figure out the next nonce
```
kevincheng96, line 266: This hasn't happened recently, but in the past, a hanging txn might actually get mined twice. This is worth investigating, but I haven't been able to replicate the issue with the most recent code.

deployments/kovan/migrations/1644388553_deploy_kovan.ts
```diff
@@ -51,19 +52,25 @@ migration('1644388553_deploy_kovan', {
       cloneNetwork
     );
 
-    await wait(await usdcProxy.changeAdmin(usdcProxyAdmin.address));
+    debug(`Changing admin of USDC proxy to ${usdcProxyAdmin.address}`);
```
kevincheng96, line 55: I've added more debug logs to make it even clearer which txns are being run and retried.

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -240,9 +247,31 @@ export class DeploymentManager {
...
+      this._signers = [];
+      return await this.asyncCallWithRetry(fn, retries);
```
scott-silver, line 268: should this include `timeLimit`?

```
this.asyncCallWithRetry(fn, retries, timeLimit)`
```

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -240,9 +247,31 @@ export class DeploymentManager {
     return await this.cache.readCache(getArtifactSpec(migration));
   }
 
-  async clone<C extends Contract>(address: string, args: any[], network?: string): Promise<C> {
+  /**
+   * Call an async function with a given amount of retries
+   * @param fn an async function that takes a signer as an argument. The function takes a signer
```
scott-silver, line 252: > a new instance of a signer needs to be used on each retry

how come?

also, `this.getSigner()` will return the same instance of the signer, right?

scott-silver: oh, i see; you're resetting `this._signers` to force it to generate new instances

plugins/deployment_manager/DeploymentManager.ts
kevincheng96: Good catch! Will add that

plugins/deployment_manager/DeploymentManager.ts
kevincheng96: An ethers signer wraps around a provider, and we need to use a new provider to submit a retry txn when the old txn is hanging. Otherwise, the retry txn will just hang again using the old provider.

deployments/kovan/migrations/1644388553_deploy_kovan.ts
```diff
@@ -51,19 +52,25 @@ migration('1644388553_deploy_kovan', {
...
+        usdc.connect(signer_).initialize(
+          'USD Coin',
+          'USDC',
+          'USD',
+          6,
+          signerAddress,
+          signerAddress,
+          signerAddress,
+          signerAddress
+        )
```
jflatow, line 73: Very nice 👏

plugins/deployment_manager/DeploymentManager.ts
jflatow: Ok good call out, something to pay attn to

jflatow (approved):
Nice improvement 💪

scott-silver (approved):
🥳

PR #491, opened 2022-07-26
Taming migration scripts

jflatow:
We are working towards a simplified/standard workflow which we still need to document:

 1. Open a PR with a migration script
 2. Run 'prepare' workflow in GitHub CI
  - this should deploy any necessary contracts
  - changes to roots should auto-commit to the PR branch
  - maybe go post in the forum about it for review
 3. Run 'enact' workflow in GitHub CI or elsewhere
  - this should make any necessary gov proposals
 4. Once the proposal is accepted by gov, merge PR with new roots

NB: currently `roots.json` changes *can* happen in migrations. In the next iteration of this, we will have a separate workflow for the (single) 'deploy' script, which should be the only thing modifying roots. Changes which happen in migrations should be automatically spidered in when any related proposal passes. Otherwise, a new root should be added and it should be part of 'deploy'.

### DIFF
--- a/plugins/deployment_manager/Migration.ts
+++ b/plugins/deployment_manager/Migration.ts
@@ -1,73 +1,38 @@
-import fg from 'fast-glob';
-import * as path from 'path';
 import { DeploymentManager } from './DeploymentManager';
 import { FileSpec } from './Cache';
 
-interface Action<T> {
+interface Actions<T> {
   prepare: (dm: DeploymentManager) => Promise<T>;
   enact: (dm: DeploymentManager, t: T) => Promise<void>;
-  enacted: (dm: DeploymentManager) => Promise<boolean>;
 }
 
-export interface Migration<T> {
+export class Migration<T> {
   name: string;
-  actions: Action<T>;
-}
+  actions: Actions<T>;
 
-export class Loader<T> {
-  migrations: { [name: string]: Migration<T> };
-
-  constructor() {
-    this.migrations = {};
-  }
-
-  addMigration(name: string, actions: Action<T>) {
-    if (this.migrations[name]) {
-      throw new Error(`Duplicate migration by name: ${name}`);
-    }
-    this.migrations[name] = {
-      name,
-      actions
-    };
-  }
-
-  getMigrations(): { [name: string]: Migration<T> } {
-    return this.migrations;
+  constructor(name: string, actions: Actions<T>) {
+    this.name = name;
+    this.actions = actions;
   }
 }
 
-export let loader: any;
-
-export function setupLoader<T>() {
-  loader = new Loader<T>();
+export async function loadMigration(path: string): Promise<Migration<any>> {
+  const { default: thing } = await import(path);
+  if (!(thing instanceof Migration))
+    throw new Error(`Does not export a valid default Migration`);
+  return thing;
 }
 
-export function getLoader<T>(): Loader<T> {
-  if (!loader) {
-    throw new Error('Loader not initialized');
+export async function loadMigrations(paths: string[]): Promise<Migration<any>[]> {
+  const migrations = [];
+  for (const path of paths) {
+    migrations.push(await loadMigration(path));
   }
-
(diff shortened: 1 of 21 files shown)


plugins/deployment_manager/Migration.ts
```diff
@@ -50,16 +48,12 @@ export function getLoader<T>(): Loader<T> {
...
-export async function loadMigrations<T>(glob: string): Promise<{ [name: string]: Migration<T> }> {
+export async function loadMigrations<T>(paths: string[]): Promise<{ [name: string]: Migration<T> }> {
   setupLoader<T>();
 
-  const entries = await fg(glob); // Grab all potential migration files
-
-  for (let entry of entries) {
-    let entryPath = path.join(process.cwd(), entry);
-
+  for (let path of paths) {
```
hayesgm, line 54: Does this change make it harder to run all?

scenario/constraints/MigrationConstraint.ts
```diff
@@ -1,17 +1,18 @@
...
-// TODO: Improvements
-function getMigrationConfigs(requirements: Requirements): boolean {
-  return requirements.migrate ?? false;
+import { execSync } from 'child_process';
+import { existsSync } from 'fs';
+
+async function getMigrations<T>(context: CometContext, requirements: Requirements): Promise<Migration<T>[]> {
+  // TODO: make this configurable from cli params/env var?
+  const deployment = context.deploymentManager.deployment; // XXX should become per instance
+  const output = execSync(`git diff --numstat main | grep 'deployments/${deployment}/{deploys,changes}/.*.ts' | awk '{ print $3 }'`);
```
hayesgm, line 13: I wonder if there's a native JavaScript git library as opposed to running exec here?

hayesgm (commented):
looking good

scenario/constraints/MigrationConstraint.ts
jflatow: There's probably `libgit2` bindings but its another native dependency and API to figure out and not sure I see the gain right now

hayesgm: yeah nodegit is too native, there is isogit and a complex process to get a diff without any native system calls: https://isomorphic-git.org/docs/en/snippets#git-diff-name-status-commithash1-commithash2

plugins/deployment_manager/Migration.ts
jflatow: Caller can always do `glob`, right? But also I think this is relatively unimportant now, since you never really want to run all, except maybe for deploys, but we'll have 1 (*maybe* 2) of those per deployment, and at worst you would just have to press the button twice (no commit to make in between)

scenario/constraints/MigrationConstraint.ts
jflatow: `statusMatrix` seems closest to what we'd want, but would be a bunch of effort to match what its doing now fairly trivially, so will leave it as a future exercise

hayesgm: Agreed.

scenario/CometScenario.ts
```diff
@@ -1,7 +1,7 @@
 import { CometProperties, scenario } from './context/CometContext';
 import { expect } from 'chai';
 
-scenario('initializes governor correctly', { migrate: true }, async ({ comet, timelock, actors }, world) => {
```
kevincheng96, line 4: Does this currently fail with MigrationConstraint? We could also just skip the scenario for now

kevincheng96 (approved):
Great changes 💪

scenario/CometScenario.ts
jflatow: I don't think so? But I changed `MigrationConstraint` to always add solutions for any 'modified' migration scripts

jflatow: So in other words no, this just isn't a requirement for `MigrationConstraint` anymore

PR #492, opened 2022-07-27
Rework Comet to export roots.json files for Webb3

scott-silver:
See https://github.com/compound-finance/webb3/pull/80


package.json
```diff
@@ -23,7 +26,6 @@
   },
   "scripts": {
     "build": "hardhat compile",
-    "prepare": "hardhat compile && tsc",
```
scott-silver, line 26: The `prepare` step gets called automatically when you install Comet as a dependency. However, the `hardhat compile` command requires all of the environment variables that we require in `hardhat.config.ts`.

As far as I can tell, we're not actually using this command anywhere. Removing it allows us to include Comet as a dependency in Webb3

package.json
```diff
@@ -1,14 +1,17 @@
...
-    "dist"
+    "dist",
+    "deployments/**/roots.json"
   ],
   "exports": {
-    ".": "./dist/index.js",
-    "./*": "./dist/*.js"
+    ".":  "./dist/index.js",
+    "./*": "./dist/*.js",
+    "./kovan/roots.json": "./deployments/kovan/roots.json",
```
scott-silver, line 13: Annoyingly, you cannot specify a glob for a directory; this does not work:

```
"./**/roots.json": "./deployments/**/roots.json"
```

scott-silver: We might be able to reduce some of this duplication by exporting the entire `deployments/` directory, instead of each roots file

jflatow (approved):
Nice 👏

kevincheng96 (approved):

PR #493, opened 2022-07-28
Simplify deploy scripts into one

kevincheng96:
This PR simplifies deploy scripts by moving them into one script per network (per deployment in the future).

Summary of changes in this PR:

- Combined all deploys (including bulker and rewards) into one script
- Combined some migrations (mint to fauceteer) into the deploy script
- Moved the update roots logic in deployment scripts to the `prepare` step from the `enact` step
- Removed unnecessary ERC20 addresses in `roots.json`.
- Add deploying rewards contract to the `deployComet` library functions


deployments/kovan/migrations/1644388553_deploy_kovan.ts
```diff
@@ -22,135 +24,227 @@ let cloneAddr = {
...
+    { all: true },
+    {},
+    contracts
+  );
+
+  // Deploy Bulker
+  const bulker = await deploymentManager.deploy<Bulker, Bulker__factory, [string, string, string]>(
+    'Bulker.sol',
+    [timelock.address, cometProxy.address, weth.address]
+  );
```
jflatow, line 174: I guess bulker is the main thing stopping the interface being 'return the args to `deployComet`, so I guess I'll start with interface being return `Roots` (for separate `deploy` that is)

deployments/kovan/migrations/1653512186_seed_rewards_with_comp.ts
```diff
@@ -10,64 +11,65 @@ migration<Vars>('1653512186_seed_rewards_with_comp', {
...
+          transferCompCalldata,
+        ],
+        'Set RewardConfig and transfer COMP to CometRewards')
+      ).wait()
+    );
+
+    const event = txn.events.find(event => event.event === 'ProposalCreated');
+    const [proposalId] = event.args;
+
+    await deploymentManager.asyncCallWithRetry(
```
jflatow, line 57: I wonder if we should give this fn a friendly name like `do` since its going to be used everywhere

jflatow (approved):
👍 lgtm! Haven't looked at the failing scen though

deployments/kovan/migrations/1653512186_seed_rewards_with_comp.ts
kevincheng96: Yeah, `asyncCallWithRetry` is kind of long. I can do a refactor PR. I'm thinking `do`, `retry`, or `doRetry`.

kevincheng96:
> 👍 lgtm! Haven't looked at the failing scen though

The failing scenario was related to the UNI token being minted before the `mintingAllowedAfter` timestamp. We were hitting this error probably because moving the minting logic into the deployment script meant less time was elapsed between deploying UNI and minting UNI. I added a fixed delay between the deploying and minting steps to avoid hitting this in scenarios.

PR #494, opened 2022-07-28
Remove all unnecessary usages of `upgrade: true` in scenarios

kevincheng96:
This PR removes unnecessary usages of `upgrade: true` from our scenarios. Some scenarios still have valid reasons for using it (e.g. interest rate scenarios that use it to upgrade the interest rate params), so these are not fully removed from all scenarios.


jflatow (approved):
Love that we're fixing up scens🔥

PR #495, opened 2022-07-29
Fix scenario admin issue for chains without GovernorSimple

kevincheng96:
Right now, scenarios set the admin signer to be the address returned by `governor.admins(0)`. This works on testnets where the governor contract is `GovernorSimple`. However, this wouldn't for mainnet because `GovernorBravo` doesn't have an `admins` list to read from. 

This PR fixes the issue by setting the admin signer to be `comet.governor()`, which could either be an EOA or a contract address (`Timelock`), and using it to directly sign txns for contract upgrades. Since the admin could be a contract address, we have to set the next block's base fee to 0 and set the gas price to 0 whenever sending a txn with it.

### DIFF
--- a/scenario/ApproveThisScenario.ts
+++ b/scenario/ApproveThisScenario.ts
@@ -1,51 +1,35 @@
 import { scenario } from './context/CometContext';
 import { expect } from 'chai';
-import { constants, utils } from 'ethers';
+import { constants } from 'ethers';
+import { setNextBaseFeeToZero } from './utils';
 
 scenario('Comet#approveThis > allows governor to authorize and rescind authorization for Comet ERC20', {}, async ({ comet, timelock, actors }, world, context) => {
-  let approveThisCalldata = utils.defaultAbiCoder.encode(["address", "address", "uint256"], [timelock.address, comet.address, constants.MaxUint256]);
-  await context.fastGovernanceExecute(
-    [comet.address],
-    [0],
-    ["approveThis(address,address,uint256)"],
-    [approveThisCalldata]
-  );
+  const { admin } = actors;
+
+  await setNextBaseFeeToZero(world);
+  await admin.approveThis(timelock.address, comet.address, constants.MaxUint256, { gasPrice: 0 });
 
   expect(await comet.isAllowed(comet.address, timelock.address)).to.be.true;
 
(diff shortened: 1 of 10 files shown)


scenario/GovernanceScenario.ts
```diff
@@ -3,7 +3,7 @@ import { expect } from 'chai';
 import { constants, utils } from 'ethers';
 import { CometModifiedFactory, CometModifiedFactory__factory } from '../build/types';
 
-scenario('upgrade Comet implementation and initialize', { upgradeAll: true }, async ({ comet, configurator, proxyAdmin }, world, context) => {
+scenario.skip('upgrade Comet implementation and initialize', { upgradeAll: true }, async ({ comet, configurator, proxyAdmin }, world, context) => {
```
kevincheng96, line 6: I'm skipping these two specifically for now because I want these to test the governance flow. Planning on supporting execution of mainnet governance proposals in a separate PR.

jflatow, line 6: So you are skipping these bc they actually need gov? Want to `XXX` these to mark and so we know to expect the coming fix?

kevincheng96: Yep. Sure, added the XXX.

jflatow (approved):
Nice improvements 💪

scenario/ApproveThisScenario.ts
```diff
@@ -1,18 +1,15 @@
...
 import { expect } from 'chai';
 import { constants } from 'ethers';
-import { setNextBaseFeeToZero } from './utils';
 
 scenario('Comet#approveThis > allows governor to authorize and rescind authorization for Comet ERC20', {}, async ({ comet, timelock, actors }, world, context) => {
   const { admin } = actors;
 
-  await setNextBaseFeeToZero(world);
-  await admin.approveThis(timelock.address, comet.address, constants.MaxUint256, { gasPrice: 0 });
+  await admin.approveThis(timelock.address, comet.address, constants.MaxUint256, world);
```
jflatow, line 8: Damn going to increase my conflicts 😆 I'm moving world into context so we don't need to pass it around everywhere, since it's also error prone and strictly tied to it

kevincheng96: Doh! Should I revert this change? I can do the re-factor after your changes are in.

kevincheng96: Reverted the change.

PR #497, opened 2022-07-30
Standardize deploy scripts, and deploy once per base

jflatow:
Cleans up and refactors a bunch of the deploy code. We no longer have a separate code path for development deploys, and we make them more efficient by only deploying once per base instead of once per scenario. As a result, we also avoid redeploying all tokens during 'upgrades' in development.

Of note: we store the world in the context to avoid passing it around everywhere, and to prevent incorrect usage. In addition, the deploy logic now lives in the deployment manager, and is performed by scenario plugin (Worker), which gives the scenario plugin a stronger dependency on the deployment manager dependency. However these plugins were already coupled for other reasons, and there are likely other benefits that can come of this.


plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -132,6 +132,28 @@ export class DeploymentManager {
...
+  async deployMissing(force: boolean = true) {
+    // XXX if this is idempotent we can just always deploy, and do the same for dev
+    //  as is, won't handle cases where the deploy script adds roots or partial redeploys
+    // if force or there are no roots, deploy
+    //  force will also have to change with idempotent deploy changes
+    //   its here for deploy task, which doesn't really care if roots exists or not
+    //    but we'll want another way to specify how idempotent should work
+    const roots = await this.getRoots();
+    if (force || roots.size == 0) {
+      // XXX noted above but cache is a misnomer since we have non-cache files its kind of managing
```
hayesgm, line 145: Cache here has always been "in-memory or disk-synced filesystem-like thing." The name came because it was a write-through cache (sometimes), or just an in-memory KV store. The key idea is the calling code never needed to worry if it was a simulation (in-memory only) or disk-synced (e.g. for deployments, etc)

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -132,6 +132,28 @@ export class DeploymentManager {
...
+      //  could either rename or move that functionality
+      const deployScript = this.cache.getFilePath({ rel: 'deploy.ts' });
```
hayesgm, line 147: I think this is fine, do we do this with other types of files? I don't mind scens and DeploymentManager depending on one another, so long as they stay as plugins that _could_ be removed from Comet, which I think this preserves.

plugins/scenario/worker/Worker.ts
```diff
@@ -59,10 +59,10 @@ export async function run<T, U, R>({ bases, config, worker }: WorkerData) {
     scenarios = getLoader<T, U, R>().getScenarios();
   }
 
-  for (let base of bases) {
-    let world = new World(hreForBase(base), base);
-    let runner = new Runner({ base, world });
-    runners[base.name] = runner;
+  for (const base of bases) {
```
hayesgm, line 62: A constant value that changes every loop ;)

scenario/constraints/CometBalanceConstraint.ts
```diff
@@ -75,12 +75,12 @@ export class CometBalanceConstraint<T extends CometContext, R extends Requiremen
       // XXX ideally when properties fail
       //  we can report the names of the solution which were applied
       const solutions = [];
-      solutions.push(async function barelyMeet(context: T, world: World) {
+      solutions.push(async function barelyMeet(context: T) {
```
hayesgm, line 78: As much as I like having a world, I do think this makes life so much easier and doesn't really change semantics.

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -132,6 +132,28 @@ export class DeploymentManager {
     );
   }
 
+  /* Deploys missing contracts from the deployment, using the user-space deploy.ts script */
+  async deployMissing(force: boolean = true) {
```
hayesgm, line 136: The idempotency comments below refer to how to make this "missing" versus all?

hayesgm (commented):
Seems quite reasonable. Great work.

kevincheng96 (commented):
This looks great, can't wait to have it running successfully in CI!

Looking at CI errors for scenarios, seems like you still need to define a `deployments/development/deploy.ts` file.

scenario/constraints/CometBalanceConstraint.ts
jflatow: its still being passed, fwiw

jflatow: Like scen plugin doesn't assume that it will be in the context, but it *is* for comet, which I think was mainly what you wanted to preserve, maybe?

plugins/scenario/Scenario.ts
```diff
@@ -16,7 +16,8 @@ export interface Constraint<T, R> {
   check(requirements: R, context: T, world: World): Promise<void>;
 }
 
-export type Property<T, U> = (properties: U, world: World, context: T) => Promise<any>;
+export type Receipt = { cumulativeGasUsed: { toNumber: () => number } };
```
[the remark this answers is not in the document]
jflatow: See above note to Geoff, but its Comet that puts it in its own context, its not required by the plugin

plugins/scenario/World.ts
```diff
@@ -1,6 +1,9 @@
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
 
+// NB: this couples this plugin to deployment manager plugin
```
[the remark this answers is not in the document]
jflatow: nota bene 😬

src/deploy/index.ts
```diff
@@ -57,13 +52,6 @@ export interface ContractsToDeploy {
   rewards?: boolean;
 }
 
-interface DeployCometOptionalParams {
-  contractsToDeploy?: ContractsToDeploy;
-  configurationOverrides?: ProtocolConfiguration;
-  contractMapOverride?: ContractMap;
-  adminSigner?: SignerWithAddress;
-}
```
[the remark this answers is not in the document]
jflatow: Yeah, maybe, but pretty verbose, was hoping a lot of the current complexity would further evaporate with the next set of refactors

plugins/deployment_manager/DeploymentManager.ts
jflatow: Yea, pretty much

plugins/deployment_manager/DeploymentManager.ts
jflatow: I think its mainly `deploy.ts`, `configuration.json`, and `migrations` which scens currently need to know about

plugins/scenario/worker/Worker.ts
jflatow: 'twill not be reassigned

src/deploy/NetworkConfiguration.ts
```diff
@@ -100,110 +89,78 @@ interface TrackingInfo {
...
+    return { [k]: overrides[k] ?? f(config), ...acc };
+  }, {});
 }
 
 export async function getConfiguration(
-  network: string,
-  hre: HardhatRuntimeEnvironment,
-  contractMapOverride?: ContractMap
+  deploymentManager: DeploymentManager,
+  contractOverrides?: ContractMap,
```
kevincheng96, line 160: Thoughts on setting the type of `contractOverrides` to `Map<string, string>` (map from alias to address) instead of `Map<string, Contract>`? We only ever use the address portion of the contracts and it's quite troublesome to set these overrides in a mainnet deploy script (we have to construct a contract instance for each token).

Another approach is to allow optionally setting the token addresses directly in `configuration.json`, which makes sense as well.

jflatow: Makes sense to me

jflatow: Should we do this in a later PR?

kevincheng96: Yeah, I can make the change in a separate PR. Ran into this issue while testing mainnet scenarios.

scenario/constraints/CometBalanceConstraint.ts
hayesgm: The idea of world is that you could encapsulate side-effects in an object, but the side-effects mostly live in the chain state, wherein the World isn't as useful as some pure thing in a mostly impure world.

hayesgm (commented):
Looking good to me.

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -286,8 +286,6 @@ export class DeploymentManager {
       debug(`Retrying with retries left: ${retries}, wait: ${wait}`);
       debug('Error is: ', e);
       if (retries === 0) throw e;
-      // XXX to be extra safe, we can also get the signer transaction count and figure out the next nonce
-      this._signers = [];
```
kevincheng96, line 267: hmm this is needed for our deploy scripts to be able to retry hanging txns successfully. I guess I should have made the intention more explicit, but we clear out signers to force `getSigners()` to rebuild a new Signer with a new provider.

kevincheng96 (approved):
👍

scott-silver (approved):

PR #498, opened 2022-08-01
Spider multiple markets

scott-silver:
Updating Spider (and the things that depend on Spider) to support multiple deployments on the same chain.

Making this update means that we'll frequently be adding an additional `--market` argument to many of the commands that previously just took a `--network` argument.

```
# before
yarn spider --network kovan

# after
yarn spider --network kovan --market 01-USDC
```

Our scripts (like `deploy-comet` and `verify-comet`) will require market to be passed in via an environment variable (`MARKET="01-USDC"`):

```
MARKET="01-USDC" yarn hardhat run scripts/deploy-comet --network kovan
```

---

Going to make two changes discussed earlier with 👨‍🍳  and 🍯:
- [x] rename "market" to "deployment"
- [x] delete the number prefix ("01-USDC")


.github/workflows/run-spider.yaml
```diff
@@ -1,33 +0,0 @@
-name: 'Run Spider'
```
scott-silver, line 1: deleted, per your question @jflatow:

https://discord.com/channels/402910780124561410/796452409773129778/1002590359756480601

README.md
```diff
@@ -228,7 +228,7 @@ Uniswap for a profit.
 To run the bot, you'll need the address of a deployed version of the Liquidator
 contract (or you can deploy a new instance of it yourself):
 
-`LIQUIDATOR_ADDRESS="0xABC..." yarn liquidation-bot`
+`LIQUIDATOR_ADDRESS="0xABC..." MARKET="01-USDC" yarn liquidation-bot --network kovan`
```
scott-silver, line 231: As far as I can tell, the only way to pass arguments to Hardhat scripts (anything run via `hardhat run ...`) is to use environment variables.

https://hardhat.org/hardhat-runner/docs/advanced/scripts#hardhat-arguments

It's definitely clunky to have to pass in some arguments via env variables and some (like `--network`) via params

hardhat.config.ts
```diff
@@ -159,24 +159,31 @@ const config: HardhatUserConfig = {
...
+        "USDC-01": kovanUSDCRelationConfigMap,
+      }
     },
   },
 
   scenario: {
     bases: [
       {
         name: 'development',
+        market: '01-USDC' // ?
```
scott-silver, line 175: this is slightly awkward. I updated `ForkSpec` to require a `market` argument as well.

it seems like `market` should be required, but it means we need to add a market for the `development` base that will just be ignored.

plugins/deployment_manager/test/DeploymentManagerTest.ts
```diff
@@ -93,23 +93,22 @@ describe('DeploymentManager', () => {
...
-        'avalanche'
       );
       expect(importResult).to.eql(fiatTokenBuildFile);
     });
   });
 
   describe('deploy', () => {
     it('should deploy succesfully', async () => {
-      let deploymentManager = new DeploymentManager('test', hre, {
+      let deploymentManager = new DeploymentManager('test', 'XXX', hre, {
```
scott-silver, line 111: - [ ] update from `XXX` to `test-market`

plugins/deployment_manager/test/ImportTest.ts
```diff
@@ -47,7 +47,7 @@ export const fiatTokenBuildFile: BuildFile = {
   version: 'v0.6.12+commit.27d51765',
 };
 
-export function mockImportSuccess(hre: HardhatRuntimeEnvironment, address: string) {
+export function mockImportSuccess(address: string) {
```
scott-silver, line 50: removed the `hre` argument from this function since it wasn't being used

MIGRATIONS.md
```diff
@@ -8,10 +8,10 @@ Migrations are simple scripts which deploy or modify contracts. The goal of migr
 To create a new migration, run:
 
 ```sh
-yarn hardhat gen:migration --network kovan my_migration
+yarn hardhat gen:migration --network kovan --market 01-USDC my_migration
```
jflatow, line 11: I know we discussed this already, but given that these may be part of URLs etc, I think we should drop the prefix and only add a suffix to differentiate if/when needed

deployments/fuji/01-USDC/deploy.ts
```diff
@@ -1,15 +1,39 @@
+<<<<<<< HEAD:deployments/fuji/deploy.ts
```
scott-silver, line 1: - [ ] delete rebase artifacts

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -38,15 +35,17 @@ export class DeploymentManager {
...
     hre: HardhatRuntimeEnvironment,
     config: DeploymentManagerConfig = {}
   ) {
+    this.network = network;
     this.deployment = deployment;
     this.hre = hre;
     this.config = config;
 
-    this.cache = new Cache(deployment, config.writeCacheToDisk ?? false, config.baseDir);
+    this.cache = new Cache(`${this.network}/${this.deployment}`, config.writeCacheToDisk ?? false, config.baseDir);
```
scott-silver, line 48: this is slightly strange; previously you could derive the path to the roots.json file from the "deployment" alone.

the other option would be to update `Cache` to take in both network and deployment.

hayesgm (commented):
Awesome, looks good to me.

plugins/deployment_manager/DeploymentManager.ts
scott-silver: updated so that Cache takes a network and deployment

.github/workflows/run-spider.yaml
jflatow: I think its gone already on main?

.github/workflows/prepare-migration.yaml
```diff
@@ -42,13 +45,13 @@ jobs:
...
           ETH_PK: ${{ github.event.inputs.eth_pk }}
           DEBUG: true
 
       - uses: actions/upload-artifact@v2 # upload test results
         if: success() || failure() # run this step even if previous step failed
         with:
-          name: ${{ github.event.inputs.network }}-${{ github.event.inputs.migration }}
-          path: deployments/${{ github.event.inputs.network }}/artifacts/${{ github.event.inputs.migration }}.json
+          name: ${{ github.event.inputs.network }}-${{ github.event.inputs.deployment }}-${{ github.event.inputs.migration }}
+          path: deployments/${{ github.event.inputs.network }}/${{ github.event.inputs.deployment }}/artifacts/${{ github.event.inputs.migration }}.json
```
jflatow, line 57: Did you see the new `deploy-market` workflow?

hardhat.config.ts
```diff
@@ -158,24 +158,31 @@ const config: HardhatUserConfig = {
...
 
   scenario: {
     bases: [
       {
         name: 'development',
+        deployment: 'usdc'
       },
       {
         name: 'kovan',
+        deployment: 'usdc',
```
jflatow, line 178: Can we get rid of the network params now that we have the network name for the deployment?

.github/workflows/prepare-migration.yaml
scott-silver: missed that one. i'll update it

hardhat.config.ts
scott-silver: we'll still need it for scenarios, right?

scenarios will iterate over `bases`; for each base it finds, it will generate an hre using the `name` field, and then look for a `roots.json` file in `${name}/${deployment}/roots.json`.

seems like we should rename `name` to `network`

jflatow: Well but the network is configured as part of the deployment now, that was the whole point of including it in the deployment info (encoded in path currently, alternatively was `network.json`)? So `base` has a name so that we can refer to it from `--bases`, and it uses a certain deployment, but should be able to pull chain args from the deployment network, right?

scott-silver: So if you've got two deployments:

`/deployments/kovan/usdc`
`/deployments/fuji/usdc`

and your bases are:

```
[
  {
    deployment: 'usdc'
  },
  {
    deployment: 'usdc'
  }
]
```

how would you specify that you only wanted to run the scenarios for `fuji/usdc`?

```
yarn scenario --bases usdc
```

or are you saying we should have the deployment name be `fuji/usdc`?

scott-silver: I guess it really needs to be `fuji/usdc` and `kovan/usdc`; right now, we're limited to running scenarios on a single deployment per network

.github/workflows/prepare-migration.yaml
scott-silver: updated

hardhat.config.ts
jflatow: It's the base name not the deployment name! Technically both deployments are named `usdc`, the bases are named whatever you name them in the hardhat config

hardhat.config.ts
```diff
@@ -158,24 +158,31 @@ const config: HardhatUserConfig = {
...
+        usdc: kovanUSDCRelationConfigMap,
+      }
     },
   },
 
   scenario: {
     bases: [
       {
         name: 'development',
+        deployment: 'dai'
```
jflatow, line 174: Yay 🙌

jflatow (approved):
Great work 🥈 ! Can't wait to get this in, it's probably the last major reorganization we needed and should let the dependent repos (webb3, devops) build upon some solid ground 🏆

scott-silver:
okay, this should be done now.

@jflatow this is the commit that updates our scenario bases, per our discussion yesterday:

https://github.com/compound-finance/comet/pull/498/commits/c85e8366ab07a101d71a172e802b82a68457c060

it involves renaming the `deployments/development/` to `deployments/hardhat` to match the convention of `deployment/${network}/${deployment}`

jflatow:
> okay, this should be done now.
> 
> @jflatow this is the commit that updates our scenario bases, per our discussion yesterday:
> 
> [c85e836](https://github.com/compound-finance/comet/commit/c85e8366ab07a101d71a172e802b82a68457c060)
> 
> it involves renaming the `deployments/development/` to `deployments/hardhat` to match the convention of `deployment/${network}/${deployment}`

Ah nice but damn, was hoping we could keep it named `development`. Did you confirm scens are doing the right thing? Just thinking the network name might be `hardhat` for forked nodes to and since this didn't pop up before want to make sure its not doing something unexpected

scott-silver:
Was able to restore `deployments/development`: https://github.com/compound-finance/comet/pull/498/commits/d015dcadabaabf8308c234252972da8ff8249288

It's determined by what we pass in to the `DeploymentManager` as the `network` argument (which then determines the path to write/read assets, e.g. `deployments/development/dai/roots.json`)

We've been special-casing the base with the name "development" in a few places. The one downside to continuing with this approach (as opposed to the original `deployments/hardhat/` approach) is that now we can only have a single development base.

Previously, we could have added an additional development base:

```
[
  {
    name: "development-base-1",
    network: "hardhat",
    deployment: "dai"
  },
  {
    name: "development-base-2",
    network: "hardhat",
    deployment: "usdc"
  }
]
````

So basically: if we're cool with there only being one `development` base, then we can just special-case based on the base name.

File: plugins/scenario/World.ts [hunk 12]
```diff
@@ -22,13 +22,14 @@ export class World {
...
-    this.deploymentManager = new DeploymentManager(base.name, hre, { debug: true });
+    const networkName = base.name === 'development' ? 'development' : base.network;
+    this.deploymentManager = new DeploymentManager(networkName, base.deployment, hre, { debug: true });
     this.snapshotDeploymentManager = this.deploymentManager;
   }
 
   // TODO: Can we do this better?
   isRemoteFork(): boolean {
-    return !!this.base.url;
+    return this.base.network !== 'hardhat';
```
jflatow, line 32: If this is better (why) should we get rid of the TODO?

plugins/scenario/World.ts
```diff
@@ -22,13 +22,14 @@ export class World {
     // Q: should we really need to fork/snapshot the deployment manager?
     this.hre = hre;
     this.base = base;
-    this.deploymentManager = new DeploymentManager(base.name, hre, { debug: true });
+    const networkName = base.name === 'development' ? 'development' : base.network;
```
jflatow, line 25: `this.isRemoteFork() ? base.network : 'development';`?

plugins/scenario/World.ts
scott-silver: deleted

plugins/scenario/World.ts
scott-silver: that works. updated

plugins/scenario/World.ts
```diff
@@ -22,12 +22,11 @@ export class World {
     // Q: should we really need to fork/snapshot the deployment manager?
     this.hre = hre;
     this.base = base;
-    const networkName = base.name === 'development' ? 'development' : base.network;
+    const networkName = this.isRemoteFork() ? base.network : 'development';
```
jflatow, line 25: I thought we agreed in channel not to special case though?

scott-silver: deleted

plugins/scenario/World.ts
scott-silver: deleted; we settled on using `deployments/hardhat`

jflatow (approved):
🔥

PR #499, opened 2022-08-01
restore mnemonic

scott-silver:
If you run unit tests or scenarios on `main` with a value set for the environment variable `ETH_PK`, then nearly all tests will fail:

image

The address that can't be found is the one for the `pauseGuardian`, which is equal to `signers[1]`.

The reason this can't be found is because in the Hardhat config, if you have `ETH_PK` set, then we only generate a single signer. So the signer array only has one member, and `signers[1]` is undefined.

We can fix this by restoring the signer generation to use `MNEMONIC`. If we really need to set the balance for the generated signers, we can use the `accountsBalance` field: https://hardhat.org/hardhat-network/docs/reference#accounts

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -123,7 +123,7 @@ const config: HardhatUserConfig = {
       gas: 12000000,
       gasPrice: 'auto',
       blockGasLimit: 12000000,
-      accounts: ETH_PK ? [{ privateKey: ETH_PK, balance: (10n ** 36n).toString() }] : { mnemonic: MNEMONIC },
+      accounts: { mnemonic: MNEMONIC },
       // this should be default commented out and only enabled during dev to allow partial testing
       // XXX comment out by default once we've made the full contract fit
       allowUnlimitedContractSize: true,


hardhat.config.ts
```diff
@@ -123,7 +123,7 @@ const config: HardhatUserConfig = {
       gas: 12000000,
       gasPrice: 'auto',
       blockGasLimit: 12000000,
-      accounts: ETH_PK ? [{ privateKey: ETH_PK, balance: (10n ** 36n).toString() }] : { mnemonic: MNEMONIC },
+      accounts: { mnemonic: MNEMONIC },
```
jflatow, line 126: Hm, but this was broken bc it means you can't use `ETH_PK` consistently. Not sure why this broke since the PR it was merged

scott-silver: when we run in CI, we probably don't run with `ETH_PK` set. so it would keep using `MNEMONIC`

but when you run it locally with `ETH_PK` it will fail

jflatow:
Closed in favor of #500

PR #501, opened 2022-08-02
Mainnet deploy script and scenarios

kevincheng96:
This PR introduces mainnet deploy scripts and allows us to fully test hypothetical deployments of Comet using scenarios without having to first deploy the contracts on-chain. This is useful for our mainnet launch, where we want to test a hypothetical mainnet deployment before actually deploying. 

A few big changes in this PR:
- Adding a new deployment script and `configuration.json` for `mainnet` that allows us to test it using scenarios
- Supporting mainnet governance scenarios by having `fastExecuteGovernance` work for deployments with a real `Timelock` that require votes to pass proposals
- Allowing the `configuration.json` file to specify an optional `address` field for the base asset and each collateral asset. This is useful for chains that already have pre-existing tokens we want to support (unlike testnets, where we deploy our own set of tokens)
- Fixing scenarios that were failing for mainnet

### DIFF
--- a/contracts/test/IGovernorBravo.sol
+++ b/contracts/test/IGovernorBravo.sol
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+/**
+ * @dev Interface for interacting with Governor bravo.
+ * Note Not a comprehensive interface
+ */
+interface IGovernorBravo {
+    enum ProposalState {
+        Pending,
+        Active,
+        Canceled,
+        Defeated,
+        Succeeded,
+        Queued,
+        Expired,
+        Executed
+    }
+
+    event ProposalCreated(
+        uint256 proposalId,
+        address proposer,
+        address[] targets,
+        uint256[] values,
+        string[] signatures,
+        bytes[] calldatas,
+        uint256 startBlock,
+        uint256 endBlock,
+        string description
+    );
+
+    event ProposalCanceled(uint256 proposalId);
+
+    event ProposalQueued(uint256 proposalId, uint256 eta);
+
+    event ProposalExecuted(uint256 proposalId);
+
+    function state(uint256 proposalId) external view returns (ProposalState);
+
+    function propose(
+        address[] memory targets,
+        uint256[] memory values,
+        string[] memory signatures,
+        bytes[] memory calldatas,
+        string memory description
+    ) external returns (uint256 proposalId);
+
+    function queue(uint256 proposalId) external payable;
+
+    function execute(uint256 proposalId) external payable;
+
+    function castVote(uint256 proposalId, uint8 support) external returns (uint256 balance);
+}
--- a/deployments/mainnet/configuration.json
+++ b/deployments/mainnet/configuration.json
@@ -32,7 +32,7 @@
       "borrowCF": 0.65,
       "liquidateCF": 0.73,
       "liquidationFactor": 0.93,
-      "supplyCap": "0e18"
+      "supplyCap": "500000e18"
     },
     "WBTC": {
       "address": "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599",
@@ -41,7 +41,7 @@
       "borrowCF": 0.7,
       "liquidateCF": 0.78,
       "liquidationFactor": 0.95,
-      "supplyCap": "0e8"
+      "supplyCap": "35000e8"
     },
     "WETH": {
       "address": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
@@ -50,7 +50,7 @@
       "borrowCF": 0.82,
       "liquidateCF": 0.9,
       "liquidationFactor": 0.95,
-      "supplyCap": "0e18"
+      "supplyCap": "1000000e18"
     },
     "UNI": {
       "address": "0x1f9840a85d5af5bf1d1762f925bdaddc4201f984",
@@ -59,7 +59,7 @@
       "borrowCF": 0.75,
       "liquidateCF": 0.83,
       "liquidationFactor": 0.93,
-      "supplyCap": "0e18"
+      "supplyCap": "50000000e18"
     },
     "LINK": {
       "address": "0x514910771af9ca656af840dff83e8264ecf986ca",
@@ -68,7 +68,7 @@
       "borrowCF": 0.79,
       "liquidateCF": 0.87,
       "liquidationFactor": 0.93,
-      "supplyCap": "0e18"
+      "supplyCap": "50000000e18"
     }
   }
 }
--- a/deployments/mainnet/configuration.json
+++ b/deployments/mainnet/configuration.json
@@ -21,7 +21,7 @@
   "tracking": {
     "indexScale": "1e15",
     "baseSupplySpeed": "0.000011574074074074073e15",
-    "baseBorrowSpeed": "0.0011458333333333333e15",
+    "baseBorrowSpeed": "0.000578703703703703703e15",
     "baseMinForRewards": "1000000e6"
   },
   "assets": {
--- a/deployments/mainnet/configuration.json
+++ b/deployments/mainnet/configuration.json
@@ -20,8 +20,8 @@
   },
   "tracking": {
     "indexScale": "1e15",
-    "baseSupplySpeed": "0.000011574074074074073e15",
-    "baseBorrowSpeed": "0.000578703703703703703e15",
+    "baseSupplySpeed": "0e15",
+    "baseBorrowSpeed": "0e15",
     "baseMinForRewards": "1000000e6"
   },
   "assets": {
--- a/deployments/mainnet/configuration.json
+++ b/deployments/mainnet/configuration.json
@@ -32,7 +32,7 @@
       "borrowCF": 0.65,
       "liquidateCF": 0.7,
       "liquidationFactor": 0.92,
-      "supplyCap": "500000e18"
+      "supplyCap": "0e18"
     },
     "WBTC": {
       "address": "0x2260fac5e5542a773aa44fbcfedf7c193bc2c599",
@@ -41,7 +41,7 @@
       "borrowCF": 0.7,
       "liquidateCF": 0.75,
       "liquidationFactor": 0.93,
-      "supplyCap": "35000e8"
+      "supplyCap": "0e8"
     },
     "WETH": {
       "address": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
@@ -50,7 +50,7 @@
       "borrowCF": 0.82,
       "liquidateCF": 0.85,
       "liquidationFactor": 0.93,
-      "supplyCap": "1000000e18"
+      "supplyCap": "0e18"
     },
     "UNI": {
       "address": "0x1f9840a85d5af5bf1d1762f925bdaddc4201f984",
@@ -59,7 +59,7 @@
       "borrowCF": 0.75,
       "liquidateCF": 0.8,
       "liquidationFactor": 0.92,
-      "supplyCap": "50000000e18"
+      "supplyCap": "0e18"
     },
     "LINK": {
       "address": "0x514910771af9ca656af840dff83e8264ecf986ca",
@@ -68,7 +68,7 @@
       "borrowCF": 0.75,
       "liquidateCF": 0.8,
       "liquidationFactor": 0.92,
-      "supplyCap": "50000000e18"
+      "supplyCap": "0e18"
     }
   }
 }
--- a/deployments/mainnet/configuration.json
+++ b/deployments/mainnet/configuration.json
@@ -4,6 +4,8 @@
   "baseTokenAddress": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
   "baseTokenPriceFeed": "0x8fFfFfd4AfB6115b954Bd326cbe7B4BA576818f6",
   "borrowMin": "1000e6",
+  "governor": "0x6d903f6003cca6255d85cca4d3b5e5146dc33925",
+  "pauseGuardian": "0xbbf3f1421d886e9b2c5d716b5192ac998af2012c",
   "storeFrontPriceFactor": 0.5,
   "targetReserves": "5000000e6",
   "rates": {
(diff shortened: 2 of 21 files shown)


contracts/test/IGovernorBravo.sol
```diff
@@ -0,0 +1,53 @@
...
+        bytes[] memory calldatas,
+        string memory description
+    ) external returns (uint256 proposalId);
+
+    function queue(uint256 proposalId) external payable;
+
+    function execute(uint256 proposalId) external payable;
+
+    function castVote(uint256 proposalId, uint8 support) external returns (uint256 balance);
+}
```
jflatow, line 53: 🙌

scenario/utils.ts
```diff
@@ -99,13 +99,64 @@ export async function setNextBaseFeeToZero(world: World) {
 
 // Instantly executes some actions through the governance proposal process
 // Note: `governor` must be connected to an `admin` signer
-export async function fastGovernanceExecute(governor: GovernorSimple, targets: string[], values: BigNumberish[], signatures: string[], calldatas: string[]) {
-  let tx = await (await governor.propose(targets, values, signatures, calldatas, 'FastExecuteProposal')).wait();
-  let event = tx.events.find(event => event.event === 'ProposalCreated');
-  let [proposalId] = event.args;
+export async function fastGovernanceExecute(world: World, governor: GovernorSimple, targets: string[], values: BigNumberish[], signatures: string[], calldatas: string[]) {
+  // XXX See if there is a better way to determine if GovernorSimple or Bravo should be used
```
jflatow, line 103: Feels like the right thing would be to actually make GovernorSimple admins just each be ultra-powerful voters, that way the interface is exactly wrt proposals, but maybe too late for that

scenario/utils.ts
```diff
@@ -99,13 +99,64 @@ export async function setNextBaseFeeToZero(world: World) {
...
+    await world.hre.ethers.provider.send('evm_setNextBlockTimestamp', [eta.toNumber()])
+    await governorAsAdmin.execute(proposalId);
+  }
+}
+
+// XXX Hardhat 2.9 has 'hardhat_mine' for mining multiple blocks
+async function mineBlocks(world: World, blocks: number) {
+  for (let i = 0; i < blocks; i++) {
+    await world.hre.network.provider.send('evm_mine', []);
+  }
```
jflatow, line 159: Yeah yikes, this might be too slow

src/deploy/NetworkConfiguration.ts
```diff
@@ -112,6 +114,12 @@ function getContractAddress(contractName: string, contractMap: ContractMap): str
   return contract.address;
 }
 
+// XXX but if contract address exists AND optional address is set in config,
+// which one takes precedence? probably override
```
jflatow, line 118: probably

src/deploy/NetworkConfiguration.ts
```diff
@@ -178,7 +186,7 @@ export async function getConfiguration(
   let contractMap = contractMapOverride ?? await deploymentManager.contracts();
 
   let symbol = networkConfiguration.symbol;
-  let baseToken = getContractAddress(networkConfiguration.baseToken, contractMap);
+  let baseToken = hasContractAddress(networkConfiguration.baseToken, contractMap) ? getContractAddress(networkConfiguration.baseToken, contractMap) : networkConfiguration.baseTokenAddress;
```
jflatow, line 189: Should we just add a third fallback/default arg to `getContractAddress`?

deployments/mainnet/configuration.json
```diff
@@ -0,0 +1,72 @@
...
+  },
+  "assets": {
+    "COMP": {
+      "address": "0xc00e94cb662c3520282e6f5717214004a7f26888",
+      "priceFeed": "0xdbd020CAeF83eFd542f4De03e3cF0C28A4428bd5",
+      "decimals": "18",
+      "borrowCF": 0.65,
+      "liquidateCF": 0.7,
+      "liquidationFactor": 0.92,
+      "supplyCap": "500000e18"
```
jflatow, line 33: I guess these will be 0

deployments/mainnet/configuration.json
```diff
@@ -0,0 +1,72 @@
...
+    "supplyBase": 0,
+    "borrowKink": 0.8,
+    "borrowSlopeLow": 0.035,
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000011574074074074073e15",
+    "baseBorrowSpeed": "0.0011458333333333333e15",
```
jflatow, line 22: I still think we should set non-zero supply speed, personally

deployments/mainnet/deploy.ts
```diff
@@ -0,0 +1,70 @@
...
+  };
+}
+
+async function getErc20ContractAt(deploymentManager: DeploymentManager, address: string): Promise<Contract> {
+  return deploymentManager.hre.ethers.getContractAt(
+    'ERC20',
+    address,
+    await deploymentManager.getSigner()
+  );
+}
```
jflatow, line 70: Awesome, so the main thing missing after all this will be the `migration` with a proposal to (1) set supply caps (2) move Timelock USDC (3) stream Comptroller COMP, right?

src/deploy/NetworkConfiguration.ts
kevincheng96: Yeah that's cleaner. I added a third arg to `getContractAddress`.

scenario/utils.ts
kevincheng96: Yeah didn't have enough foresight to make that decision. A little late to do it now with the launch coming up, but definitely worth considering after mainnet launch since it will help us to simplify the code paths

scenario/utils.ts
kevincheng96: Should we consider bumping up our hardhat version to 2.9+? They have `hardhat_mine`

deployments/mainnet/deploy.ts
kevincheng96: Yep!

src/deploy/Network.ts
```diff
@@ -99,14 +100,14 @@ export async function deployNetworkComet(
     targetReserves,
     assetConfigs,
   } = {
-    governor: timelock.address,
-    pauseGuardian: timelock.address,
+    governor: timelock ? timelock.address : ethers.constants.AddressZero,
+    pauseGuardian: timelock ? timelock.address : ethers.constants.AddressZero,
```
kevincheng96, line 104: Thoughts on these changes? `timelock` will be undefined if the script does not deploy a `SimpleTimelock` AND there is no existing `timelock` in `aliases`. This happens for deployments that haven't been deployed on-chain yet and already have an existing `Timelock` contract on that chain (e.g. mainnet).

jflatow: But wouldn't it be good to err, since shouldn't any deployment actually have a `timelock`?

scenario/utils.ts
jflatow: I can try it again today, we had an issue with it before related to a change in the http library. I think it'll also require new cuts of our own `hardhat`, since they still haven't patched the issue that was breaking `hardhat-cover` (see https://github.com/NomicFoundation/hardhat/pull/2399 and https://github.com/NomicFoundation/hardhat/pull/2400)

deployments/mainnet/configuration.json
kevincheng96: The current configuration is 1 and 99 COMP per day, but the launch doc says 0 and 50. Should I go and make those changes? I do agree that non-zero supply speed would be nice though.

deployments/mainnet/configuration.json
kevincheng96: Done.

deployments/mainnet/configuration.json
```diff
@@ -0,0 +1,74 @@
...
+    "baseMinForRewards": "1000000e6"
+  },
+  "assets": {
+    "COMP": {
+      "address": "0xc00e94cb662c3520282e6f5717214004a7f26888",
+      "priceFeed": "0xdbd020CAeF83eFd542f4De03e3cF0C28A4428bd5",
+      "decimals": "18",
+      "borrowCF": 0.65,
+      "liquidateCF": 0.7,
+      "liquidationFactor": 0.92,
```
kevincheng96, line 34: Have we finalized on these values yet? I can update these values to the ones in the `Comet Launch` doc if they are final. Otherwise, we can update these in a follow-up PR.

src/deploy/Network.ts
kevincheng96: It wouldn't necessarily be an error, since the timelock could still be set in the configuration overrides and/or directly in `configuration.json`. Pre-launch mainnet, for example, would have no `SimpleTimelock` and no existing `timelock` in `aliases`. But we specify `governor` and `pauseGuardian` directly in the `configuration.json`.

Tbh, this is all very confusing though because there's so many ways a single value can be set, and each way having a different precedence over the other.

deployments/mainnet/configuration.json
jflatow: Yea I still think we should do non-zero, but I think 50 total is more final

deployments/mainnet/configuration.json
jflatow: I think we should bring it into alignment with the doc to make sure scens run, although can't promise people won't change their minds again

src/deploy/Network.ts
jflatow: It'll be an error when we try to use `timelock` and there's no address, right? Agree its confusing and I'm not sure I totally know what we are talking about or why we need so many paths, so maybe we can parking lot today.

deployments/mainnet/configuration.json
kevincheng96: Oh man, this is actually causes a lot of the mainnet scenarios to fail (which kind of makes sense I guess?)

deployments/mainnet/configuration.json
kevincheng96: Ok left supply speed as non-zero and changed borrow speed to 50.

deployments/mainnet/configuration.json
jflatow: Yeah, makes a lot of sense, in fact we would hope that. So ideally when we have the real migration and proposal constraint, that one will pass, but need to think how we control/see/verify what passes is correct 🤔

deployments/mainnet/configuration.json
kevincheng96: Ok done, updated these values to be aligned with the doc.

src/deploy/Network.ts
kevincheng96: Yeah it'll throw an error trying to access address on an `undefined` value.

plugins/scenario/utils/TokenSourcer.ts
```diff
@@ -48,7 +48,7 @@ async function removeTokens(
   let tokenContract = new ethers.Contract(asset, erc20, signer);
   let currentBalance = await tokenContract.balanceOf(address);
   if (currentBalance.lt(amount)) throw 'Error: Insufficient address balance';
-  await tokenContract.transfer('0x0000000000000000000000000000000000000000', amount);
+  await tokenContract.transfer('0x0000000000000000000000000000000000000001', amount);
```
jflatow, line 51: What's the reason for this one again?

kevincheng96: Some ERC20s disallow transfers to the zero address

deployments/mainnet/configuration.json
kevincheng96: During parking lot, we decided to temporarily revert these to non-zero values to make sure all scenarios are passing. Once we are sure scenarios pass (which they are right now), we can set these values back to zero and merge.

deployments/mainnet/configuration.json
kevincheng96: Actually, setting supply speed as non-zero means people can start accruing rewards before supply caps have been increased from zero. I'm not sure if this is something we want, since I assume we don't want ppl to be accruing rewards until the governance proposal has been executed.

deployments/mainnet/configuration.json
```diff
@@ -0,0 +1,74 @@
+{
+  "symbol": "📈USDC",
```
scott-silver, line 2: Just a heads-up: if #498 gets merged before this PR, we'll need to move these files into `deployments/mainnet/usdc/`

hardhat.config.ts
```diff
@@ -165,6 +165,12 @@ const config: HardhatUserConfig = {
 
   scenario: {
     bases: [
+      {
```
scott-silver, line 168: this would change, too, if #498 gets merged.

it would be:

```
{
  name: 'mainnet',
  network: 'mainnet',
  deployment: 'usdc',
  allocation: 0.1
}
```

plugins/scenario/utils/hreForBase.ts
```diff
@@ -58,7 +58,7 @@ function hreForBase(base: ForkSpec): HardhatRuntimeEnvironment {
 
   const hardhatArguments = getEnvHardhatArguments(HARDHAT_PARAM_DEFINITIONS, process.env);
 
-  const config = loadConfigAndTasks(hardhatArguments);
+  const { resolvedConfig: config } = loadConfigAndTasks(hardhatArguments);
```
scott-silver, line 61: what's the difference between `config` and `resolvedConfig`?

deployments/mainnet/configuration.json
jflatow: Oh interesting, good point, bc people can supply base, but they can't borrow until supply caps are non-zero 🤔 I guess we could raise `baseMinForRewards` and lower it in the proposal also, though doubt we like that idea. I guess we don't really have any way of preventing people from supplying base until proposal is executed regardless. If the rewards are tiny, I'm not sure it matters, and maybe its even a good thing to get supply in...

jflatow: Probably a good argument to set rewards speeds in the proposal as well

kevincheng96: Yeah setting reward speeds in the proposal is a good option.

plugins/scenario/utils/hreForBase.ts
jflatow: They added a resolved config and user config

jflatow: This was from my bump to 2.9 PR which got merged into this one (for `mineBlocks`)

deployments/mainnet/configuration.json
kevincheng96: I've set the initial reward speeds to 0.

jflatow (approved):
👏

PR #505, opened 2022-08-03
Patch for fastGovernanceExecute to avoid nonce issue

jflatow:
Hopefully this is temporary and we have deeper fixes for this function.


jflatow:
⛑️ found a better fix

PR #506, opened 2022-08-03
add recipient to Liquidator.sol

scott-silver:
adding an immutable recipient address to the Liquidator contract. when you deploy, you specify the address that should receive the profits of any liquidations.

this should prevent generalized front-runners from taking profitable transactions from the caller of the Liquidator contract, since a front-runner that mimicked the arguments would only end up sending profits to the specified address.

it has the added benefit of allowing us to initiate transactions from a hot wallet but automatically transfer liquidated funds to a cold wallet.

cc: @ajb413

### DIFF
--- a/test/liquidation/makeLiquidatableProtocol.ts
+++ b/test/liquidation/makeLiquidatableProtocol.ts
@@ -139,9 +139,14 @@ export default async function makeLiquidatableProtocol() {
     totalBorrowBase: 2e13
   });
 
+
+  // create underwater user
+  const [signer, underwaterUser, recipient] = await ethers.getSigners();
+
   // build Liquidator
   const Liquidator = await ethers.getContractFactory('Liquidator') as Liquidator__factory;
   const liquidator = await Liquidator.deploy(
+    recipient.address,
     ethers.utils.getAddress(SWAP_ROUTER),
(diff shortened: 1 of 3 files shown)


jflatow (approved):
💪

test/liquidation/makeLiquidatableProtocol.ts
```diff
@@ -139,9 +139,14 @@ export default async function makeLiquidatableProtocol() {
     totalBorrowBase: 2e13
   });
 
+
```
kevincheng96, line 142: nit: extra line

kevincheng96 (approved):
Looks good. Just one thought about whether or not we should make the recipient mutable. Probably not, but we'll just have to redeploy the contract if we ever want to change that field.

test/liquidation/makeLiquidatableProtocol.ts
scott-silver: deleted

scott-silver:
> Looks good. Just one thought about whether or not we should make the recipient mutable. Probably not, but we'll just have to redeploy the contract if we ever want to change that field.

yeah, I think we'd just redeploy. otherwise we'd need to add a governor address that is allowed to update the recipient address, which seems like more trouble than its worth

PR #508, opened 2022-08-04
Remove extraneous relation configs

jflatow:


kevincheng96 (approved):

hardhat.config.ts
```diff
@@ -130,8 +128,7 @@ const config: HardhatUserConfig = {
       blockGasLimit: 12000000,
       accounts: ETH_PK ? [...deriveAccounts(ETH_PK)].map(privateKey => ({ privateKey, balance: (10n ** 36n).toString() })) : { mnemonic: MNEMONIC },
       // this should be default commented out and only enabled during dev to allow partial testing
-      // XXX comment out by default once we've made the full contract fit
-      allowUnlimitedContractSize: true,
+      //allowUnlimitedContractSize: true,
```
kevincheng96, line 131: Seems like we still need this. Coverage is failing due to the contract size being too large.

jflatow: Added back but changed the comment to not be XXX anymore since it seems like its going to be permanent

PR #509, opened 2022-08-04
Add Seacrest Support

hayesgm:
This patch adds Seacrest to our GitHub Actions for Deploy, Prepare and Enact. As per [Seacrest](https://github.com/hayesgm/seacrest), when you run these actions, you will be prompted with a QR Code to connect WalletConnect. Once you scan that QR code (with, e.g. MetaMask Mobile), the rest of the code will depend on that account for any transactions (e.g. deploying a contract). You will individually sign off on any requests during the progression of the action without any risk of key exfiltration.
    
Note: this patch removes `eth_pk` as an option, but technically we could have this as an either/or. I just tend to think it's superior enough that we should supplant the old flow.

### DIFF
--- a/.github/workflows/deploy-market.yaml
+++ b/.github/workflows/deploy-market.yaml
@@ -15,7 +15,7 @@ on:
         type: boolean
         description: Simulate
       eth_pk:
-        description: Ethereum Private Key
+        description: Ignore if you plan to use WalletConnect, otherwise, you can paste in a Ethereum private key
 jobs:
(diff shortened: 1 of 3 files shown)


jflatow (commented):
Cool that this is so simple, excited to give it a try! I think we need to discuss how the actual governance proposal is going to get made and which wallet its going to come from before deciding what exactly to do here

jflatow (approved):
🙌

PR #511, opened 2022-08-04
Start cleaning up a little to simplify some further refactoring

jflatow:
Eventually setting up for idempotent deploy


plugins/import/import.ts
```diff
@@ -14,7 +13,7 @@ export async function loadContract(source: string, network: string, address: str
   }
   switch (source) {
     case 'etherscan':
-      return await loadEtherscanContractMemoized(network, address);
+      return await loadEtherscanContract(network, address);
```
scott-silver, line 16: i think we're going to miss memoizing this, no?

jflatow: I don't ever see it being a hit, and probably because we already have multiple levels of caching

scott-silver: 👌

scott-silver (approved):

antonina-cherednichenko (approved):

jflatow:
Covered by #517 #520 and #521

PR #512, opened 2022-08-04
Set RewardConfig during initial deployment

kevincheng96:
This PR updates the deploy function to set the `RewardConfig` in the `CometRewards` contract as part of the deployment process. 

Two optional fields have been added to `configuration.json` to specify the reward token for a certain deployment: `rewardToken` (token symbol) and `rewardTokenAddress`. If the rewards contract is to be deployed, then one of these two fields must be specified in the deployment's `configuration.json`.

### DIFF
--- a/src/deploy/Network.ts
+++ b/src/deploy/Network.ts
@@ -77,6 +77,16 @@ export async function deployNetworkComet(
     );
   }
 
+  const protocolConfiguration = {
+    governor: timelock ? timelock.address : ethers.constants.AddressZero,
+    pauseGuardian: timelock ? timelock.address : ethers.constants.AddressZero,
+    ...await getConfiguration(deploymentManager, contractMapOverride, configurationOverrides),
+  };
+
+  // Validate that the configuration is properly set before trying to deploy contracts
+  // XXX awkward that this has to be here instead of in the beginning of the function
+  validateConfiguration(contractsToDeploy, protocolConfiguration);
+
   const {
     symbol,
     governor,
@@ -100,11 +110,7 @@ export async function deployNetworkComet(
     targetReserves,
     assetConfigs,
     rewardTokenAddress
-  } = {
-    governor: timelock ? timelock.address : ethers.constants.AddressZero,
-    pauseGuardian: timelock ? timelock.address : ethers.constants.AddressZero,
-    ...await getConfiguration(deploymentManager, contractMapOverride, configurationOverrides),
-  };
+  } = protocolConfiguration;
 
   if (shouldDeploy(contractsToDeploy.all, contractsToDeploy.cometExt)) {
     const extConfiguration = {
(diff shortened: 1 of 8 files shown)


src/deploy/Network.ts
```diff
@@ -245,6 +237,31 @@ export async function deployNetworkComet(
...
 
+
+  if (shouldDeploy(contractsToDeploy.all, contractsToDeploy.rewards)) {
+    rewards = await deploymentManager.deploy<CometRewards, CometRewards__factory, [string]>(
+      'CometRewards.sol',
+      [adminSigner.address]
+    );
+
+    // Set the rewards config for Comet in CometRewards
+    // XXX we should validate rewardTokenAddress is set earlier
```
kevincheng96, line 248: Planning on adding some input validation so we can fail early.

kevincheng96: Done

src/deploy/Network.ts
```diff
@@ -77,6 +77,16 @@ export async function deployNetworkComet(
...
   }
 
+  const protocolConfiguration = {
+    governor: timelock ? timelock.address : ethers.constants.AddressZero,
+    pauseGuardian: timelock ? timelock.address : ethers.constants.AddressZero,
+    ...await getConfiguration(deploymentManager, contractMapOverride, configurationOverrides),
+  };
+
+  // Validate that the configuration is properly set before trying to deploy contracts
+  // XXX awkward that this has to be here instead of in the beginning of the function
```
kevincheng96, line 87: We can't move this validation to the beginning right now because the code paths for deploying `SimpleTimelock` and `GovernorSimple` need to be executed before we can generate `protocolConfiguration`. I'm thinking about moving the deployment of `SimpleTimelock` and `GovernorSimple` out of `deployNetworkComet` and into each testnet's own `deploy.ts` since they are not production contracts.

Will tackle this refactor in a separate PR though.

src/deploy/Network.ts
```diff
@@ -245,6 +243,30 @@ export async function deployNetworkComet(
...
+    debug(`Setting RewardConfig in CometRewards`);
+    await deploymentManager.asyncCallWithRetry(
+      (signer_) => wait(rewards.connect(adminSigner_ ?? signer_).setRewardConfig(cometProxy.address, rewardTokenAddress))
+    );
+
+    // Transfer ownership of CometRewards
+    debug(`Transferring ownership of CometRewards`);
+    await deploymentManager.asyncCallWithRetry(
+      (signer_) => wait(rewards.connect(adminSigner_ ?? signer_).transferGovernor(governor))
+    );
```
jflatow, line 263: Basically these are the 2 steps I should need for my refactor right?

jflatow (approved):
👍

src/deploy/Network.ts
kevincheng96: Yep

src/deploy/Network.ts
kevincheng96: This will actually be tackled by Jared in his refactoring/idempotent deploys PR.

PR #513, opened 2022-08-04
Proposal constraint

antonina-cherednichenko:
Add proposal constraint for executing `GovernorBravo` pending proposals, works only for the mainnet


scenario/constraints/ProposalConstraint.ts
```diff
@@ -1,45 +1,144 @@
...
+
 export class ProposalConstraint<T extends CometContext, R extends Requirements> implements Constraint<T, R> {
   async solve(requirements: R, context: T, world: World) {
     let solutions: Solution<T>[] = [];
 
-    // XXX pull all pending proposals for the associated gov (chain)
-    //  produce all the subsets and queue/execute each batch of them
-    //   to form each solution
-    let pendingProposals = [];
+    // Only run migration for mainnet scenarios
```
jflatow, line 65: Nice, I would keep an `XXX` here since this isn't meant to be a permanent limitation

scenario/constraints/ProposalConstraint.ts
```diff
@@ -1,45 +1,144 @@
...
+
+          // Execute proposal
+          await governorAsAdmin.execute(proposalId);
+
+          debug(`Pending proposal ${proposalId} was executed`);
+          return context;
+        } catch (err) {
+          debug(`Failed with error ${err}`);
+          return context;
+        }
```
jflatow, line 134: Look good!

jflatow: Wondering if there might be a way to break this down into a bit more composable units 🤔

scenario/constraints/ProposalConstraint.ts
```diff
@@ -1,45 +1,144 @@
...
+    const voters = [
+      '0xea6c3db2e7fca00ea9d7211a03e83f568fc13bf7',
+      '0x61258f12c459984f32b83c86a6cc10aa339396de',
+      '0x9aa835bc7b8ce13b9b0c9764a52fbf71ac62ccf1',
+      '0x683a4f9915d6216f73d6df50151725036bd26c02',
+      '0xa1b61405791170833070c0ea61ed28728a840241',
+      '0x88fb3d509fc49b515bfeb04e23f53ba339563981',
+      '0x8169522c2c57883e8ef80c498aab7820da539806'
+    ];
+    await fundVoters(world, voters, context.primaryActor());
```
kevincheng96, line 82: You can also consider setting gas to 0 like this: https://github.com/compound-finance/comet/pull/512/commits/12608c9396715be72ca47c6a94f7d2dfe68f2ec8

scenario/constraints/ProposalConstraint.ts
kevincheng96: Since most of the logic in this for-loop seems to be the same as `context.fastGovernanceExecute(calldata)`, we could create a version of `context.fastGovernanceExecute(proposalId)` that takes a proposalId as an argument.

The two versions of `context.fastGovernanceExecute` will pretty much share the same code and we won't have to duplicate the logic here.

kevincheng96 (commented):
Looks good! Just tagging onto 🍯 's comment of potentially sharing some of the code with `fastGovernanceExecute`.

scenario/constraints/ProposalConstraint.ts
antonina-cherednichenko: agree, I actually working on such a version right now. So that `fastGovernanceExecute` and proposal constraint reuse the same code.

scenario/constraints/ProposalConstraint.ts
antonina-cherednichenko: good point, resolved

scenario/constraints/ProposalConstraint.ts
antonina-cherednichenko: I added that instead of `fundVoters`

kevincheng96 (approved):
LGTM!

scenario/constraints/ProposalConstraint.ts
```diff
@@ -1,45 +1,104 @@
...
+    const voters = [
+      '0xea6c3db2e7fca00ea9d7211a03e83f568fc13bf7',
+      '0x61258f12c459984f32b83c86a6cc10aa339396de',
+      '0x9aa835bc7b8ce13b9b0c9764a52fbf71ac62ccf1',
+      '0x683a4f9915d6216f73d6df50151725036bd26c02',
+      '0xa1b61405791170833070c0ea61ed28728a840241',
+      '0x88fb3d509fc49b515bfeb04e23f53ba339563981',
+      '0x8169522c2c57883e8ef80c498aab7820da539806'
+    ];
+    await context.setNextBaseFeeToZero();
```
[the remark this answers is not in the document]
antonina-cherednichenko: done, probably worth setting gas price to 0 for the networks

scenario/context/CometContext.ts
```diff
@@ -52,6 +52,18 @@ export class CometContext {
...
+  // XXX find a better way to do this without hardcoding whales
+  voters = [
+    '0xea6c3db2e7fca00ea9d7211a03e83f568fc13bf7',
+    '0x61258f12c459984f32b83c86a6cc10aa339396de',
+    '0x9aa835bc7b8ce13b9b0c9764a52fbf71ac62ccf1',
+    '0x683a4f9915d6216f73d6df50151725036bd26c02',
+    '0xa1b61405791170833070c0ea61ed28728a840241',
+    '0x88fb3d509fc49b515bfeb04e23f53ba339563981',
+    '0x8169522c2c57883e8ef80c498aab7820da539806'
+  ];
```
jflatow, line 65: We should figure out how to reconcile this list for `mainnet` with the sourced ones 🥈 will be working on

scenario/context/CometContext.ts
```diff
@@ -248,12 +260,58 @@ export class CometContext {
...
+    // Queue proposal
+    await this.setNextBaseFeeToZero();
+    const queueTxn = await (await governorAsAdmin.queue(proposalId, { gasPrice: 0 })).wait();
+    const queueEvent = queueTxn.events?.find(event => event.event === 'ProposalQueued');
+    await this.setNextBlockTimestamp(queueEvent?.args?.eta.toNumber());
+
+    // Execute proposal
+    await this.setNextBaseFeeToZero();
+    await governorAsAdmin.execute(proposalId, { gasPrice: 0 });
+  }
```
jflatow, line 307: 👏

scenario/context/CometContext.ts
```diff
@@ -248,12 +260,58 @@ export class CometContext {
...
+        ) as IGovernorBravo;
+        await this.setNextBaseFeeToZero();
+        await govAsVoter.castVote(proposalId, 1, { gasPrice: 0 });
+      } catch (err) {
+        debug(`Error while voting ${err}`);
+      }
+    }
+    await this.mineBlocks(blocksFromStartToEnd);
+
+    const adminSigner = await this.world.impersonateAddress(this.voters[0]);
```
jflatow, line 291: Just FYI, on my branch I added `value` arg to `impersonateAddress`, which for signers which *can* receive eth is slightly better than having to set fees/price to zero -- only really necessary if the call is in a function we would call as part of `deploy` and this works totally well for scenario-only code

scenario/context/CometContext.ts
```diff
@@ -282,26 +335,7 @@ export class CometContext {
...
-      }
-
-      await this.mineBlocks(blocksFromStartToEnd);
-      const queueTxn = await (await governorAsAdmin.queue(proposalId)).wait();
-      const queueEvent = queueTxn.events.find(event => event.event === 'ProposalQueued');
-      let [proposalId_, eta] = queueEvent.args;
-
-      await this.setNextBlockTimestamp(eta.toNumber());
-      await governorAsAdmin.execute(proposalId);
+      this.executePendingProposal(proposalId, startBlock, endBlock);
```
jflatow, line 338: Is this missing `await`?

antonina-cherednichenko: good catch, definitely missing here :(

scenario/context/CometContext.ts
antonina-cherednichenko: I can do in separate pr once it's done

scenario/context/CometContext.ts
antonina-cherednichenko: Noted, once merged, I can change that

PR #516, opened 2022-08-08
Improvements to CometProxyAdmin and Configurator

kevincheng96:
This PR pulls out and isolates the contract changes introduced in Jared's refactoring PR #515. I've added unit tests and scenarios to test the new code paths.

### DIFF
--- a/contracts/CometProxyAdmin.sol
+++ b/contracts/CometProxyAdmin.sol
@@ -17,4 +17,14 @@ contract CometProxyAdmin is ProxyAdmin {
         address newCometImpl = configuratorProxy.deploy(address(cometProxy));
         upgrade(cometProxy, newCometImpl);
     }
+
+    /**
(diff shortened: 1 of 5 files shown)


jflatow (approved):
Nice! For posterity/reviewers, there are 2 changes here discovered when trying to streamline deployments for mainnet/scenarios. 

The first adds an `deployUpgradeToAndCall` so that we can atomically perform the first Comet impl deploy + initialize, and so that governance proposals only need a single action to upgrade and initialize a new Comet impl.

The second is a change to Configurator which re-allows re-setting the Configuration for a Comet, but only if the invariants on `baseToken` and `tokenIndexScale` are not violated. This means we can always call `setConfiguration(newConfig)` e.g. in scenarios to upgrade, instead of having to diff and call each of the setters for any changed fields (which is also error prone), and may one day be useful for an otherwise-too-large governance proposal.

scott-silver (approved):

PR #517, opened 2022-08-09
Massive refactor to get clean deploys and proper scen upgrades

jflatow:
Lots of fixes/improvements to spider as well, since aliases really need to work.

* Start cleaning up a little to simplify some further refactoring
* Eventually setting up for idempotent deploy
* Make spider powerful enough to pull *all* the aliases we have
* WIP 1
* WIP 2
* WIP 3
* WIP 4
* WIP 5
* WIP 6
* WIP 7
* WIP 8
 spider for USDC proxyAdmin
* WIP 9
* scenarios passing on kovan upgrades and deploys
* Mainnet deployment and scens working
* Fuji deployment and scens


deployments/fuji/usdc/configuration.json
```diff
@@ -21,7 +21,6 @@
     "baseBorrowSpeed": "0.0011458333333333333e15",
     "baseMinForRewards": "1000000e6"
   },
-  "rewardToken": "WAVAX",
```
jflatow, line 24: Note: this is because scens are correctly failing to 'complete' the deploy, as this is not the currently configured rewards token in the roots contracts, and the normal signer cannot make the change (for an upgrade we would impersonate, but not for the 'initial' deploy). I believe the correct thing to do would be either (a) create a migration with the change, then possibly updating `configuration.json` or (b) wipe the `roots.json` and do a redeploy (which we should probably do now).

plugins/deployment_manager/ContractMap.ts
```diff
@@ -31,7 +31,9 @@ export async function getContractsFromAliases(
       address
     );
 
-    contracts.set(alias, contract);
+    if (contract) {
```
hayesgm, line 34: This is so we don't accidentally wipe something if the contract is missing? When would it be missing?

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -133,46 +146,105 @@ export class DeploymentManager {
     );
   }
 
+  /* Unconditionally casts a contract as the given artifact type, without caching */
+  async cast<C extends Contract>(address: string, artifact: string): Promise<C> {
```
hayesgm, line 150: By no means suggesting now, but we should probably break these functions out into modules with tests at a later point when we have time.

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -133,46 +146,105 @@ export class DeploymentManager {
...
+      const buildFile = await this.import(address, fromNetwork);
+      const contract: C = await this._deployBuild(buildFile, deployArgs, retries);
+      await this.putAlias(alias, contract.address);
+      return contract;
+    }
+    return maybeExisting;
+  }
+
+  /* Conditionally deploy a contract with its alias if it does not exist, or if forced */
+  async deploy<C extends Contract, DeployArgs extends Array<any>>(
```
hayesgm, line 176: This is more like "I'm ron burgundy?", I mean `deploy?`. or really `fetchOrDeploy`

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -133,46 +146,105 @@ export class DeploymentManager {
...
+    if (!maybeExisting || force) {
+      // NB: might use an override alias on deployOpts, but it doesn't invalidate cache...
+      const contract: C = await this._deploy(contractFile, deployArgs, retries);
+      await this.putAlias(alias, contract.address);
+      return contract;
+    }
+    return maybeExisting;
+  }
+
+  async existing<C extends Contract>(
```
hayesgm, line 193: This is really `fetchOrImport`, right?

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -133,46 +146,105 @@ export class DeploymentManager {
...
+    const deployScript = this.cache.getFilePath({ rel: 'deploy.ts' });
+    const { default: deployFn } = await import(deployScript);
+    if (!deployFn || !deployFn.call) {
+      throw new Error(`Missing deploy function in ${deployScript}.`);
     }
+    const deployed = await deployFn(this, deploySpec);
+    const newCount = this.counter;
+    const newSpider = await this.spider(deployed);
+    return {
+      old: { count: oldCount, spider: oldSpider },
```
hayesgm, line 245: What's the point of tracking the counts?

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -190,54 +262,49 @@ export class DeploymentManager {
...
+    await storeAliases(this.cache, crawl.aliases);
+    await storeProxies(this.cache, crawl.proxies);
+    this.invalidateContractsCache(); // TODO should we just write new contracts from spider?
+    return crawl;
+  }
+
+  /* Stores a new alias, which can then be referenced via `deploymentManager.contract()` */
+  async putAlias(alias: Alias, address: Address) {
+    await putAlias(this.cache, alias, address);
+    this.invalidateContractsCache();
```
hayesgm, line 293: I forget why we need to do this, but I think it's meant to basically try and save some in-memory space by keeping some things cached so we don't overcall `new Contract`

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -289,24 +356,26 @@ export class DeploymentManager {
     this.cache.writeCacheToDisk = writeCacheToDisk;
   }
 
-  /* Generates a new migration file, e.g. `deployments/<network>/<deployment>/migrations/1644385406_my_new_migration.ts`
-   **/
+  /* Generates a new migration file, e.g. `deployments/<network>/<deployment>/migrations/1644385406_my_new_migration.ts` */
   async generateMigration(name: string, timestamp?: number): Promise<string> {
-    return await generateMigration(this.cache, name, timestamp);
+    return generateMigration(this.cache, name, timestamp);
```
hayesgm, line 361: FWIW, I still think `return await` is more expressive.

plugins/deployment_manager/Import.ts
```diff
@@ -29,22 +47,18 @@ export async function importContract(
...
       throw e;
     }
 
-    console.log(`Import failed for ${network}@${address}`);
-    console.log(`retrying in ${retryDelay / 1000}s; ${retries} retries left`);
+    console.warn(`Import failed for ${network}@${address} (${e.message}), retrying in ${retryDelay / 1000}s; ${retries} retries left`);
 
-    await new Promise((resolve) => setTimeout(resolve, retryDelay));
-    return await importContract(network, address, retries - 1, retryDelay * 2);
+    await new Promise(ok => setTimeout(ok, retryDelay));
```
hayesgm, line 59: 👌

plugins/deployment_manager/NonceManager.ts
```diff
@@ -5,6 +5,10 @@ import { providers } from 'ethers';
 
 // NonceManager does not implement `_signTypedData`, which is needed for the EIP-712 functions
 export class ExtendedNonceManager extends NonceManager implements TypedDataSigner {
+  async _reset() {
```
hayesgm, line 8: So nice.

plugins/deployment_manager/Spider.ts
```diff
@@ -1,211 +1,265 @@
...
-  return cross(addresses, aliasTemplates);
+  context: Ctx,
+  config: RelationInnerConfig,
+  defaultKeyAndTemplate: string
+): Promise<DiscoverNode[]> {
+  const addresses = await readField(contract, getFieldKey(config, defaultKeyAndTemplate), context);
+  const templates = config.alias ? asArray(config.alias) : [defaultKeyAndTemplate];
+  return addresses.map((address, i) => ({
+    address,
+    aliasRender: { template: templates[i % templates.length], i },
```
hayesgm, line 66: I'm a little confused with this one.

plugins/deployment_manager/Spider.ts
```diff
@@ -1,211 +1,265 @@
 import { Contract } from 'ethers';
-import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import { HardhatRuntimeEnvironment as HRE } from 'hardhat/types';
```
hayesgm, line 2: The changes overall to spider are really hard to see in a diff. Need to review further, but overall, this returns more or less the same data as the previous version?

plugins/deployment_manager/test/DeploymentManagerTest.ts
```diff
@@ -215,21 +211,6 @@ describe('DeploymentManager', () => {
     // TODO: Test cache invalidation?
   });
 
-  describe('putRoots', () => {
```
hayesgm, line 218: Why remove this test?

deployments/fuji/usdc/configuration.json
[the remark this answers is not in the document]
kevincheng96: I added this field in the recent PR: https://github.com/compound-finance/comet/pull/512. Should we bring COMP over to fuji or just say there isn't a rewards contract there?

deployments/fuji/usdc/deploy.ts
```diff
@@ -1,171 +1,123 @@
...
   );
 
-  let wbtc = await deploymentManager.clone(cloneAddr.wbtc, [], cloneNetwork);
-
-  let wavax = await deploymentManager.clone(cloneAddr.wavax, [], cloneNetwork);
-  // Give admin 0.01 WAVAX tokens [this is a precious resource here!]
-  debug(`Minting some WAVAX`);
-  await deploymentManager.asyncCallWithRetry(
-    (signer_) => wait(wavax.connect(signer_).deposit({ value: exp(0.01, 18) }))
+  const usdcProxyAdmin = await deploymentManager.deploy('USDC:admin', 'vendor/proxy/transparent/ProxyAdmin.sol', []);
```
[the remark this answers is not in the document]
jflatow: Yeah, exactly - same as old deploy but with the 'if not exists alias, deploy {clone, existing}' logic - this is why aliases really need to be powerful/strict in spider - we should get the same one back running fresh or crawling from roots

deployments/fuji/usdc/deploy.ts
```diff
@@ -1,171 +1,123 @@
...
+  const usdcImpl = await deploymentManager.clone('USDC:implementation', cloneAddr.usdcImpl, [], cloneNetwork);
+  const usdcProxy = await deploymentManager.clone('USDC', cloneAddr.usdcProxy, [usdcImpl.address], cloneNetwork);
+  const usdcProxyAdminSlot = '0x10d6a54a4754c8869d6886b5f5d7fbfa5b4522237ea5c60d11bc4e7a1ff9390b';
+  const USDC = usdcImpl.attach(usdcProxy.address);
+
+  await deploymentManager.idempotent(
+    async () => !sameAddress(await ethers.provider.getStorageAt(usdcProxy.address, usdcProxyAdminSlot), usdcProxyAdmin.address),
```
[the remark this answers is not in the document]
jflatow: Just that it adds the retry on the action and awaits the condition, and theoretically we might to add something like a label or move the debug message to a third arg so it would always say what was done/not done without explicit printing in the actions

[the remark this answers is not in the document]
jflatow: The idempotent in the branch name more refers to the additional alias arg you saw in `deploy`, `clone`, and `existing`

deployments/fuji/usdc/roots.json
```diff
@@ -1,6 +1,6 @@
 {
-  "comet": "0x0Ab7B12122Cb863C5198da8Ef245F29f568f43DA",
-  "configurator": "0x40d84100b3907d10bee625b6603aba9b7e23932e",
-  "fauceteer": "0x3C9360273195fBB0077C999eFa67b347C7A56b7f",
-  "rewards": "0x77084a315C6D4401D097D9cE149136783106E6AA"
+    "comet": "0x0Ab7B12122Cb863C5198da8Ef245F29f568f43DA",
```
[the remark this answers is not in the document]
jflatow: Idk, spider did this when I re-ran and I didn't fight it

deployments/relations.ts
```diff
@@ -47,41 +63,42 @@ let relationConfigMap: RelationConfigMap = {
...
     }
   },
+
   FiatTokenProxy: {
-    proxy: {
+    artifact: 'contracts/ERC20.sol:ERC20',
+    relations: {
+      fiatTokenAdmin: {
+        field: {
+          slot: '0x10d6a54a4754c8869d6886b5f5d7fbfa5b4522237ea5c60d11bc4e7a1ff9390b',
```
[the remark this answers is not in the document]
jflatow: The name is literally what it is https://github.com/OpenZeppelin/openzeppelin-labs/blob/master/initializer_contracts_with_args/contracts/AdminUpgradeabilityProxy.sol#L26

plugins/deployment_manager/ContractMap.ts
jflatow: This was for the `USDC:admin` not being a contract on `mainnet`, but which we need to crawl for testnet (since we deploy/use the admin, though still not sure why we need that, but didn't change it). In general spider fully supports crawling/aliasing non-contract addresses too now, but this was the last spot that was breaking when we actually try to include some

plugins/deployment_manager/DeploymentManager.ts
jflatow: Yeah, although by nature if you don't get the type of contract you want, you just won't be able to make your calls, so they are kind of tested by all the tests already (e.g. `cast` is core part of `deployComet` now to get the handle before the impl is set, which is why this was added)

plugins/deployment_manager/DeploymentManager.ts
jflatow: `idempotent` deploy 😬

plugins/deployment_manager/DeploymentManager.ts
jflatow: The point is we should be using these everywhere we actually do deploys by default, IMO (and are now on this branch)

plugins/deployment_manager/DeploymentManager.ts
jflatow: So we can print it out before the diff as an extra sanity check when running deploy from CLI or the initial deploy to each world for the scens, its a message that prints with/without DEBUG so you can eyeball that the number of contracts being deployed makes sense. In general once we have a real deployment there shouldn't be too many contracts ever being deployed

plugins/deployment_manager/DeploymentManager.ts
jflatow: Which is, `putAlias`? I just moved this one down, if you are talking about the fact that it invalidates? We definitely have to rebuild contracts after this...which agree we could optimize all these invalidations but can def follow that one on

plugins/deployment_manager/DeploymentManager.ts
jflatow: I guess we weren't doing it totally consistently, so I just picked a style. But if we wanted to lint that everything should have the `await` I'm not really opposed

jflatow: Also there were some cases where it's actually important, so I was touching all the ones and making them consistent as I encountered the functions

plugins/deployment_manager/Spider.ts
jflatow: Yeah I didn't really modify the interface (except adding the roots to return value). But it does basically rewrite the internal loop. There were a few issues before, and if you follow my WIP branches you'll see where I hit the main issue, when I was trying to preserve the way it was written and eventually couldn't and had to rewrite. I believe the killer was the fact that it was tracking visited by address, and not by alias, and there was a need to get a different interface for the same address under a different alias. The alias render stuff I had added before that to support the dynamic naming of `${symbol}:priceFeed`, and also the `Ctx` idea, and `artifact`. I added path after the rewrite (although attempted to add it earlier) to support the `${symbol}:admin` on FiatTokenProxy for crawling testnet aliases.

plugins/deployment_manager/Spider.ts
jflatow: The fact that its modding? This is just for the case where the array of alias templates is a different length than the field addresses. If there's a single alias template, we already relied on using that for all of them, which this still supports, it just now treats the alias list as a cycle. We could probably just throw an error if they are different lengths and aliases is not length 1, but I figured the cycle is less code and *could* actually be useful theoretically

plugins/deployment_manager/Spider.ts
```diff
@@ -1,211 +1,265 @@
...
+  relations: RelationConfigMap,
+  node: DiscoverNode,
+  context: Ctx,
+  aliases: Aliases,
+  proxies: Proxies,
+  contracts: ContractMap,
+): Promise<Alias> {
+  const { aliasRender, address, path } = node;
+  const { template: aliasTemplate } = aliasRender;
+  //debug(`Crawling ${address}`, aliasRender);
```
jflatow, line 102: Uncomment all these `debug`s if you are trying to follow what's happening for some reason and it'll all be super clear

plugins/deployment_manager/DeploymentManager.ts
hayesgm: I guess I tend to feel like it erases type information, since it's not obvious that the return value is, in fact, a promise. E.g.

```ts
async f() {
  return g();
}
```

```ts
async f() {
  return await g();
}
```

It's more obvious to me that in the latter `g` is a async function or returns a promise, where the in the top one, it's not exceedingly obvious. As far as I know, these shouldn't be a case where the result is different, since awaits are synatic sugar and, I believe, should be erased when they end a function.

plugins/deployment_manager/DeploymentManager.ts
hayesgm: Yeah, I'm just arguing that we could do better than invalidating here every time.

plugins/deployment_manager/DeploymentManager.ts
hayesgm: yeah `idempotentDeploy` might be nice, since it's a bit more descriptive to what this is?

plugins/deployment_manager/DeploymentManager.ts
hayesgm: Just looking into this, it looks like this will become (respecitvely): https://tc39.es/ecma262/#await

```ts
function f() {
  return Promise.resolve(g());
}
```

```ts
function f() {
  return Promise.resolve(g()).then((x) => {
    return Promise.resolve(x);
  }
}
```

These should be functionally equivalent, though the latter may be slightly slower depending on the quality of the JIT compiler.

plugins/deployment_manager/Spider.ts
hayesgm: Yeah, but I thought before we did a cross product, versus a cycle-list. I think this changes behavior pretty substantially, though maybe for the better.

plugins/deployment_manager/Spider.ts
hayesgm: This marks the third rewrite of Spider. A little confused how you track visited by alias, since we don't know the aliases until we actual spider the contracts, right?

plugins/deployment_manager/Spider.ts
jflatow: I think the cross product was just wrong? Commented about this in the discord channel and didn't get a response, but pair-wise wouldn't make sense right?

plugins/deployment_manager/Spider.ts
jflatow: That's the crux of why I had to rewrite, we put the contract in by alias after we visit it, but we do have to do a little bit of work whenever we visit a node regardless to figure out how to proceed

plugins/deployment_manager/test/DeploymentManagerTest.ts
jflatow: Nobody should be calling it now on DeploymentManager (and I removed it from DeploymentManager itself). We still have the tests on roots, which DM uses directly in only one place (after it spiders).

deployments/fuji/usdc/configuration.json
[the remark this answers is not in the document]
jflatow: Added COMP for #521 but didn't add as an asset to Fuji yet...I think we can revisit Fuji final config when we come back to Avalanche overall

plugins/deployment_manager/Spider.ts
hayesgm: If I recall correctly, and it's been a few months, but the idea was that you could provide multiple _ways_ of aliasing something, e.g. by fixed name (like "baseToken") _and_ by something like `asset.symbol()`. So in that case, cross-product made sense. I guess my question is: did we diverge in how we write `relations.ts` that makes zip more appropriate than cross?

jflatow: I think `Nx1` cross makes sense but that's kind of a degen case and also covered by the cycle behavior if you have e.g. array of `N` addresses from `field` with `1` function for computing `alias`

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -122,9 +125,19 @@ export class DeploymentManager {
     this.contractsCache = null;
   }
 
+  /* Conditionally executes an action */
+  async idempotent<T>(
```
kevincheng96, line 129: IMO, feels a bit odd to call this fn `idempotent`. I understand it's main use case is to support idempotent deploys, but this is more of a generic helper to conditionally execute some function. Maybe we could name it `conditionalRetry`? 

e.g.
```
await deploymentManager.conditionalRetry(
  async () => !sameAddress($configuratorImpl, configuratorImpl.address),
  async () => {
    debug(`Setting Configurator implementation to ${configuratorImpl.address}`);
    await wait(cometAdmin.connect(admin).upgrade(configurator.address, configuratorImpl.address));
  }
);
```

vs

```
await deploymentManager.idempotent(
  async () => !sameAddress($configuratorImpl, configuratorImpl.address),
  async () => {
    debug(`Setting Configurator implementation to ${configuratorImpl.address}`);
    await wait(cometAdmin.connect(admin).upgrade(configurator.address, configuratorImpl.address));
  }
);
```

Seems somewhat clearer what's happening

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -133,46 +146,105 @@ export class DeploymentManager {
...
+    if (!deployFn || !deployFn.call) {
+      throw new Error(`Missing deploy function in ${deployScript}.`);
     }
+    const deployed = await deployFn(this, deploySpec);
+    const newCount = this.counter;
+    const newSpider = await this.spider(deployed);
+    return {
+      old: { count: oldCount, spider: oldSpider },
+      new: { count: newCount, spider: newSpider }
+    };
```
kevincheng96, line 247: Wow this is very neat! Just tried it out locally and being able to see the diffs in Spider is sweet

deployments/fuji/usdc/configuration.json
kevincheng96: Probably redeploy right? I think you're doing that with the new gov anyways?

deployments/kovan/usdc/deploy.ts
```diff
@@ -1,251 +1,176 @@
...
-    deploymentManager,
-    { all: true },
-    {},
-    contracts
-  );
+  const deployed = await deployComet(deploymentManager, deploySpec);
+  const { comet } = deployed;
 
   // Deploy Bulker
-  const bulker = await deploymentManager.deploy<Bulker, Bulker__factory, [string, string, string]>(
```
kevincheng96, line 172: We should wrap this around a retry as well right?

deployments/kovan/usdc/deploy.ts
```diff
@@ -1,251 +1,176 @@
...
-
-export default async function deploy(deploymentManager: DeploymentManager) {
-  return await deploymentManager.doThenVerify(
-    () => deployAll(deploymentManager)
-  );
-}
-
-async function deployAll(deploymentManager: DeploymentManager) {
-  const newRoots = await deployContracts(deploymentManager);
-  deploymentManager.putRoots(new Map(Object.entries(newRoots)));
```
kevincheng96, line 42: Curious where roots are set now? Is it implicitly being set?

deployments/kovan/usdc/deploy.ts
```diff
@@ -1,251 +1,176 @@
...
-  configurator: string,
-  fauceteer: string,
-  rewards: string,
-  bulker: string
-};
-
-export default async function deploy(deploymentManager: DeploymentManager) {
-  return await deploymentManager.doThenVerify(
-    () => deployAll(deploymentManager)
-  );
```
kevincheng96, line 37: We still need this to lazy verify the contracts. I don't think the current script verifies the contracts. At least it wasn't when I ran it locally.

deployments/kovan/usdc/deploy.ts
```diff
@@ -1,251 +1,176 @@
...
+        signer.address, // Blacklister
+        signer.address  // Owner
+      ));
+    }
+  );
+
+  const WBTC = await deploymentManager.clone('WBTC', cloneAddr.wbtc, [], cloneNetwork);
+  const WETH = await deploymentManager.clone('WETH', cloneAddr.weth, [], cloneNetwork);
+  const COMP = await deploymentManager.clone('COMP', cloneAddr.comp, [signer.address], cloneNetwork);
+  const LINK = await deploymentManager.clone('LINK', cloneAddr.link, [], cloneNetwork);
```
kevincheng96, line 81: Let me know if I'm understanding this correctly. The reason that we no longer have to pass in this set of contracts to `deployComet` is because idempotent deploys can now grab these directly from `aliases`?

kevincheng96 (commented):
Nice! Seems to be working pretty flawlessly when trying it locally with different combinations of roots.

I have to admit, some of the spider changes are kind of going over my head, so would you mind summarizing real quick what has changed there?

src/deploy/index.ts
```diff
@@ -35,30 +30,11 @@ export interface ProtocolConfiguration {
...
-  governor: GovernorSimple;
-  rewards: CometRewards;
-  tokens?: ERC20[];
+// If `all` is specified, it takes precedence.
+// Other options are independent of one another.
+export interface DeploySpec {
+  all?: boolean;       // Re-deploy everything (including proxies and proxy admin)
+  cometMain?: boolean; // Re-deploy the main interface (config impl + comet factory + comet impl)
+  cometExt?: boolean;  // Re-deploy the ext interface (comet ext)
+  rewards?: boolean;   // Re-deploy the rewards contract
```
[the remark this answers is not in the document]
jflatow: These are basically the 'flows' -- but perhaps you are suggesting we should also have a simpler one just for pure param changes, which is probably a good idea but we only really use 2 of these from tests now and I think having too many will be harder to test

plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -133,46 +146,105 @@ export class DeploymentManager {
...
+      const contract: C = await this._deployBuild(buildFile, deployArgs, retries);
+      await this.putAlias(alias, contract.address);
+      return contract;
+    }
+    return maybeExisting;
+  }
+
+  /* Conditionally deploy a contract with its alias if it does not exist, or if forced */
+  async deploy<C extends Contract, DeployArgs extends Array<any>>(
+    alias: Alias,
```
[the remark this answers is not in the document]
jflatow: Yeah but that's why you `--simulate` first and/or run scens to see what's actually going to be done. I had some concerns about it at first too but with simulate I think it works amazingly well

plugins/deployment_manager/DeploymentManager.ts
jflatow: Yeah it could also be called `maybe`, tbh I didn't think about the naming of this one at all

jflatow: I suggest we do another stylistic search and replace after this if we can agree about some other names for things

deployments/kovan/usdc/deploy.ts
jflatow: Deploy retries

deployments/kovan/usdc/deploy.ts
jflatow: Only after you spider

deployments/kovan/usdc/deploy.ts
jflatow: Oh but it does, in the `deploy` task

deployments/kovan/usdc/deploy.ts
jflatow: Yes they are all aliased

deployments/kovan/usdc/deploy.ts
kevincheng96: I don't see verify being called in the task. It's just setting verification to `lazy`, but not actually calling `verify` or using `doThenVerify` anywhere.

jflatow: Oh see `clone-gov` branch, I basically am keeping development going in a line now, can't maintain these 3 branches -- we can merge `clone-gov` to `main` I think when ready (or back them into this one, but what's the point, really

plugins/deployment_manager/DeploymentManager.ts
jflatow: Oh man you need to try `clone-gov` branch, it has a bunch of new fixes/features!

deployments/kovan/usdc/deploy.ts
kevincheng96: Oh I see. Didn't finish reviewing that branch yet

deployments/fuji/usdc/configuration.json
jflatow: Yeah I did redeploy I guess, let's worry about realistic `fuji` settings later though, we can easily redeploy again (or migrate)

jflatow:
> Nice! Seems to be working pretty flawlessly when trying it locally with different combinations of roots.
> 
> I have to admit, some of the spider changes are kind of going over my head, so would you mind summarizing real quick what has changed there?

We can discuss more live (bc there's a lot, and some of it's already hard to remember), but summary I think is:
 - visited nodes tracked by alias not address (which fixes a bug that we couldn't crawl an alias correctly if we had already visited the address before)
  - lots more context for the alias/field functions to access
  - dynamic aliases to support things like `USDC:priceFeed` which were impossible to do before
  - ability to hardcode artifacts that should be used for an alias template or contract name (which fixes an issue that it was impossible to call certain functions on proxies from certain functions, and also solves dealing with unverified contracts, and *could* also be used to decrease dependency on etherscan overall)
  -  re-organization of when we compute aliases and stricter distinction of the differences b/w all these things in order to support all of the above

jflatow:
Rebased and including in #521

PR #519, opened 2022-08-09
Fix empty solutions being returned by proposal constraint

kevincheng96:
This bug is causing all our mainnet scenarios to run with 0 solutions (basically mainnet scenarios are skipped).

### DIFF
--- a/scenario/constraints/ProposalConstraint.ts
+++ b/scenario/constraints/ProposalConstraint.ts
@@ -87,7 +87,7 @@ export class ProposalConstraint<T extends CometContext, R extends Requirements> 
       });
     }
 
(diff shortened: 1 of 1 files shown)


scenario/constraints/ProposalConstraint.ts
```diff
@@ -87,7 +87,7 @@ export class ProposalConstraint<T extends CometContext, R extends Requirements>
       });
     }
 
-    return solutions;
+    return solutions.length > 0 ? solutions : null;
```
jflatow, line 90: 👍

jflatow (approved):
🎣

PR #520, opened 2022-08-09
Make the ERC20 name settable by the ext constructor/config

jflatow:
Also updates the names/symbols for real deployments.

### DIFF
--- a/contracts/CometExt.sol
+++ b/contracts/CometExt.sol
@@ -15,9 +15,6 @@ contract CometExt is CometExtInterface {
 
     /** Public constants **/
 
-    /// @notice The name of this contract
-    string public override constant name = "Compound Comet";
-
     /// @notice The major version of this contract
     string public override constant version = "0";
 
@@ -35,6 +32,9 @@ contract CometExt is CometExtInterface {
 
     /** Immutable symbol **/
 
+    /// @dev The ERC20 name for wrapped base token
+    bytes32 internal immutable name32;
+
     /// @dev The ERC20 symbol for wrapped base token
     bytes32 internal immutable symbol32;
 
@@ -43,6 +43,7 @@ contract CometExt is CometExtInterface {
      * @param config The mapping of initial/constant parameters
      **/
     constructor(ExtConfiguration memory config) {
+        name32 = config.name32;
         symbol32 = config.symbol32;
     }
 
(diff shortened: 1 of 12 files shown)


contracts/CometExt.sol
```diff
@@ -72,6 +73,26 @@ contract CometExt is CometExtInterface {
 
     /** Additional ERC20 functionality and approval interface **/
 
+    /**
+     * @notice Get the ERC20 name for wrapped base token
+     * @return The name as a string
+     */
+    function name() override public view returns (string memory) {
+        uint8 i;
```
kevincheng96, line 81: I think `uint256` should be more gas and space efficient since it avoids an extra downscale

contracts/CometExt.sol
```diff
@@ -72,6 +73,26 @@ contract CometExt is CometExtInterface {
...
+        for (i = 0; i < 32; ) {
+            if (name32[i] == 0) {
+                break;
+            }
+            unchecked { i++; }
+        }
+        bytes memory name_ = new bytes(i);
+        for (uint8 j = 0; j < i; ) {
```
kevincheng96, line 89: Same here

kevincheng96 (approved):
Looks good

contracts/CometExt.sol
jflatow: Hm I just copied what we had for `symbol`, you think I should change both now?

jflatow: I'm not sure that it actually has to downscale if its all in memory (but who knows maybe it does), anyway I say we keep them as is in the name of correct intention / smaller change. Plus I'm re-deploying `kovan` already now... 😬

contracts/CometExt.sol
jflatow: See above

contracts/CometExt.sol
kevincheng96: Yeah not a big deal, let's keep it as is

hayesgm (commented):
looks good to me.

kevincheng96:
Btw, I think scenarios might be failing because Hardhat isn't able to recognize the custom errors in `CometExt` now that the contract differs from what's deployed on-chain (hope they fix this bug soon). The fix is to use the modern constraint to re-deploy `CometExt` for those failing scenarios

jflatow:
> Btw, I think scenarios might be failing because Hardhat isn't able to recognize the custom errors in `CometExt` now that the contract differs from what's deployed on-chain (hope they fix this bug soon). The fix is to use the modern constraint to re-deploy `CometExt` for those failing scenarios

Hm..maybe? Although I believe I had run it locally (but maybe I only tried dev). Anyway #521 has this change and its passing now, but maybe just because `roots.json` are wiped

scott-silver (approved):
👍

jflatow:
Rebased and including in #521

PR #521, opened 2022-08-10
Clone gov contracts as much as possible from mainnet

jflatow:


scott-silver (approved):
🥳

.gitignore
```diff
@@ -10,9 +10,7 @@ artifacts/
 build/
 cache/
 dist/
-deployments/*/*/artifacts
-deployments/*/*/cache
-deployments/*/*/contracts
+deployments/.contracts
```
scott-silver, line 13: do we need to update the files deleted in `yarn spider --clean`, too?

jflatow: Probably, didn't know about that command -- where is it?

scenario/context/CometContext.ts
```diff
@@ -215,73 +206,52 @@ export class CometContext {
...
+        signatures,
+        calldatas,
+        'FastExecuteProposal',
+        { gasPrice: 0 }
+      )
+    ).wait();
+    const proposeEvent = proposeTxn.events.find(event => event.event === 'ProposalCreated');
+    const [proposalId, , , , , , startBlock, endBlock] = proposeEvent.args;
+
+    await this.executePendingProposal(proposalId, startBlock, endBlock);
```
kevincheng96, line 254: 🧼

kevincheng96 (approved):
I reviewed the diffs between this PR and #517 and everything LGTM 👍

.gitignore
scott-silver: https://github.com/compound-finance/comet/blob/main/tasks/spider/task.ts#L5

kevincheng96, line 13: So contracts for all chains will live in the same directory? What happens if two contracts share the same address on different chains?

jflatow: I mean they can't, the addresses are deterministic

jflatow: If they share an address they are the same contract

kevincheng96: true

jflatow: Will follow on with a remove proxies PR hat touches this file, this is getting to large

PR #524, opened 2022-08-10
Liquidation Bot deploy script

scott-silver:
```
RECIPIENT=0xabc DEPLOYMENT=usdc yarn hardhat run scripts/liquidation_bot/deploy.ts --network kovan
```

Newly-deployed liquidator.sol:

https://kovan.etherscan.io/address/0xB72dA9264D2621AAa98c5674857f8b596158B2E4#code

### DIFF
--- a/scripts/liquidation_bot/deploy.ts
+++ b/scripts/liquidation_bot/deploy.ts
@@ -14,7 +14,7 @@ async function main() {
   throwIfMissing(RECIPIENT, 'missing required env variable: RECIPIENT');
 
   const network = hre.network.name;
-  if (["hardhat", "fuji"].includes(network)) {
+  if (['hardhat', 'fuji'].includes(network)) {
     throw new Error(`Uniswap unavailable on network: ${network}`);
   }
 
--- a/scripts/liquidation_bot/deploy.ts
+++ b/scripts/liquidation_bot/deploy.ts
@@ -34,6 +34,7 @@ async function main() {
   const assets = await Promise.all(Array(numAssets).fill(0).map((_, i) => comet.getAssetInfo(i)));
 
   const liquidator = await dm.deploy(
+    'liquidator',
     'liquidator/Liquidator.sol',
     [
       RECIPIENT, // _recipient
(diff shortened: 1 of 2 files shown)


scripts/liquidation_bot/deploy.ts
```diff
@@ -0,0 +1,60 @@
...
+  const liquidator = await dm.deploy(
+    'liquidator/Liquidator.sol',
+    [
+      RECIPIENT, // _recipient
+      SWAP_ROUTER, // _swapRouter
+      comet.address, // _comet
+      UNISWAP_V3_FACTORY_ADDRESS, // _factory
+      WETH9, // _WETH9
+      0, // _liquidationThreshold,
+      assets.map(a => a.asset), // _assets
```
hayesgm, line 46: To be clear, we need to deploy a new liquidator if say there's a new asset added to Comet?

hayesgm (approved):
looks nice

scripts/liquidation_bot/deploy.ts
jflatow: Great ❓

scott-silver: shouldn't be required; the contract will just fallback to the default config:

https://github.com/compound-finance/comet/blob/main/contracts/liquidator/Liquidator.sol#L118

scripts/liquidation_bot/deploy.ts
```diff
@@ -34,6 +34,7 @@ async function main() {
   const assets = await Promise.all(Array(numAssets).fill(0).map((_, i) => comet.getAssetInfo(i)));
 
   const liquidator = await dm.deploy(
+    'liquidator',
```
jflatow, line 37: Oh interesting this is using the deployment manager but not a deployment 🤔

scott-silver: it uses a deployment, right? it needs to know `network` + `deployment` to find the correct instance of Comet

jflatow: So the deployment is in `scripts/liquidation_bot`? I guess the deploy script can be anywhere if its not a base, just hadn't really thought about this. So you can't run sims on it but you can add to roots/aliases 🤯 Kinda cool I think it all works out with the cli tool too because roots is always updated non-destructively now

jflatow (approved):

PR #527, opened 2022-08-11
Make contracts cached per network properly

jflatow:
I was wrong to think contract addresses are always a function of bytecode.
This adds caching per network the right way, being aware of the (true) network we are cloning from or deploying to.
This was a bit easier to add with proxies out of the way.


jflatow:
Merged #526 into the wrong branch

PR #528, opened 2022-08-11
Fix test, enable proposal constraint for all networks

antonina-cherednichenko:
- Fix occasional rewards test fail 
- Enable proposal constraint for all testnets

### DIFF
--- a/scenario/RewardsScenario.ts
+++ b/scenario/RewardsScenario.ts
@@ -58,10 +58,12 @@ scenario(
 
     const supplySpeed = (await comet.baseTrackingSupplySpeed()).toBigInt();
     const trackingIndexScale = (await comet.trackingIndexScale()).toBigInt();
-    const expectedRewardsOwed = calculateRewardsOwed(albertBalance, totalSupplyBalance, supplySpeed, 86400, trackingIndexScale, rewardScale, rescaleFactor.toBigInt());
-    const expectedRewardsReceived = calculateRewardsOwed(albertBalance, totalSupplyBalance, supplySpeed, 86400 + timeElapsed, trackingIndexScale, rewardScale, rescaleFactor.toBigInt());
+    const timestampDelta = preTxnTimestamp - supplyTimestamp;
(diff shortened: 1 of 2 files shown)


scenario/RewardsScenario.ts
```diff
@@ -111,10 +111,12 @@ scenario(
...
     const supplySpeed = (await comet.baseTrackingSupplySpeed()).toBigInt();
     const trackingIndexScale = (await comet.trackingIndexScale()).toBigInt();
-    const expectedRewardsOwed = calculateRewardsOwed(albertBalance, totalSupplyBalance, supplySpeed, 86400, trackingIndexScale, rewardScale, rescaleFactor.toBigInt());
-    const expectedRewardsReceived = calculateRewardsOwed(albertBalance, totalSupplyBalance, supplySpeed, 86400 + timeElapsed, trackingIndexScale, rewardScale, rescaleFactor.toBigInt());
+    const timestampDelta = preTxnTimestamp - supplyTimestamp;
+    const expectedRewardsOwed = calculateRewardsOwed(albertBalance, totalSupplyBalance, supplySpeed, timestampDelta, trackingIndexScale, rewardScale, rescaleFactor.toBigInt());
+    const expectedRewardsReceived = calculateRewardsOwed(albertBalance, totalSupplyBalance, supplySpeed, timestampDelta + timeElapsed, trackingIndexScale, rewardScale, rescaleFactor.toBigInt());
 
-    expect(preTxnTimestamp - supplyTimestamp).to.be.equal(86400);
+    // Occasionally `timestampDelta` is equal to 86401
```
jflatow, line 118: 💪 Hm ok but not a perfect fix, but at least confirms what the problem is. I was trying to add a function last night to help us *precisely* control time which I think is still a good idea to avoid some of these gymnastics

jflatow (approved):

scenario/RewardsScenario.ts
jflatow: Actually, before merging, if we don't add the function, I think we also need to add the fix to the other version of this in this file

antonina-cherednichenko: good point, I added that

PR #510, opened 2022-08-04
Migration for initial mainnet proposal

kevincheng96:
This is the migration script to create the initial mainnet proposal for:

1. Raising supply caps
2. Setting a non-zero borrow COMP speed
3. Seeding Comet reserves with some USDC
4. Streaming COMP to the CometRewards contract

This PR also sets the initial mainnet deployment's supply caps (in `configuration.json`) to 0.

Note: This is tested via the migration and proposal constraints and the numbers for supply caps and USDC reserves are still subject to change.

PR #529, opened 2022-08-11
Separate final fixes and cleanup for mainnet proposal branch

jflatow:
Rebase/cleanup mainnet proposal and get ready to deploy

Remove unnecessary testnet migrations

We had migrations to seed COMP on testnet (where it exists), but we now do it as part of deploy.
In general the correct principal for what should be a migration seems to be, only things that *must* be done as proposals.
In particular, changes made after a live deployment is handed over to governance.
Migrations should be deleted after the proposal passes and the PR with tests is merged.
This seems like the best policy, because migrations and proposals are expensive, even with potential improvements they combinatorially  multiply the number of scenarios we need to run.
So since we can seed rewards initially, we do.

A little more cleanup of utils


plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -310,6 +310,13 @@ export class DeploymentManager {
     return this.contractsCache;
   }
 
+  /* Gets all the contracts, connected to signer, as an object */
+  async getContracts(signer?: SignerWithAddress): Promise<{ [alias: Alias]: Contract }> {
```
scott-silver, line 314: following on from: https://github.com/compound-finance/comet/pull/510#discussion_r943935518

whatever the solution, it seems like we should avoid having two functions called `contracts` and `getContracts` that do very similar, but importantly-different things

jflatow: You should just use `getContracts`, and we'll merge them soon, this will be an easy fix just takes a little time that we don't have right now

jflatow: `getContracts` is really just more convenient though so scripts can do e.g. `const { configurator, comet } = await deploymentManager.getContracts()` I don't think there's any trap here

plugins/scenario/worker/Report.ts
```diff
@@ -21,49 +21,57 @@ export function pluralize(n, singular, plural = null) {
...
-  let avgTime = testCount > 0 ? totalTime / testCount : 0;
-  let succText = pluralize(succCount, 'success', 'successes');
-  let errText = pluralize(errCount, 'error', 'errors');
-  let skipText = pluralize(skipCount, 'skipped');
-  let avgText = `[avg time: ${avgTime.toFixed(0)}ms]`;
-
-  console.log(`\n\n${prefix} Results: ${succText}, ${errText}, ${skipText} ${avgText}\n`);
+  console.log('\n\n');
+  for (const [base, stats] of Object.entries(statsPer)) {
+    const prefix = stats.errCount === 0 ? '✅' : '❌';
```
scott-silver, line 66: image

looks great!

scott-silver (approved):

PR #532, opened 2022-08-12
Add tracer to DeploymentManager to capture spend and improve logging

jflatow:
This replaces debug calls with calls to a tracer from the DeploymentManager.
The trace function makes logging more consistent with network names and transaction response support, which allows us to capture a detailed spend history and report it on the cli.

### DIFF
--- a/plugins/deployment_manager/Deploy.ts
+++ b/plugins/deployment_manager/Deploy.ts
@@ -1,28 +1,39 @@
 import * as path from 'path';
 import * as fs from 'fs/promises';
 
-import { Contract, Signer } from 'ethers';
+import { Contract, ContractFactory, Signer } from 'ethers';
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
 
 import { putVerifyArgs } from './VerifyArgs';
 import { Cache } from './Cache';
 import { storeBuildFile } from './ContractMap';
-import { BuildFile } from './Types';
+import { BuildFile, TraceFn } from './Types';
 import { debug, getPrimaryContract } from './Utils';
 import { VerifyArgs, verifyContract, VerificationStrategy } from './Verify';
 
-export abstract class Deployer<Contract, DeployArgs extends Array<any>> {
-  abstract connect(signer: Signer): this;
-  abstract deploy(...args: DeployArgs): Promise<Contract>;
-}
-
 export interface DeployOpts {
+  cache?: Cache; // caches the build file, if included
+  connect?: Signer; // signer for the returned contract
   network: string; // the real name of the network we would actually or hypothetically deploy on
   overwrite?: boolean; // should we overwrite existing contract link
-  connect?: Signer; // signer for the returned contract
-  verificationStrategy?: VerificationStrategy; // strategy for verifying contracts on etherscan
   raiseOnVerificationFailure?: boolean; // if verification is considered critical
-  cache?: Cache; // caches the build file, if included
+  trace?: TraceFn; // the trace fn to use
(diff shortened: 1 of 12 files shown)


plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -388,6 +389,23 @@ export class DeploymentManager {
...
+  tracer(): TraceFn {
+    return (first, ...rest) => {
+      if (typeof first === 'string') {
+        debug(`[${this.network}] ${first}`, ...rest);
+      } else {
+        first.wait().then(async (tx) => {
+          const cost = Number(txCost(tx) / (10n ** 12n)) / 1e6;
+          const logs = tx.events.map(e => `${e.event}(${e.args})`).join(' ');
+          const info = `@ ${tx.transactionHash}[${tx.transactionIndex}]`;
+          const desc = `${info} in blockNumber: ${tx.blockNumber} logs: ${logs}`;
```
scott-silver, line 401: nit: maybe `events:` instead of `logs:`?

seeing a few `logs: undefined(undefined)` here and there

image

jflatow: Yeah saw those, that means we don't have the abi, I think maybe still worth knowing?

deployments/kovan/usdc/roots.json
```diff
@@ -1,7 +0,0 @@
-{
-    "comet": "0xa7D85950E6E1bB7115c626cBB388Fa0a8C927c1c",
```
scott-silver, line 2: did you mean to delete the Kovan roots?

plugins/deployment_manager/DeploymentManager.ts
scott-silver: that makes sense, since in this case we're working on a clone of LINK

deployments/kovan/usdc/roots.json
jflatow: No good catch, just meant to do that to test Import logs

plugins/deployment_manager/Deploy.ts
```diff
@@ -32,15 +43,11 @@ async function deployFromBuildFile<C extends Contract>(
...
   const [ethersSigner] = await hre.ethers.getSigners();
   const signer = deployOpts.connect ?? ethersSigner;
-  const contractFactory = new hre.ethers.ContractFactory(metadata.abi, metadata.bin, signer);
-  debug(`Deploying ${Object.keys(buildFile.contracts)} with args`, deployArgs);
-  const contract = await contractFactory.deploy(...deployArgs);
-  const deployed = await contract.deployed();
-  debug(`Deployed ${contractName} @ ${contract.address}`);
-  return deployed as C;
+  const factory = new hre.ethers.ContractFactory(metadata.abi, metadata.bin, signer);
+  return doDeploy(contractName, factory, deployArgs, deployOpts, 'build file');
```
scott-silver, line 50: are the types on the returned contract the same as before?

does this need to be `doDeploy<C>(`?

jflatow: Yeah still C extends contract and nah it figures it out from the signatures

scott-silver (approved):
super useful!

hayesgm (approved):
This is cool, but funny didn't we literally _just_ remove `dm.debug()` thinking it was redundant when it's basically this?

jflatow:
> This is cool, but funny didn't we literally _just_ remove `dm.debug()` thinking it was redundant when it's basically this?

One pass at a time ;)

PR #533, opened 2022-08-12
Make getAssetIndex public, fix bumpSupplyCaps

jflatow:
This makes a contract change to configurator, to make `getAssetIndex` public.
It was useful to be able to call this directly when trying to debug these  issues and I think it ought to be public.

Somehow the upgrade path in bumpSupplyCaps was completely broken and had never been used (at least since adding the per-market configuration).
This fixes the call to updateSupplyCap and bypasses governance for this constraint helper to speed it up.

Also remove some whales, since we are voting with all of them and its unnecessary.

### DIFF
--- a/scenario/context/CometContext.ts
+++ b/scenario/context/CometContext.ts
@@ -114,20 +114,12 @@ export class CometContext {
 
   async bumpSupplyCaps(supplyAmountPerAsset: Record<string, bigint>) {
     const comet = await this.getComet();
-    const configurator = await this.getConfigurator();
-    const proxyAdmin = await this.getCometAdmin();
 
     // Update supply cap in asset configs if new collateral supply will exceed the supply cap
     let shouldUpgrade = false;
     const newSupplyCaps: Record<string, bigint> = {};
     for (const asset in supplyAmountPerAsset) {
-      let assetInfo: AssetInfoStructOutput;
-      try {
-        assetInfo = await comet.getAssetInfoByAddress(asset);
-      } catch (e) {
-        continue; // skip if asset is not a collateral asset
-      }
-
+      const assetInfo = await comet.getAssetInfoByAddress(asset);
       const currentTotalSupply = (await comet.totalsCollateral(asset)).totalSupplyAsset.toBigInt();
       const newTotalSupply = currentTotalSupply + supplyAmountPerAsset[asset];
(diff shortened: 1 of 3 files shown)


scenario/context/CometContext.ts
```diff
@@ -138,18 +130,13 @@ export class CometContext {
...
     // Set new supply caps in Configurator and do a deployAndUpgradeTo
     if (shouldUpgrade) {
-      const [targets, values, signatures, calldata] = [[], [], [], []];
-      for (const asset in newSupplyCaps) {
-        targets.push(configurator.address);
-        values.push(0);
-        signatures.push('updateAssetSupplyCap(address,uint128)');
-        calldata.push(utils.defaultAbiCoder.encode(['address', 'uint128'], [asset, newSupplyCaps[asset]]))
+      const admin = await this.world.impersonateAddress(await comet.governor(), 10n**18n);
+      const cometAdmin = (await this.getCometAdmin()).connect(admin);
```
scott-silver, line 134: not necessary for this PR, but we should rename this function to `getCometProxyAdmin`; it's confusing that we use the term admin to refer to both signers and contracts

scott-silver (approved):

scenario/context/CometContext.ts
jflatow: oh true, we should just call the actor `governor` then

jflatow: Fixed

scenario/context/CometContext.ts
```diff
@@ -272,9 +259,9 @@ export class CometContext {
     const queueEvent = queueTxn.events?.find(event => event.event === 'ProposalQueued');
     await this.setNextBlockTimestamp(queueEvent?.args?.eta.toNumber() + 1);
 
-    // Execute proposal
+    // Execute proposal (w/ gas limit so we see if exec reverts, not a gas estimation error)
```
kevincheng96, line 262: Oh wow, this is going to greatly improve debugging

jflatow: Yeah now it says "Execution reverted with Timelock" instead of UNPREDICTABLE GAS LIMIT 😸  But overall debugging should be a lot simpler now with some of the latest improvements to tracing

PR #534, opened 2022-08-12
Add a `safeSupplyAsset` instead of bumping everywhere first

jflatow:
We often get the bumping wrong, as in the case of borrowing.
This also makes bumping slightly more efficient and correct, since it doesn't allow arbitrary unknown assets, but just ignores the case of base token.
With this, mainnet scenarios pass even without the proposal!

### DIFF
--- a/scenario/context/CometContext.ts
+++ b/scenario/context/CometContext.ts
@@ -114,17 +114,20 @@ export class CometContext {
 
   async bumpSupplyCaps(supplyAmountPerAsset: Record<string, bigint>) {
     const comet = await this.getComet();
+    const baseToken = await comet.baseToken();
 
     // Update supply cap in asset configs if new collateral supply will exceed the supply cap
     let shouldUpgrade = false;
     const newSupplyCaps: Record<string, bigint> = {};
     for (const asset in supplyAmountPerAsset) {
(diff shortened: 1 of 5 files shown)


scott-silver (approved):
🙌

scenario/context/CometContext.ts
```diff
@@ -114,17 +114,20 @@ export class CometContext {
...
     let shouldUpgrade = false;
     const newSupplyCaps: Record<string, bigint> = {};
     for (const asset in supplyAmountPerAsset) {
-      const assetInfo = await comet.getAssetInfoByAddress(asset);
-      const currentTotalSupply = (await comet.totalsCollateral(asset)).totalSupplyAsset.toBigInt();
-      const newTotalSupply = currentTotalSupply + supplyAmountPerAsset[asset];
-      if (newTotalSupply > assetInfo.supplyCap.toBigInt()) {
-        shouldUpgrade = true;
-        newSupplyCaps[asset] = newTotalSupply * 2n;
+      if (asset != baseToken) {
```
hayesgm, line 123: `!==`?

jflatow: I suppose... I probably feel the way about these the way you do about `const`

jflatow: Will fix in next patch

PR #535, opened 2022-08-12
prefer const

scott-silver:
https://eslint.org/docs/latest/rules/prefer-const

### DIFF
--- a/test/helpers.ts
+++ b/test/helpers.ts
@@ -220,7 +220,7 @@ export async function makeProtocol(opts: ProtocolOpts = {}): Promise<Protocol> {
   const signers = await ethers.getSigners();
 
   const assets = opts.assets || defaultAssets();
-  let priceFeeds = {};
+  const priceFeeds = {};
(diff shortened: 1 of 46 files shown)


hayesgm:
FWIW, I generally disagree with changing from `let` to `const`, primarily focused on `let` being 100x cooler sounding of a word than `const`. I generally think we've spent too many cycles already substituting one word for another that are 99% semantically the same.

test/helpers.ts
```diff
@@ -261,8 +261,7 @@ export async function makeProtocol(opts: ProtocolOpts = {}): Promise<Protocol> {
     const initial = config.initial || 1e6;
     const name = config.name || symbol;
     const factory = config.factory || FaucetFactory;
-    let token;
-    token = (tokens[symbol] = await factory.deploy(initial, name, decimals, symbol));
```
scott-silver, line 265: not exactly sure how we ended up with these strange lines, where we declare `token` and define it on the next line

scott-silver:
maybe someday 😞

PR #539, opened 2022-08-12
Simplify

hayesgm:


jflatow (approved):

.github/workflows/enact-migration.yaml
```diff
@@ -63,11 +63,8 @@ jobs:
...
-            export ETH_PK="${{ inputs.eth_pk }}"
-          else
-            export NETWORK_PROVIDER="http://localhost:8585"
-            export REMOTE_ACCOUNTS="true"
-          fi
-
           yarn hardhat migrate --network ${{ github.event.inputs.network }} --deployment ${{ github.event.inputs.deployment }} --enact --overwrite ${{ fromJSON('["", "--simulate"]')[github.event.inputs.simulate == 'true'] }} ${{ github.event.inputs.migration }}
+        env:
+          ETH_PK: "${{ inputs.eth_pk }}"
+          NETWORK_PROVIDER: ${{ fromJSON('["", "http://localhost:8585"]')[github.event.inputs.eth_pk == ''] }}
```
scott-silver, line 69: how does this work?

doesn't it evaluate to `["", "http://localhost:8585"][true]`?

did you mean to convert the second statement into an index?

```
github.event.inputs.eth_pk == '' ? 0 : 1
```

jflatow: I thought it was like an "Iverson bracket" based on other usage

hayesgm: This is copied from other examples, and super weird, but it's described here:

Screen Shot 2022-08-12 at 12 25 15 PM

hayesgm: So yes, the boolean is coerced to an integer and then used to index an array. This code is lifted from other examples of this weird pattern.

PR #540, opened 2022-08-15
Add caching for pending proposals

toninorair:
An approach that allows:
- fetching pending proposals once inside worker
- caching them 
- reading this info from cache for all scenarios.

### DIFF
--- a/plugins/scenario/worker/Worker.ts
+++ b/plugins/scenario/worker/Worker.ts
@@ -9,6 +9,7 @@ import { ScenarioConfig } from '../types';
 import { SimpleWorker } from './SimpleWorker';
 import hreForBase from '../utils/hreForBase';
 import { pluralize } from './Report';
+import { getAllPendingProposals, getProposalCacheId } from '../utils/proposals';
 
 interface Message {
   scenario?: {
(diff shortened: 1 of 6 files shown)


jflatow (commented):
So this assumes the slow part is getting all pending proposals? I tend to think its the running the migrations/executing them. Want to try on the mainnet proposal branch to benchmark?

antonina-cherednichenko:
> So this assumes the slow part is getting all pending proposals? I tend to think its the running the migrations/executing them. Want to try on the mainnet proposal branch to benchmark?

About 2 min difference

plugins/scenario/worker/Worker.ts
```diff
@@ -62,6 +63,11 @@ export async function run<T, U, R>({ bases, config, worker }: WorkerData) {
   for (const base of bases) {
     const world = new World(hreForBase(base), base), dm = world.deploymentManager;
     const delta = await dm.runDeployScript({ allMissing: true });
+
+    // Reading pending proposals and storing them in the cache for all scenarios
+    const proposals = await getAllPendingProposals(world);
+    dm.cache.storeCache(getProposalCacheId(base.network), proposals);
```
jflatow, line 69: The cache is stored across networks, so not sure we would want to do this without accounting for forking. Also, the plugin code is meant to be agnostic to Compound-specific stuff, so we wouldn't want it to be aware of proposals, right?

jflatow:
Closing as this is better addressed by #700

PR #541, opened 2022-08-15
Add claim rewards action to Bulker

kevincheng96:
This PR adds the ability to claim rewards using the Bulker contract.


contracts/ICometRewards.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+/**
+ * @dev Interface for interacting with the CometRewards contract
+ */
+interface ICometRewards {
```
kevincheng96, line 7: What naming convention should we use here? For most interfaces, we go with the `IContract` convention (e.g. `IComp`), but for Comet, we named it `CometInterface`.

contracts/Bulker.sol
```diff
@@ -18,15 +20,18 @@ contract Bulker {
...
     uint public constant ACTION_WITHDRAW_ETH = 5;
+    uint public constant ACTION_CLAIM_REWARD = 6;
 
     /** Custom errors **/
     error InvalidArgument();
     error FailedToSendEther();
     error Unauthorized();
 
-    constructor(address admin_, address comet_, address payable weth_) {
+    // XXX do we want to be constantizing the comet address and rewards address
```
kevincheng96, line 30: This question came up while working on this with 🥈. Do we want to make `Bulker` more generic by no longer making `comet` (and `rewards`) an immutable and instead, have it be passed in as calldata during `invoke`? This would allow us to use a single `Bulker` for multiple deployments on one chain. It also means we wouldn't have to re-deploy `Bulker` whenever the `rewards` contract changes.

contracts/Bulker.sol
```diff
@@ -137,4 +145,11 @@ contract Bulker {
         (bool success, ) = to.call{ value: amount }("");
         if (!success) revert FailedToSendEther();
     }
+
+    /**
+     * @notice Claim reward for a user
+     */
+    function claimReward(address src, bool shouldAccrue) internal {
+        ICometRewards(rewards).claim(comet, src, shouldAccrue);
```
kevincheng96, line 153: `claimTo` is more generic than `claim` but is more gassy due to the permission check. I think for our use-cases, `claim` is sufficient, but we could always add a new action for `claimTo` if we need it.

contracts/Bulker.sol
jflatow: Sounds like a good idea, redeploying for multiple instances would suck, especially if we had different versions of bulker

contracts/ICometRewards.sol
jflatow: We only need like one of these functions for Bulker right? Would it be terrible to just inline the interface we need at the top of Bulker and/or call it something like IClaimable?

jflatow (commented):
💪 Nice work, thanks for getting this off the ground!

contracts/ICometRewards.sol
kevincheng96: Yeah, I think inlining is the best probably. Will make that change

contracts/Bulker.sol
kevincheng96: Cool, I'll get rid of the XXX and start a new branch for this change.

kevincheng96: Created #542 for this

contracts/Bulker.sol
```diff
@@ -5,10 +5,17 @@ import "./CometInterface.sol";
 import "./ERC20.sol";
 import "./IWETH9.sol";
 
+interface IClaimable {
```
scott-silver, line 8: does the Rewards contract need to explicitly implement this interface now?

scott-silver (approved):

jflatow (approved):

contracts/Bulker.sol
kevincheng96: Nope, not explicitly, assuming you mean extending `IClaimable` directly in the Rewards contract. The only expectation is that the Rewards contract implements the functions defined in `IClaimable`, otherwise the call from Bulker will revert.

PR #543, opened 2022-08-17
Ad hoc scenario to test mainnet proposal

kevincheng96:
Quick and dirty scenario to test end states after executing the mainnet proposal.

One solution (with the executed proposal) passes and the other (without the executed proposal) fails.


kevincheng96:
Closing now that proposal 116 has been executed successfully.

PR #545, opened 2022-08-17
Some XXX fixes

toninorair:


test/tracking-index-bounds-test.ts
```diff
@@ -128,6 +128,66 @@ describe('total tracking index bounds', function () {
...
+      comet, tokens, users: [alice]
+    } = await makeProtocol({
+      base: 'USDC',
+      trackingIndexScale: 1e15,
+      baseTrackingSupplySpeed: 1e8, // supplySpeed=0.0000001 (1e-7) Comp/s
+    });
+    const { USDC } = tokens;
+
+    // allocate and approve transfers
+    await USDC.allocateTo(alice.address, 1e6);
```
kevincheng96, line 145: This actually doesn't accrue rewards because `totalSupplyBase < baseMinForRewards`. We should set `totalSupplyBase > baseMinForRewards` in this test to see the effects of a small supply/borrow when rewards are accruing.

kevincheng96 (commented):
Thanks for cleaning these up 👍

test/tracking-index-bounds-test.ts
jflatow: Indeed

PR #546, opened 2022-08-17
Run scenarios for all collateral assets; Introduce FilterConstraint

kevincheng96:
This PR increases the coverage of our supply/transfer/withdraw scenarios to all collateral assets (instead of just the first one). 

I also added a FilterConstraint to filter out solutions early on, before any other constraints that could revert are run.

### DIFF
--- a/scenario/SupplyScenario.ts
+++ b/scenario/SupplyScenario.ts
@@ -1,6 +1,83 @@
-import { scenario } from './context/CometContext';
+import { CometContext, scenario } from './context/CometContext';
 import { expect } from 'chai';
 import { expectApproximately, expectRevertMatches, getExpectedBaseBalance, getInterest } from './utils';
+import { ContractReceipt } from 'ethers';
+
+const NUM_ASSETS = 15;
+
+async function isValidAssetIndex(ctx: CometContext, num: number): Promise<boolean> {
+  const comet = await ctx.getComet();
+  return num < await comet.numAssets();
+}
+
+// XXX introduce a SupplyCapConstraint to separately test the happy path and revert path instead
+// of testing them conditionally
+async function testSupplyCollateral(context: CometContext, assetNum: number): Promise<null | ContractReceipt> {
+  const comet = await context.getComet();
+  if (assetNum >= (await comet.numAssets())) return;
+  const { albert } = await context.actors;
+  const { asset: assetAddress, scale: scaleBN, supplyCap } = await comet.getAssetInfo(assetNum);
+  const collateralAsset = context.getAssetByAddress(assetAddress);
+  const scale = scaleBN.toBigInt();
+  const toSupply = 100n * scale;
+
+  expect(await collateralAsset.balanceOf(albert.address)).to.be.equal(toSupply);
+
+  await collateralAsset.approve(albert, comet.address);
+
+  const totalCollateralSupply = (await comet.totalsCollateral(collateralAsset.address)).totalSupplyAsset.toBigInt();
+  if (totalCollateralSupply + toSupply > supplyCap.toBigInt()) {
+    await expect(
+      albert.supplyAsset({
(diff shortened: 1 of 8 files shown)


scenario/SupplyScenario.ts
```diff
@@ -1,8 +1,105 @@
...
+      return await testSupplyCollateral(context, i);
+    }
+  );
+}
+
+for (let i = 0; i < NUM_ASSETS; i++) {
+  scenario(
+    `Comet#supplyFrom > collateral asset ${i}`,
+    {
+      filter: async (ctx) => await isValidAssetIndex(ctx, i),
```
jflatow, line 92: Whoa!

jflatow (commented):
Should this be based on `main`? Looks like we have some issues sourcing on testnets...I wonder where we're actually going to logs, most things should be source-able from fauceteer now 🤔

kevincheng96:
> Should this be based on `main`?

I based it off of your branch because I thought it would be cool to test it with the new proposal. I definitely don't plan to merge it into your branch, so I can rebase it on main whenever.

>  Looks like we have some issues sourcing on testnets...I wonder where we're actually going to logs, most things should be source-able from fauceteer now 🤔

Yeah , since we're trying to source a constant amount of tokens in these scenarios (e.g. 100 units of assetX), some tokens with smaller supplies like WBTC are failing to be sourced, even from the fauceteer. The fauceteer also does not hold any `WETH`, so sourcing `WETH` always fails.

I'm adding a new filter that skips any solutions where the tokens are not sourceable.

kevincheng96:
Rebased on top of main

scenario/utils.ts
```diff
@@ -238,4 +240,22 @@ export async function fetchQuery(
...
-}
\ No newline at end of file
+}
+
+export async function isValidAssetIndex(ctx: CometContext, assetNum: number): Promise<boolean> {
+  const comet = await ctx.getComet();
+  return assetNum < await comet.numAssets();
+}
+
+export async function isSourceable(ctx: CometContext, assetNum: number, amount: number): Promise<boolean> {
```
jflatow, line 250: Fancy! Technically like 'isSourceableFromFauceteer' then, but makes sense

kevincheng96: I actually originally named it something like that, but it felt weird because the function returns `true` if `fauceteer` doesn't exist on that chain.

jflatow: Yeah true, 'sourceable from fauceteer or fauceteer doesnt exist'...maybe `isTriviallySourceable`?

kevincheng96: Done

scenario/WithdrawScenario.ts
```diff
@@ -1,8 +1,80 @@
...
+
+  return txn; // return txn to measure gas
+}
+
+for (let i = 0; i < NUM_ASSETS; i++) {
+  const amountToWithdraw = 100; // in units of asset, not wei
+  scenario(
+    `Comet#withdraw > collateral asset ${i}`,
+    {
+      filter: async (ctx) => await isValidAssetIndex(ctx, i) && await isTriviallySourceable(ctx, i, amountToWithdraw),
```
jflatow, line 51: So, why wouldn't we have a valid asset index?

jflatow: Oh, should `NUM_ASSETS` be `MAX_ASSETS`?

kevincheng96: Oh good point, yeah I'll rename it `MAX_ASSETS`.

kevincheng96: Done

jflatow (approved):
This is pretty 🍠 , my only concern is how much slower scens might be now, but looks like its not too bad, potentially only a few extra minutes in CI so far

PR #547, opened 2022-08-19
Verify the comet impl as part of normal cli verification process

jflatow:


scripts/verify-comet.ts
```diff
@@ -1,73 +0,0 @@
...
-}
-
-// We recommend this pattern to be able to use async/await everywhere
-// and properly handle errors.
-main()
-  .then(() => process.exit(0))
-  .catch((error) => {
-    console.error(error);
-    process.exit(1);
-  });
```
kevincheng96, line 73: Unfortunately, we'll probably still need this script for when Comet is deployed not through the initial deploy process (e.g. governance proposals).

kevincheng96 (approved):
🙌

scripts/verify-comet.ts
jflatow: We can still run the deploy cli in those cases and it should work! (just ran it with simulate first to test, then for real, and the only action it does is attempt to verify the impl, which of course is already verified)

kevincheng96: Oh nice! Should we document this in the README? Since it's less obvious how we can manually verify now.

kevincheng96: Also probably good to just mention in general that we need to manually verify after every upgrade to Comet.

README.md
```diff
@@ -218,10 +218,22 @@ These simulations are extremely useful for testing deployments before actually c
 
 0. Create the deployment script and configuration file, and test locally
 1. Open a PR containing the new deployment directory files
-2. Trigger the `run-deploy` workflow action through the GitHub UI
+2. Trigger the `deploy-market` workflow action through the GitHub UI
 3. Inspect the new `roots.json` which the workflow automatically commited to your PR
 4. Start using the new protocol deployment and/or create further migrations to modify it
 
+##### Verifying Deployments
```
kevincheng96, line 225: 👍

kevincheng96 (approved):

PR #549, opened 2022-08-19
[do not merge] deploy Bulker to unblock UI development

scott-silver:
Hacked-together deploy of the Bulker contract to unblock Mykel and David in developing the bulker UI.

We shouldn't merge this change, obviously, since it deletes most of the Kovan deploy script. But the updated address should unblock development for now.

[update: i didn't need to delete all that, did i? our script is now clever enough to be re-run and it will just deploy what is missing]

[update; doing this the correct way now]


kevincheng96:
Closing since Bulker was redeployed on Kovan in #548

PR #550, opened 2022-08-22
Bulker scenarios

kevincheng96:
This PR adds scenarios for Bulker.

### DIFF
--- a/scenario/BulkerScenario.ts
+++ b/scenario/BulkerScenario.ts
@@ -9,7 +9,7 @@ scenario(
   {
     filter: async (ctx) => await isBulkerSupported(ctx),
     tokenBalances: {
-      albert: { $asset0: 100 },
+      albert: { $base: '== 0', $asset0: 100 },
       $comet: { $base: 5000 },
     },
   },
@@ -34,6 +34,7 @@ scenario(
 
     // Initial expectations
     expect(await collateralAsset.balanceOf(albert.address)).to.be.equal(toSupplyCollateral);
+    expect(await baseAsset.balanceOf(albert.address)).to.be.equal(0n);
     expect(await comet.balanceOf(albert.address)).to.be.equal(0n);
 
     // Albert's actions:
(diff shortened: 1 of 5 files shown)


deployments/kovan/usdc/roots.json
```diff
@@ -3,5 +3,5 @@
     "configurator": "0xDb3c6Ae44FE4689f142Ed8dE1a87304249d3d5a6",
     "rewards": "0xC694877D91A8aEfb9D95cf34393cdC0DDdAded18",
     "fauceteer": "0xA42d20100750b4E4d777bD067134423C0775B869",
-    "bulker": "0x258fb42cCe70fC0EbAA00bFe9857eAdD41E56FA9"
+    "bulker": "0x37b29E24de3AE3a71665d3217f8DEc846f82C505"
```
kevincheng96, line 6: This the most up-to-date `Bulker` deployed by 🥈 last week in #549.

scenario/BulkerScenario.ts
```diff
@@ -0,0 +1,165 @@
+import { scenario } from './context/CometContext';
+import { constants, utils } from 'ethers';
+import { expect } from 'chai';
+import { isRewardSupported, isBulkerSupported, getExpectedBaseBalance } from './utils';
+import { exp } from '../test/helpers';
+
+scenario(
+  'Comet#bulker > all non-reward actions in one txn',
```
kevincheng96, line 8: I decided to keep this scenario even though we also have another scenario that tests all actions (including claim reward) in one txn. This is because that scenario filters out chains without an initialized rewards contract, so keeping this scenario should still give us Bulker scenario coverage on those chains.

jflatow (approved):

deployments/kovan/usdc/roots.json
kevincheng96: This is no longer applicable after rebasing on main, which recently deployed a new Kovan Bulker in #548

PR #551, opened 2022-08-23
Intercept sendTransaction in NonceManager to reset pending counts

jflatow:
If there's an error sending a transaction we want to reset the nonce.
This sneakily intercepts errors and resets in our NonceManager subclass.


kevincheng96 (commented):
🧼

kevincheng96 (approved):

PR #552, opened 2022-08-24
Support supply max base in Bulker

kevincheng96:
WIP


jflatow:
Cool WIP 💪 Not sure if/when we'll need it, do we want to keep it open?

kevincheng96:
> Cool WIP 💪 Not sure if/when we'll need it, do we want to keep it open?

I'll keep this open for now. But I think the better approach might be to support it natively in Comet instead.

Hektoreth (commented):
??

PR #553, opened 2022-08-24
Fix proposal constraint and fetchQuery

jflatow:
Scen plugin should not depend on scens.
Errors should be thrown in fetchQuery to avoid misusage.
All open proposals should be executed, not just pending/active.

This increases the need for 'static constraints' as proposal constraint is very expensive.
This also hacks in a constant for timelock buf which should really be computed.

### DIFF
--- a/plugins/scenario/utils/TokenSourcer.ts
+++ b/plugins/scenario/utils/TokenSourcer.ts
@@ -1,8 +1,7 @@
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
-import { BigNumber, Contract, Event } from 'ethers';
+import { BigNumber, Contract, Event, EventFilter } from 'ethers';
 import { erc20 } from './ERC20';
 import { DeploymentManager } from '../../deployment_manager/DeploymentManager';
-import { fetchQuery } from '../../../scenario/utils';
 
 const getMaxEntry = (args: [string, BigNumber][]) =>
   args.reduce(([a1, m], [a2, e]) => (m.gte(e) == true ? [a1, m] : [a2, e]));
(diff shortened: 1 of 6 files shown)


plugins/scenario/utils/TokenSourcer.ts
```diff
@@ -15,6 +14,40 @@ interface SourceTokenParameters {
...
+      const nextFrom = fromBlock - BLOCK_SPAN;
+      if (nextFrom < 0) {
+        throw(new Error('No events found by chain genesis'));
+      }
+      return fetchQuery(contract, filter, nextFrom, nextToBlock, originalBlock);
+    }
+  } catch (err) {
+    if (err.message.includes('query returned more')) {
+      const midBlock = (fromBlock + toBlock) / 2;
+      return fetchQuery(contract, filter, midBlock, toBlock, originalBlock);
```
kevincheng96, line 44: Missing `await`

scenario/utils.ts
```diff
@@ -272,4 +237,20 @@ export async function isRewardSupported(ctx: CometContext): Promise<boolean> {
...
+  fromBlock: number,
+  toBlock: number,
+  BLOCK_SPAN = 2048 // NB: sadly max for fuji
+): Promise<Event[]> {
+  if (toBlock - fromBlock > BLOCK_SPAN) {
+    const midBlock = fromBlock + BLOCK_SPAN;
+    const logs = await contract.queryFilter(filter, fromBlock, midBlock);
+    return logs.concat(await fetchLogs(contract, filter, midBlock + 1, toBlock));
+  } else {
+    return contract.queryFilter(filter, fromBlock, toBlock);
```
kevincheng96, line 254: Missing `await` as well

kevincheng96 (approved):
Great fixes/improvements 👏

plugins/scenario/utils/TokenSourcer.ts
jflatow: Those aren't missing... when you return `await` its the same as just returning the promise

kevincheng96: Oh interesting, TIL. Apparently there's still a slight difference. `return await` is slightly costlier but better for debugging since it keeps the function on the call stack so it shows up in the stack trace.

jflatow: word

PR #554, opened 2022-08-26
Delete executed migration to initialize USDC market on mainnet

jflatow:


kevincheng96 (approved):
Personal opinion, but I still think keeping the migrations in the repo and nullifying them somehow (e.g. adding a check in `enact` like `'if proposal 116 exists, then do nothing`) is more preferable than deleting them. Using commit history to track previous migrations seems a lot more cumbersome and unobvious.

Going to approve anyway since this is the current agreed upon approach.

PR #555, opened 2022-08-26
WIP CToken Migrator

hayesgm:
This patch starts to describe the CToken Migrator machine, which allows users to easily move a position from v2 to v3 in one transaction*. This includes using Uniswap for a flash loan for migration, and exchanging borrow token types (e.g. for DAI).

* Offer not valid in Illinois, Utah and.. err, "one transaction after various required approvals"

### DIFF
--- a/contracts/migrator/README.md
+++ b/contracts/migrator/README.md
@@ -24,6 +24,10 @@ Users can specify the following parameters, generally:
  * `borrowToken: Erc20` **immutable**: The underlying borrow token (e.g. `USDC`).
  * `sweepee: address` **immutable**: Address of an address to send swept tokens to, if for any reason they remain locked in this contract.
 
+## Events
+
+TODO
+
 ## Contract Functions
 
 ### Constructor
@@ -119,9 +123,8 @@ This function may only be called during a migration command. We ensure this by m
 
 #### Inputs
 
- - `sender: address`: The address which called `swap` on Uniswap. Should be this contract's address.
- - `uint amount0`: The amount of token 0 transfered. Ingored.
- - `uint amount1`: The amount of token 1 transfered. Ingored.
+ - `uint fee0`: The fee for borrowing token0 from pool. Ingored.
+ - `uint fee1`: The fee for borrowing token1 from pool. Ingored.
  - `calldata data`: The data encoded above, which is the ABI-encoding of XXX.
 
 #### Bindings
@@ -130,7 +133,7 @@ This function may only be called during a migration command. We ensure this by m
 
 #### Function Spec
 
-`function uniswapV2Call(address sender, uint amount0, uint amount1, bytes calldata data)`
+`function uniswapV3FlashCallback(uint256 fee0, uint256 fee1, bytes calldata data)`
 
   - **REQUIRE** `inMigration == 1`
   - **REQUIRE** `msg.sender == uniswapLiquidityPool`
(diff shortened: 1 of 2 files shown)


contracts/migrator/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+- **REQUIRE** `inMigration == 0`
+- **STORE** `inMigration += 1`
+- **BIND** `user = msg.sender`
+- **WHEN** `repayAmount == type(uint256).max)`:
+  - **BIND READ** `repayAmountActual = cToken.borrowBalanceCurrent(user)`
+- **ELSE**
+  - **BIND** `repayAmountActual = repayAmount`
+- **BIND** `borrowAmountTotal = repayAmountActual + FullMath.mulDivRoundingUp(repayAmountActual, uniswapLiquidityPoolFee, 1e6)` # TODO: FullMath
+- **BIND** `data = abi.encode(MigrationCallbackData{user, repayAmountActual, repayBorrowBehalf, collateral})`
```
kevincheng96, line 137: `repayBorrowBehalf` should be `borrowAmountTotal`?

contracts/migrator/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+`function uniswapV3FlashCallback(uint256 fee0, uint256 fee1, bytes calldata data)`
+
+  - **REQUIRE** `inMigration == 1`
+  - **REQUIRE** `msg.sender == uniswapLiquidityPool`
+  - **REQUIRE** `sender == address(this)`
+  - **BIND** `MigrationCallbackData{user, repayAmountActual, borrowAmountTotal, collateral} = abi.decode(data, (MigrationCallbackData))`
+  - **CALL** `borrowCToken.repayBorrowBehalf(user, repayAmountActual)`
+  - **FOREACH** `(cToken, amount)` in `collateral`:
+    - **CALL** `cToken.transferFrom(user, amount == type(uint256).max ? cToken.balanceOf(user) : amount)`
```
kevincheng96, line 171: Missing the `dst` in the `transferFrom` call.

contracts/migrator/README.md
```diff
@@ -0,0 +1,193 @@
...
+    - **CALL** `cToken.redeem(cToken.balanceOf(address(this)))`
+    - **CALL** `comet.supplyCollateral(address(this), user, cToken.underlying(), cToken.underlying().balanceOf(address(this)))`
```
kevincheng96, line 173: I'm assuming we're calling `balanceOf` here instead of using the `amount` calculated in the `transferFrom` case to handle tokens with transfer fees? I'm not sure if that's totally necessary because `Comet` doesn't support tokens with transfer fees.

contracts/migrator/README.md
```diff
@@ -0,0 +1,193 @@
...
+  - **CALL** `comet.withdrawBase(user, address(this), borrowAmountTotal)`
+  - **CALL** `pay(borrowToken, address(this), msg.sender, borrowAmountTotal)`
```
kevincheng96, line 175: I think this should actually pay down the flash loan from Uniswap first, then pay any remaining `borrowToken` to the user.

contracts/migrator/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Sweep Function
+
+Sends any tokens in this contract to the sweepee address. This contract should never hold tokens, so this is just to fix any anomalistic situations where tokens end up locked in the contract.
```
kevincheng96, line 179: Another way to do this is to have an `admin` that can sweep tokens to any recipient. That way, you wouldn't have to hardcode the `sweepee` and could set the `admin` as the `Timelock`. That's how we set it up in the `Bulker`.

hayesgm:
This has been moved into [https://](https://github.com/compound-finance/comet_v2_migrator)

PR #557, opened 2022-08-30
Potential rewards proposal

jflatow:
~~We would want to adjust the actual COMP speed to equal the sum of the streams we are replacing (TODO).~~

This has been adjusted, now need to finish testing

### DIFF
--- a/deployments/mainnet/usdc/migrations/1661899622_rewards.ts
+++ b/deployments/mainnet/usdc/migrations/1661899622_rewards.ts
@@ -1,13 +1,15 @@
 import { DeploymentManager, migration } from '../../../../plugins/deployment_manager';
-import { calldata, debug, exp, proposal } from '../../../../src/deploy';
+import { calldata, exp, proposal } from '../../../../src/deploy';
 
 import { expect } from 'chai';
 
 const cETHAddress = '0x4Ddc2D193948926D02f9B1fE9e1daa0718270ED5';
+const cBATAddress = '0x6C8c6b02E7b2BE14d4fA6022Dfd6d75921D90E4E';
 const cCOMPAddress = '0x70e36f6bf80a52b3b46b3af8e106cc0ed743e8e4';
 const cLINKAddress = '0xface851a4921ce59e912d19329929ce6da6eb0c7';
 const cUNIAddress = '0x35a18000230da775cac24873d00ff85bccded550';
 const cWBTC2Address = '0xccF4429DB6322D5C611ee964527D42E5d685DD6a';
+const cZRXAddress = '0xB3319f5D18Bc0D84dD1b4825Dcde5d5f7266d407';
 
 export default migration('1661899622_rewards', {
   async prepare(deploymentManager: DeploymentManager) {
@@ -15,6 +17,7 @@ export default migration('1661899622_rewards', {
   },
 
   async enact(deploymentManager: DeploymentManager) {
+    const trace = deploymentManager.tracer();
     const ethers = deploymentManager.hre.ethers;
 
     const {
@@ -33,9 +36,9 @@ export default migration('1661899622_rewards', {
         contract: comptrollerV2,
         signature: '_setCompSpeeds(address[],uint256[],uint256[])',
         args: [
-          [cETHAddress, cCOMPAddress, cLINKAddress, cUNIAddress, cWBTC2Address],
-          [exp(5.375, 15), 0, 0, 0, 0],
-          [0, 0, 0, 0, 0],
+          [cETHAddress, cBATAddress, cCOMPAddress, cLINKAddress, cUNIAddress, cWBTC2Address, cZRXAddress],
+          [exp(5.375, 15), 0, 0, 0, 0, 0, 0],
+          [0, 0, 0, 0, 0, 0, 0],
         ],
       },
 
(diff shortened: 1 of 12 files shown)


deployments/mainnet/usdc/migrations/1661899622_rewards.ts
```diff
@@ -0,0 +1,77 @@
...
+      {
+        contract: cometAdmin,
+        signature: 'deployAndUpgradeTo(address,address)',
+        args: [configurator.address, comet.address],
+      },
+
+      // 4. Transfer COMP
+      {
+        target: comptrollerAddress,
+        signature: '_grantComp(address,uint256)',
```
kevincheng96, line 60: Transfer instead of stream?

scott-silver (approved):

kevincheng96 (approved):
LGTM. Would be great to have an adhoc scenario to test the states (v2 speeds, v3 speed, reward contract balance) after the proposal execution.

jflatow:
> LGTM. Would be great to have an adhoc scenario to test the states (v2 speeds, v3 speed, reward contract balance) after the proposal execution.

Yeah agree and good idea, gonna add one of those next, just trying to make sure all scens pass normally at least locally first

deployments/mainnet/usdc/migrations/1661899622_rewards.ts
jflatow: Yes, mainly to avoid people having to keep submitting txes whenever the stream gets dry

deployments/mainnet/usdc/migrations/1661899622_rewards.ts
```diff
@@ -0,0 +1,103 @@
...
+      async () => (await governor.propose(...await proposal(actions, description))).wait()
+    );
+
+    const event = txn.events.find(event => event.event === 'ProposalCreated');
+    const [proposalId] = event.args;
+
+    debug(`Created proposal ${proposalId}.`);
+  },
+
+  async verify(deploymentManager: DeploymentManager) {
```
jflatow, line 74: Added a `verify` step to check the results (this could/should be done statically with `MigrationConstraint` and `ProposalConstraint`)

scott-silver: nice!

scenario/constraints/MigrationConstraint.ts
```diff
@@ -54,3 +52,23 @@ export class MigrationConstraint<T extends CometContext, R extends Requirements>
...
+          if (migration.actions.verify) {
+            await migration.actions.verify(ctx.deploymentManager);
+            debug(`${label} Verified migration "${migration.name}"`);
+          }
+        }
+        return ctx;
+      }
+    ];
+  }
+
```
kevincheng96, line 72: I think you can add this to the `check` function of `MigrationConstraint` instead of creating a separate constraint.

jflatow: These have to run after the proposal executes not after its created

kevincheng96: Yeah, doesn't check happen after all constraints (e.g. proposal constraint) are run?

deployments/mainnet/usdc/migrations/1661899622_rewards.ts
```diff
@@ -0,0 +1,103 @@
...
+        signature: '_grantComp(address,uint256)',
+        args: [rewards.address, exp(25_000, 18)],
+      },
+    ];
+    const description = "# Initialize Compound III COMP Distribution\nThis proposal includes Compound III users in the COMP Distribution, which is accomplished by re-allocating 161.42 COMP per day (14.17% of the total) from v2 markets. The total COMP Distribution is unchanged by this proposal.\n\nETH borrowers: -35.31\nWBTC suppliers and borrowers: -70.63\nLINK suppliers and borrowers: -9.60\nUNI suppliers and borrowers: -9.60\nCOMP suppliers: -16.42\nBAT suppliers and borrowers: -9.60\nZRX suppliers and borrowers: -9.60\nCompound III USDC: +161.42\n\nLastly, 25,000 COMP are transferred from the Comptroller to the Compound III Rewards contract. Based on the parameters set earlier, this equals 154 days of distribution before governance would need to replenish or modify the distribution to Compound III users.\n\nIf successful, this proposal will take effect after Proposal 119 cures the Compound v2 price feed, and prior to the Merge.\n\n[Full proposal and forum discussion](https://www.comp.xyz/t/initialize-compound-iii-usdc-on-ethereum/3499/5)\n";
```
torreyatcitty, line 63: I was able to run the method on the old backend `Shared.GovHelper.title_and_description` which parses the title out of the description and got the following:
```
{:ok,                                                                                                {"Initialize Compound III COMP Distribution",                                                        "This proposal includes Compound III users in the COMP Distribution, which is accomplished by re-allocating 161.42 COMP per day (14.17% of the total) from v2 markets. The total COMP Distribution is unchanged by this proposal.\n\nETH borrowers: -35.31\nWBTC suppliers and borrowers: -70.63\nLINK suppliers and borrowers: -9.60\nUNI suppliers and borrowers: -9.60\nCOMP suppliers: -16.42\nBAT suppliers and borrowers: -9.60\nZRX suppliers and borrowers: -9.60\nCompound III USDC: +161.42\n\nLastly, 25,000 COMP are transferred from the Comptroller to the Compound III Rewards contract. Based on the parameters set earlier, this equals 154 days of distribution before governance would need to replenish or modify the distribution to Compound III users.\n\nIf successful, this proposal will take effect after Proposal 119 cures the Compound v2 price feed, and prior to the Merge.\n\n[Full proposal and forum discussion](https://www.comp.xyz/t/initialize-compound-iii-usdc-on-ethereum/3499/5)"}}
```

coburncoburn (commented):
this gave me a random idea for something to build, :
Fork at 1000 blocks ago, propose and execute this proposal, and then attempt to replay all the transactions from the intermittent 1000 blocks, expecting them to emit identical events but for whatever expected changes the proposal makes. Just dropping it in so as not to forget the idea completely

jflatow:
> this gave me a random idea for something to build, : Fork at 1000 blocks ago, propose and execute this proposal, and then attempt to replay all the transactions from the intermittent 1000 blocks, expecting them to emit identical events but for whatever expected changes the proposal makes. Just dropping it in so as not to forget the idea completely

cool idea, you could also fork now and run historical txes, but I guess they are more likely to succeed the first time. pretty complex though, good for back burner

scenario/constraints/MigrationConstraint.ts
jflatow: Ah hm good point! Although this will all be affected as I split out static constraints, so I don't feel too bad having added the extra constraint, maybe

kevincheng96: Ok sounds good, let's keep it as a separate constraint then

kevincheng96:
Pushed a commit that adds a scenario for testing borrow-side rewards, fixes the `isRewardSupported` filter, and makes some other improvements to the reward-related scenarios.

scenario/constraints/MigrationConstraint.ts
```diff
@@ -30,7 +25,12 @@ export class MigrationConstraint<T extends CometContext, R extends Requirements>
         // Make proposer the default signer
         ctx.deploymentManager._signers.unshift(proposer);
 
+        // Order migrations deterministically and store in the context (i.e. for verification)
         migrationList.sort((a, b) => a.name.localeCompare(b.name))
+        ctx.migrations = migrationList;
+
+        // XXX This should check that a migration has not already been run/proposed on-chain.
+        // Otherwise the scenario could be running the same proposal twice.
```
kevincheng96, line 33: Added this XXX because the rewards migration is still being run in the scenario even though it's already been created as a proposal on-chain. I have to comment out the `MigrationConstraint` and `VerifyMigrationConstraint` locally when running scenarios because the rewards migration is not idempotent.

jflatow: Do we want to add back the enacted flag? Maybe we write it out in enact automatically

kevincheng96: We will do this in a separate PR

PR #558, opened 2022-09-03
Deploy to Goerli

scott-silver:
attempting to create a Goerli/USDC deployment, in part because we'll need it to test a Mumbai (Avalanche testnet) l2 deployment and in part because there's a chance that Etherscan may become unusable for Kovan soon.

The set of [ChainLink feeds for Goerli](https://docs.chain.link/docs/ethereum-addresses/#Goerli%20Testnet) is currently very limited, and doesn't include COMP 😞:

image

~~So this deployment does not include COMP (or any other reward token).~~ [now with COMP!] It uses LINK as the first collateral asset (the first collateral asset is weirdly important for scenarios, since we reference it via `asset0` frequently).


scenario/BulkerScenario.ts
```diff
@@ -38,7 +38,7 @@ scenario(
     expect(await comet.balanceOf(albert.address)).to.be.equal(0n);
 
     // Albert's actions:
-    // 1. Supplies 100 units of collateral
+    // 1. Supplies 1000 units of collateral
```
scott-silver, line 41: This test currently assumes that 100 units of `asset0` are worth 1500 of the base unit.

The price of LINK on Goerli is currently around $7.

Bumping up the supplied collateral to cover the borrow,

scenario/SupplyScenario.ts
```diff
@@ -295,15 +295,14 @@ scenario(
...
       betty.supplyAssetFrom({
         src: albert.address,
         dst: betty.address,
         asset: collateralAsset.address,
         amount: 100n * scale,
-      }),
-      [/ERC20: transfer amount exceeds allowance/, /transfer amount exceeds spender allowance/]
-    );
+      })
+    ).to.be.reverted;
```
scott-silver, line 305: These scenarios are great, but they assume that the token returns an error message of some sort. LINK reverts without a reason string for these two instances.

jflatow: Oy

kevincheng96:
> * do we want to have COMP as an asset, even if there is currently no ChainLink feed for COMP? (if so, should we use the ChainLink feed of a different asset?)
>   
>   * other option, from Adam: use SimplePriceFeed until ChainLink publishes a COMP feed

I think using a `SimplePriceFeed` until we have Chainlink support is best. It'll be easy to switch the price feed once Chainlink supports COMP.

> * related: do we want to have a reward token on this deployment? Is that reward asset COMP, or something with a ChainLink price feed?

Yeah, it'll be nice to be able to test rewards on testnet. If we set up a `SimplePriceFeed` for COMP, we should be able to use COMP as the reward asset without any issues. Might be interesting/cool to test a different reward asset on testnet, but it probably doesn't make sense on Goerli since it should mirror mainnet.

.github/workflows/deploy-market.yaml
```diff
@@ -9,6 +9,7 @@ on:
           - fuji
           - kovan
           - mainnet
+          - goerli
```
kevincheng96, line 12: You might to also add an entry for `goerli` in line 33 (`ethereum_url` field). Same for the other workflows.

README.md
```diff
@@ -50,10 +50,6 @@ Currently, Avalanche mainnet and testnet (fuji) are supported. This means that d
 
 To use this project with other chains, the block explorer API key for your target chain must be set in .env (e.g. `SNOWTRACE_KEY` for Avalanche).
 
-An example deployment command looks like:
-
-`yarn deploy --network fuji`
```
kevincheng96, line 55: Is this no longer true?

deployments/goerli/usdc/configuration.json
```diff
@@ -0,0 +1,51 @@
...
+    "borrowKink": 0.8,
+    "borrowSlopeLow": 0.035,
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000011574074074074073e15",
+    "baseBorrowSpeed": "0.0011458333333333333e15",
+    "baseMinForRewards": "1000000e6"
```
kevincheng96, line 23: Should we lower this on testnets to $10k or $100k instead?

deployments/goerli/usdc/deploy.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+  );
+
+  const WBTC = await deploymentManager.clone('WBTC', clone.wbtc, []);
+  const WETH = await deploymentManager.clone('WETH', clone.weth, []);
+  const LINK = await deploymentManager.clone('LINK', clone.link, []);
+
+  // Deploy all Comet-related contracts
+  const deployed = await deployComet(deploymentManager, deploySpec);
+  // const { comet, rewards } = deployed;
```
kevincheng96, line 58: Can delete?

kevincheng96 (commented):
🎉

README.md
scott-silver: yeah, `yarn deploy` isn't a command anymore

jflatow: `yarn hardhat deploy`

scott-silver: so really this line should read:

```
yarn hardhat deploy --network fuji --deployment usdc
```

.github/workflows/deploy-market.yaml
scott-silver: good catch.

that line is going to be awful when we've got 5-6 deployments (with their corresponding testnets)

scott-silver: updated

README.md
scott-silver: updated

deployments/goerli/usdc/configuration.json
scott-silver: yep, that works. updated to $10K

deployments/goerli/usdc/deploy.ts
scott-silver: deleted

scott-silver:
Added COMP as a collateral asset; updated to use SimplePriceFeed as its feed

jflatow:
Superseded by #565

PR #560, opened 2022-09-06
L2 Configurator Diagram

scott-silver:
image


jflatow (approved):
Good diagram which I think has held up relatively well, shall we merge?

se6799 (commented):
Need. Money

PR #561, opened 2022-09-09
Mumbai deployment

scott-silver:
Mumbai deployment, to go along with the Goerli deployment.

Includes:
-  an updated version of the Timelock (called `BridgeTimelock`) that includes immutable values and custom errors
- an `ITimelock` interface
- a `PolygonBridgeReceiver`
- Mumbai configuration
- a Mumbai deploy script

Does not yet include:
- [x] functioning ProposalConstraint (#568)
- [x] functioning GovernanceScenario (#568)
- [x] changes to migration workflow (#575)
- [x] a base `BridgeReceiver` contract that our various bridge contracts can inherit from (#568)
- [x] Vendoza config for the imported Polygon contract

Opening this PR any way so we can start answering some of the remaining open questions

Thanks for the help ⛑️ ! And a bunch of this work is ported from Mykel's original PR: #468.

Open questions:

- ~~**should we be able to cancel transactions that are queued in the BridgeTimelock?**~~
  - ~~if so, should this be done by a Polygon/Mumbai multi-sig?~~
  - conclusion: no, we will not include functionality that allows for the L2 Timelock transactions to be canceled
- ~~should the BridgeReceiver contract be a proxy?~~
  - conclusion: no, it will be a non-upgradeable contract


contracts/bridges/BridgeTimelock.sol
```diff
@@ -0,0 +1,113 @@
...
+    uint public immutable maximumDelay;
+
+    address public admin;
+    address public pendingAdmin;
+    uint public delay;
+
+    mapping (bytes32 => bool) public queuedTransactions;
+
+    constructor(address admin_, uint delay_, uint gracePeriod_, uint minimumDelay_, uint maxiumumDelay_) public {
+        if (delay_ < minimumDelay_) revert BadDelay();
```
scott-silver, line 27: I converted all of the `require`s from the original Timelock into `revert`s with Custom Errors.

probably worth some extra scrutiny to make sure I got the logic inversions correct

contracts/bridges/BridgeTimelock.sol
```diff
@@ -0,0 +1,113 @@
...
+
+import "../ITimelock.sol";
+
+contract BridgeTimelock is ITimelock {
+    /** Custom errors **/
+    error BadDelay();
+    error BadETA();
+    error TransactionExpired();
+    error TransactionNotQueued();
+    error TransactionNotReady();
```
scott-silver, line 12: better name than `TransactionNotReady`? `TransactionNotExecutable`?

.github/workflows/deploy-market.yaml
```diff
@@ -30,7 +31,7 @@ jobs:
       - name: Seacrest
         uses: hayesgm/seacrest@v1
         with:
-          ethereum_url: "${{ fromJSON('{\"fuji\":\"https://api.avax-test.network/ext/bc/C/rpc\",\"kovan\":\"https://kovan.infura.io/v3/$INFURA_KEY\",\"mainnet\":\"https://mainnet.infura.io/v3/$INFURA_KEY\",\"goerli\":\"https://goerli.infura.io/v3/$INFURA_KEY\"}')[inputs.network] }}"
+          ethereum_url: "${{ fromJSON('{\"fuji\":\"https://api.avax-test.network/ext/bc/C/rpc\",\"kovan\":\"https://kovan.infura.io/v3/$INFURA_KEY\",\"mainnet\":\"https://mainnet.infura.io/v3/$INFURA_KEY\",\"goerli\":\"https://goerli.infura.io/v3/$INFURA_KEY\",\"mumbai\":\"https://polygon-mumbai.infura.io/v3/$INFURA_KEY\"}')[inputs.network] }}"
```
scott-silver, line 34:
```
"https://polygon-mumbai.infura.io/v3/$INFURA_KEY\"
```

🤔 does this actually work?

contracts/bridges/BridgeTimelock.sol
```diff
@@ -0,0 +1,113 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+import "../ITimelock.sol";
+
+contract BridgeTimelock is ITimelock {
```
scott-silver, line 6: this Timelock no longer has any bridge-specific logic in it; we can probably rename it to just `Timelock` (or `TimelockBravo` to distinguish from the original?) and move it down to `contracts/`

scenario/context/CometContext.ts
```diff
@@ -396,7 +396,7 @@ async function forkContext(c: CometContext, w: World): Promise<CometContext> {
 export const constraints = [
   new FilterConstraint(),
   new MigrationConstraint(),
-  new ProposalConstraint(),
+  // new ProposalConstraint(),
```
scott-silver, line 399: - [ ] Re-enable ProposalConstraint

.github/workflows/deploy-market.yaml
jflatow: You mean the escaping?

scott-silver: yeah

deployments/mumbai/usdc/configuration.json
```diff
@@ -0,0 +1,59 @@
...
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000011574074074074073e15",
+    "baseBorrowSpeed": "0.0011458333333333333e15",
+    "baseMinForRewards": "1000000e6"
+  },
+  "assets": {
+    "SAND": {
```
scott-silver, line 26: we use `asset0` in a number of scenarios, so the first listed asset needs to be something that we can easily mint a large number of. i've never heard of SAND before, but it's very easy to mint

contracts/bridges/BridgeTimelock.sol
kevincheng96: `TransactionEtaNotReached`?

contracts/bridges/BridgeTimelock.sol
kevincheng96: True. I don't have a strong opinion and am fine with either name.

contracts/bridges/polygon/FxBaseChildTunnel.sol
```diff
@@ -0,0 +1,78 @@
...
+        uint256 stateId,
+        address rootMessageSender,
+        bytes calldata data
+    ) external;
+}
+
+/**
+ * @notice Mock child tunnel contract to receive and send message from L2
+ */
+abstract contract FxBaseChildTunnel is IFxMessageProcessor {
```
kevincheng96, line 16: We don't need this. It's more of an example from Polygon.

contracts/bridges/polygon/FxBaseChildTunnel.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+// IFxMessageProcessor represents interface to process message
+interface IFxMessageProcessor {
```
kevincheng96, line 5: If we remove `FxBaseChildTunnel` from this file, then we can consider moving this interface definition to the top of  `PolygonBridgeReceiver.sol`.

deployments/mumbai/usdc/configuration.json
kevincheng96: lol interesting that they support SAND but not LINK on mumbai

scott-silver: weirdly, they have a `LINK / MATIC` feed, but not `LINK / USD` feed

contracts/bridges/polygon/FxBaseChildTunnel.sol
scott-silver: alternatively, we could track this file with Vendoza and then delete the unnecessary parts (but that deletion would at least be tracked/reproducible/updatable)

.github/workflows/deploy-market.yaml
jflatow: yea it has been, right?

contracts/bridges/BridgeTimelock.sol
jflatow: We should be able to get rid of `SimpleTimelock` at the same time (and I think we should do this to check assumptions and make sure things work as expected)

deployments/mumbai/usdc/configuration.json
jflatow: 🤔

contracts/bridges/polygon/FxBaseChildTunnel.sol
jflatow: I still don't understand why there can't be a generic child that everyone shares to get the same interface as Optimism

contracts/bridges/BridgeTimelock.sol
scott-silver: deleted this change to keep Timelock as similar as possible to its original version

contracts/bridges/BridgeTimelock.sol
scott-silver: deleted this change to keep Timelock as similar as possible to its original version

contracts/bridges/polygon/FxBaseChildTunnel.sol
scott-silver: replace `FxBaseChildTunnel` with `FxChild` contract; moved to a `bridges/vendor` directory and started tracking with Vendoza

scott-silver: and deleted all the unnecessary bits

contracts/bridges/BridgeTimelock.sol
kevincheng96: Fair enough, though I think personally think that, just like converting constants to immutables, custom errors are a safe improvement to make.

contracts/vendor/Timelock.sol
```diff
@@ -3,33 +3,25 @@ pragma solidity 0.8.15;
...
-    error TransactionExpired();
-    error TransactionNotQueued();
-    error TransactionNotReady();
-    error TransactionReverted();
-    error Unauthorized();
-
-    uint public immutable gracePeriod;
-    uint public immutable minimumDelay;
-    uint public immutable maximumDelay;
+contract Timelock is ITimelock {
```
jflatow, line 6: Hm, why is this vendor actually?

contracts/vendor/Timelock.sol
```diff
@@ -99,15 +91,17 @@ contract BridgeTimelock is ITimelock {
             callData = abi.encodePacked(bytes4(keccak256(bytes(signature))), data);
         }
 
+        // solium-disable-next-line security/no-call-value
```
jflatow, line 94: I assume we're just trying to minimize the diff? I think we could get rid of these solium hints...

contracts/vendor/Timelock.sol
jflatow: Oh pulling in with vendoza...

scott-silver: yeah, my goals were to reduce the number of changes to Timelock and to track all the things that I changed

contracts/vendor/Timelock.sol
scott-silver: sure, will do

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    address public initializer;
+    address public mainnetTimelock;
+    address public l2Timelock;
+
+    constructor(address _initializer) {
+        initializer = _initializer;
+    }
+
+    function initialize(address _mainnetTimelock, address _l2Timelock) external {
```
scott-silver, line 24: The deploy flow for the BridgeReceiver/L2 Timelock goes like this:

- deploy the BridgeReceiver with an EOA as initializer
- deploy the Timelock with BridgeReceiver address as admin
- initializer EOA calls BridgerReceiver.initialize with the mainnetTimelock and l2Timelock addresses

at that point, BridgeReceiver.initializer is set to the zero address, so there is no action that the initializer EOA can take on the BridgeReceiver contract

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,81 @@
...
+        if (initializer == address(0)) revert AlreadyInitialized();
+        mainnetTimelock = _mainnetTimelock;
+        l2Timelock = _l2Timelock;
+        initializer = address(0);
+        emit Initialized(_mainnetTimelock, _l2Timelock);
+    }
+
+    function acceptL2TimelockAdmin() external {
```
scott-silver, line 32: this function is not actually required in the current flow, but it seems possible that you might in a situation where the BridgeReceiver would need to call `l2Timelock.acceptAdmin()`

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,81 @@
...
+            data,
+            (address[], uint256[], string[], bytes[])
+        );
+
+        if (values.length != targets.length) revert BadData();
+        if (signatures.length != targets.length) revert BadData();
+        if (calldatas.length != targets.length) revert BadData();
+
+        uint delay = ITimelock(l2Timelock).delay();
+        uint eta = block.timestamp + delay;
```
scott-silver, line 72: eta is set to the minimum delay value; we could encode a relative time if we wanted to into the calldata, but it might be unnecessary

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: The other options we considered were:

1. making `initialize` a function that could only be called by the `l2Timelock`; the major downside is that this would require initialization to be queued as a transaction, so getting the contracts initialized would take 2 days (or whatever the minimum delay is)

2. making `initialize` a public function that could be called by anyone. it would be difficult to do anything malicious with this function (calling `l2timelock.acceptAdmin` would simply revert if the timelock hadn't set the BridgeReceiver as the pendingAdmin), but it felt weird to leave that function callable by anyone

contracts/bridges/BaseBridgeReceiver.sol
jflatow: yea, doesn't seem necessary, I think we just queue ASAP (param could always be added)

scenario/context/CometContext.ts
scott-silver: handling in #568

contracts/bridges/BaseBridgeReceiver.sol
jflatow: We shouldn't necessarily be afraid of public `initialize`

plugins/scenario/worker/Worker.ts
```diff
@@ -60,10 +66,14 @@ export async function run<T, U, R>({ bases, config, worker }: WorkerData) {
...
-    const delta = await dm.runDeployScript({ allMissing: true });
-    console.log(`[${base.name}] Deployed ${dm.counter} contracts, spent ${dm.spent} to initialize world 🗺`);
-    console.log(`[${base.name}]\n${dm.diffDelta(delta)}`);
+    const world = new World(base);
+    const dm = world.deploymentManager;
+    await runDeployScript(dm, base.name);
+
+    if (world.auxiliaryDeploymentManager) {
+      await runDeployScript(world.auxiliaryDeploymentManager, base.auxiliaryBase, true);
+    }
```
jflatow, line 75: I wonder if we really need this one

scott-silver: yep, you're right. it's unnecessary. removed

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: I've come to like the initializer more, the more I think about it.

it's a role with a very limited lifespan (as soon as you initialize, you are locked out of the role forever). and it prevents someone from griefing a newly-deployed BridgeReceiver contract (an unlikely event, but one that would take a governance proposal to undo)

contracts/bridges/BridgeTimelock.sol
scott-silver: going to remove `SimpleTimelock` as a follow-up; not going to add it to this PR (which has gotten rather large)

.github/workflows/deploy-market.yaml
kevincheng96, line 34: Missing `POLYGONSCAN_KEY: ${{ secrets.POLYGONSCAN_KEY }}` above

.github/workflows/prepare-migration.yaml
```diff
@@ -33,7 +34,7 @@ jobs:
       - name: Seacrest
         uses: hayesgm/seacrest@v1
         with:
-          ethereum_url: "${{ fromJSON('{\"fuji\":\"https://api.avax-test.network/ext/bc/C/rpc\",\"kovan\":\"https://kovan-eth.compound.finance\",\"mainnet\":\"https://mainnet-eth.compound.finance\",\"goerli\":\"https://goerli.infura.io/v3/$INFURA_KEY\"}')[inputs.network] }}"
+          ethereum_url: "${{ fromJSON('{\"fuji\":\"https://api.avax-test.network/ext/bc/C/rpc\",\"kovan\":\"https://kovan-eth.compound.finance\",\"mainnet\":\"https://mainnet-eth.compound.finance\",\"goerli\":\"https://goerli.infura.io/v3/$INFURA_KEY\",\"mumbai\":\"https://polygon-mumbai.infura.io/v3/$INFURA_KEY\"}')[inputs.network] }}"
```
kevincheng96, line 37: Also missing `POLYSCAN_KEY` above

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,92 @@
...
+    error Unauthorized();
+
+    event Initialized(address indexed l2Timelock, address indexed mainnetTimelock);
+    event NewL2Timelock(address indexed oldL2Timelock, address indexed newL2Timelock);
+    event NewMainnetTimelock(address indexed oldMainnetTimelock, address indexed newMainnetTimelock);
+    event ProcessMessage(address indexed messageSender, address[] targets, uint256[] values, string[] signatures, bytes[] calldatas, uint eta);
+
+    address public initializer;
+    address public mainnetTimelock;
+    address public l2Timelock;
```
kevincheng96, line 18: nit: just wanted to leave a note here that we should probably think of another name, since the child chain might not necessarily be an L2.

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,92 @@
...
+
+        if (values.length != targets.length) revert BadData();
+        if (signatures.length != targets.length) revert BadData();
+        if (calldatas.length != targets.length) revert BadData();
+
+        uint delay = ITimelock(l2Timelock).delay();
+        uint eta = block.timestamp + delay;
+
+        for (uint8 i = 0; i < targets.length; ) {
+            ITimelock(l2Timelock).queueTransaction(targets[i], values[i], signatures[i], calldatas[i], eta);
```
kevincheng96, line 76: Should this first check that an action is not already queued, like in [GovernorBravo](https://github.com/compound-finance/compound-protocol/blob/b9b14038612d846b83f8a009a82c38974ff2dcfe/contracts/Governance/GovernorBravoDelegate.sol#L133)?

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,92 @@
...
+
+        emit ProcessMessage(messageSender, targets, values, signatures, calldatas, eta);
+    }
+
+    function executeTransaction(
+        address target,
+        uint value,
+        string memory signature,
+        bytes memory data,
+        uint eta
```
kevincheng96, line 88: I still think that assigning each action an id and using that to execute is way easier, especially since each action needs to be executed separately and there could be 5+ actions per proposal just for updating supply caps.

kevincheng96, line 88: I still think that assigning each action an id and using that to execute is way easier, especially since each action needs to be executed separately and there could be 5+ actions per proposal just for updating supply caps.

contracts/bridges/polygon/PolygonBridgeReceiver.sol
```diff
@@ -0,0 +1,33 @@
...
+import "../BaseBridgeReceiver.sol";
+
+contract PolygonBridgeReceiver is IFxMessageProcessor, BaseBridgeReceiver {
+    error InvalidChild();
+
+    event NewFxChild(address indexed oldFxChild, address indexed newFxChild);
+
+    address public fxChild;
+
+    constructor(address _fxChild, address _initializer) BaseBridgeReceiver(_initializer) {
```
kevincheng96, line 14: I guess the benefit of making `initialize` publicly callable is we can get rid of this `_initializer` constructor arg. Something to consider 🤔

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,92 @@
...
+    ) external payable returns (bytes memory) {
+        return ITimelock(l2Timelock).executeTransaction(target, value, signature, data, eta);
```
kevincheng96, line 90: Does this need to be `payable`? The call to `l2Timelock` doesn't actually attach any `value`.

contracts/vendor/manifest.json
```diff
@@ -117,6 +207,142 @@
...
+      "patches": [
+        {
+          "oldStart": 1,
+          "oldLines": 29,
+          "newStart": 1,
+          "newLines": 34,
+          "lines": [
+            "-// SPDX-License-Identifier: BUSL-1.1",
+            "-pragma solidity 0.8.15;",
+            "+// SPDX-License-Identifier: BSD-3-Clause",
```
kevincheng96, line 228: Do we actually want to change the license for the Timelock?

contracts/bridges/BaseBridgeReceiver.sol
kevincheng96: Assigning action ids will also allow us to more easily support action states (akin to proposal states in GovernorBravo) such as queued, executed, expired. Not sure if we really want to have action states here, but it would better align with the interfaces we already use in `ProposalConstraint`.

deployments/mumbai/usdc/roots.json
```diff
@@ -0,0 +1,3 @@
+{
+  "fxChild": "0xCf73231F28B7331BBe3124B907840A94851f9f11"
```
kevincheng96, line 2: This is still incomplete because we haven't deployed right?

deployments/goerli/usdc/roots.json
```diff
@@ -8,5 +8,6 @@
     "comet": "0x44Ac0af6F39c98d59Ed6D16404210656dB912Ffd",
     "configurator": "0xdE0e9466F406Eb0922746861Bc09785bED912C80",
     "rewards": "0xef9e070044d62C38D2e316146dDe92AD02CF2c2c",
-    "bulker": "0x2677D9ca7477A81A394C324B97F3D7907C6727Fb"
+    "bulker": "0x2677D9ca7477A81A394C324B97F3D7907C6727Fb",
+    "fxRoot": "0x3d1d3e34f7fb6d26245e6640e1c50710efff15ba"
```
kevincheng96, line 12: Sorry to go back and forth on this, but I guess it could be a bit weird to put `fxRoot` in roots, since currently roots are specific to a certain deployment. What we really want is for `fxRoot` to be in roots for all of goerli, not just the usdc deployment of goerli. So maybe it's better to just hardcode this address in the migration script until we have the concept of a network-level roots?

deployments/mumbai/usdc/roots.json
scott-silver: yep, that's correct

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: I've been using `governanceX`/`bridgedX` in other parts of the code

maybe these are `governanceTimelock`/`bridgedTimelock`?

deployments/goerli/usdc/roots.json
scott-silver: As a counter-argument: we currently specify a specific parent _deployment_ (not just a parent network).

So when we're attaching an `auxiliaryBase` to a scenario base, we're saying that governance of this bridged deployment (e.g. Mumbai/USDC) is the governor of a specific base (e.g. Goerli/USDC).

This makes sense, since it would be possible (albeit unlikely) for there to be a second deployment on the parent network (e.g. Goerli/DAI) that was governed by an entirely different GovernorBravo/Timelock/etc.

And while we don't have shared roots for a network (so `fxRoot` would be duplicated in `deployments/goerli/dai/roots.json`), we do have the ability to share relations (the way we have Goerli's relations inherit from base relations), so there would a limit to how much code would be duplicated across deployments.

scenario/AllowBySigScenario.ts
```diff
@@ -10,7 +10,7 @@ scenario(
     expect(await comet.isAllowed(albert.address, betty.address)).to.be.false;
 
     const nonce = await comet.userNonce(albert.address);
-    const expiry = (await world.timestamp()) + 1_000;
+    const expiry = (await world.timestamp()) + 10_000;
```
scott-silver, line 13: This scenario (`Comet#allowBySig > allows a user to authorize a manager by signature`) fails inconsistently when run in CI.

This it the first scenario in the first scenario file (alphabetically).

Is it possible that the value of `world.timestamp` is returning the last mined block (created when World was first created by the Worker) and that by the time this scenario is executed, the next mined block is more than 1000 blocks past that initial block?

I wonder if it would solve this problem to mine a block before getting the world.timestamp value, instead of bumping the expiry to a very large number.

.github/workflows/deploy-market.yaml
scott-silver: good catch. must have gotten lost in rebase ☹️ 

updated

.github/workflows/prepare-migration.yaml
scott-silver: updated

contracts/bridges/polygon/FxBaseChildTunnel.sol
scott-silver: A couple arguments against it:

1. We need the bridge receiver contract to calculate `eta` using the L2 chain's block timestamp (instead of having it be an estimate that the 

2. there is no guarantee that relayed messages will be sent in a particular order (or that they will be received at all). with the bridge receiver, we're able to send a bundle of transactions to be queued that will all arrive at the same time. if we were relying on a generic receiver contract, we would need to send X transactions in a row and hope that they were all received and all received in the correct order, which won't always be the case

additionally, if we update the changes Kevin is suggesting below, the bridge receiver would also be responsible for ensuring that duplicate transactions were not queued and that the API for executeTransaction was simpler (taking a single transaction id)

deployments/goerli/usdc/roots.json
kevincheng96: Fair enough, that makes a lot of sense!

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: The hashed value includes the `eta`, which in this case we're calculating immediately before queuing

```
require(!timelock.queuedTransactions(keccak256(abi.encode(target, value, signature, data, eta))), "GovernorBravo::queueOrRevertInternal: identical proposal action already queued at eta");
```

So this would only prevent you from queuing the same action twice in the same block?

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: yep, good point. deleted

contracts/vendor/manifest.json
scott-silver: good question. i updated it to match our other Comet contracts, but I don't have a strong opinion on this

deployments/mumbai/usdc/configuration.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "Compound USDC",
+  "symbol": "cUSDCv3",
+  "baseToken": "USDC",
+  "baseTokenPriceFeed": "0x572dDec9087154dC5dfBB1546Bb62713147e0Ab0",
+  "borrowMin": "1000e6",
```
kevincheng96, line 6: I think this should be 100 like mainnet unless we're setting it higher on purpose

deployments/mumbai/usdc/configuration.json
```diff
@@ -0,0 +1,59 @@
...
+    "borrowKink": 0.8,
+    "borrowSlopeLow": 0.035,
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000011574074074074073e15",
+    "baseBorrowSpeed": "0.0011458333333333333e15",
+    "baseMinForRewards": "1000000e6"
```
kevincheng96, line 23: I think we also dropped this to something more achievable for testnets

deployments/mumbai/usdc/deploy.ts
```diff
@@ -0,0 +1,192 @@
...
+  await mintTokens(deploymentManager);
+  return deployed;
+}
+
+async function deployContracts(deploymentManager: DeploymentManager, deploySpec: DeploySpec): Promise<Deployed> {
+  const trace = deploymentManager.tracer()
+  const ethers = deploymentManager.hre.ethers;
+  const signer = await deploymentManager.getSigner();
+
+  const fxChild = await deploymentManager.contract('fxChild');
```
kevincheng96, line 26: Sorry about this one too (I think I suggested making this change 🤦 ), but it seems circular to require `fxChild` to exist in roots before we call `deploy`. I think we're betting off hardcoding `fxChild` in the migration script and just setting it in roots at the end of `deploy`. WDYT?

deployments/mumbai/usdc/deploy.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+  );
+
+  const WBTC = await deploymentManager.clone('WBTC', clone.wbtc, []);
+  const WETH = await deploymentManager.clone('WETH', clone.weth, []);
+  const WMATIC = await deploymentManager.clone(
+    'WMATIC',
+    clone.wmatic,
+    [],
+    'polygon' // NOTE: cloned from Polygon, not mainnet
```
kevincheng96, line 94: 🆒

plugins/scenario/utils/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  await dm.hre.network.provider.send('hardhat_setNextBlockBaseFeePerGas', ['0x0']);
+}
+
+export async function mineBlocks(dm: DeploymentManager, blocks: number) {
+  await dm.hre.network.provider.send('hardhat_mine', [`0x${blocks.toString(16)}`]);
+}
+
+export async function setNextBlockTimestamp(dm: DeploymentManager, timestamp: number) {
+  await dm.hre.ethers.provider.send('evm_setNextBlockTimestamp', [timestamp]);
+}
```
kevincheng96, line 22: nit: Comment probably got lost in the other PR, but I was suggesting these 3 functions could be moved to `scenario/utils.ts` since these aren't used anywhere in the plugin itself.

plugins/scenario/worker/Worker.ts
```diff
@@ -60,10 +66,14 @@ export async function run<T, U, R>({ bases, config, worker }: WorkerData) {
...
-    const world = new World(hreForBase(base), base), dm = world.deploymentManager;
-    const delta = await dm.runDeployScript({ allMissing: true });
-    console.log(`[${base.name}] Deployed ${dm.counter} contracts, spent ${dm.spent} to initialize world 🗺`);
-    console.log(`[${base.name}]\n${dm.diffDelta(delta)}`);
+    const world = new World(base);
+    const dm = world.deploymentManager;
+    await runDeployScript(dm, base.name);
+
+    if (world.auxiliaryDeploymentManager) {
+      await world.auxiliaryDeploymentManager.spider();
```
kevincheng96, line 74: I think we might need to call `runDeployScript` on the auxiliary chain as well. Most of the time, it'll probably just be a no-op.

scenario/AllowBySigScenario.ts
kevincheng96: Good call out. I think you might be right about mining a block being a potential fix for this.

scenario/utils.ts
```diff
@@ -254,3 +266,178 @@ export async function fetchLogs(
...
+  const governor = await dm.getContractOrThrow('governor');
+  const blockNow = await dm.hre.ethers.provider.getBlockNumber();
+  const blocksUntilStart = startBlock - blockNow;
+  const blocksUntilEnd = endBlock - Math.max(startBlock, blockNow);
+
+  if (blocksUntilStart > 0) {
+    await mineBlocks(dm, blocksUntilStart);
+  }
+
+  const compWhales = dm.network === "mainnet" ? COMP_WHALES.mainnet : COMP_WHALES.testnet;
```
kevincheng96, line 283: nit: single quotes

contracts/bridges/polygon/PolygonBridgeReceiver.sol
scott-silver: I think the publicly-callable `initialize` function was for when the `initialize` function was basically just calling `timelock.acceptAdmin()`

If we made this version publicly-callable, then I think we could get in a bad state:

- PolygonBridgeReceiver is deployed (with no `mainnetTimelock` or `l2Timelock` set)
- Timelock is deployed, with PolygonBridgeReceiver as admin

at this point, if `initialize` was publicly-callable, then someone could take over the PolygonBridgeReceiver by setting their own contract as `mainnetTimelock`

scenario/GovernanceScenario.ts
```diff
@@ -147,3 +151,53 @@ scenario('add new asset',
...
+  {
+    filter: async (ctx) => ctx.world.base.network === 'mumbai'
+  },
+  async ({ timelock, bridgeReceiver }, _context, world) => {
+    const governanceDeploymentManager = world.auxiliaryDeploymentManager;
+    if (!governanceDeploymentManager) {
+      throw new Error("cannot execute governance without governance deployment manager");
+    }
+
+    const proposer = await impersonateAddress(governanceDeploymentManager, COMP_WHALES.testnet[0]);
```
kevincheng96, line 166: You can just call `await context.getProposer()`

contracts/bridges/polygon/PolygonBridgeReceiver.sol
scott-silver: If we want this function to be pubicly-callable/to get rid of `initializer`, then I think the flow would have to be something like this:

- Timelock is deployed (with an EOA as admin)
- PolygonBridgeReceiver is deployed, setting both `mainnetTimelock` and `l2Timelock` in the constructor
- Timelock sets PolygonBridgeReceiver as pending admin
- PolygonBridgeReceiver.initialize function calls `timelock.acceptAdmin()`

scenario/utils.ts
kevincheng96: Also, this logic is duplicated in `CometContext.getCompWhales()`, but I see we're not passing in context here so it's probably easier just to leave it as is

scenario/utils.ts
```diff
@@ -254,3 +266,178 @@ export async function fetchLogs(
...
+  governanceDeploymentManager: DeploymentManager,
+  bridgeDeploymentManager: DeploymentManager
+) {
+  const bridgeNetwork = bridgeDeploymentManager.network;
+  switch (bridgeNetwork) {
+    case 'mumbai':
+      await relayMumbaiMessage(governanceDeploymentManager, bridgeDeploymentManager);
+      break;
+    default:
+      throw new Error(`No governance execution strategy for network: ${bridgeNetwork}`);
```
kevincheng96, line 410: nit: `No message relay implementation from ${governanceNetwork} -> ${bridgetNetwork}`

scenario/utils.ts
```diff
@@ -254,3 +266,178 @@ export async function fetchLogs(
...
+  await fastGovernanceExecute(
+    governanceDeploymentManager,
+    proposer,
+    targets,
+    values,
+    signatures,
+    calldatas
+  );
+
+  await relayMessage(governanceDeploymentManager, bridgeDeploymentManager);
```
kevincheng96, line 432: Wouldn't `relayMessage` have to be called before `fastGovernanceExecute` to initiate the event listener on `fxRoot`?

scenario/utils.ts
```diff
@@ -254,3 +266,178 @@ export async function fetchLogs(
...
+}
+
+export async function executeOpenProposalAndRelay(
+  governanceDeploymentManager: DeploymentManager,
+  bridgeDeploymentManager: DeploymentManager,
+  openProposal: OpenProposal
+) {
+
+  await executeOpenProposal(governanceDeploymentManager, openProposal);
+  await relayMessage(governanceDeploymentManager, bridgeDeploymentManager);
```
kevincheng96, line 442: Same comment about ordering

plugins/scenario/worker/Worker.ts
kevincheng96: Right, which should be a pretty rare case, but should be handled nonetheless

scenario/AllowBySigScenario.ts
jflatow: Great thought!

plugins/scenario/worker/Worker.ts
jflatow: If aux is meant for gov chain, I don't see why we'd ever need to deploy there?

deployments/mumbai/usdc/roots.json
scott-silver: - [ ] synced offline with Kevin: going to move this fxChild address to be a constant inside the mumbai deploy script (since it's strange to have a value in roots.json that is required to exist before you can deploy

deployments/goerli/usdc/roots.json
scott-silver: synced offline.

- [ ] better approach seems like it would be to add `fxRoot` to the Goerli deploy script, so that this `fxRoot` address gets added automatically to `deployment/goerli/usdc/roots.json`

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: synced offline with Kevin.

we realized that we really need this to be storing the equivalent to the Proposals on GovernorBravo: a set of transactions that should be executed atomically

- [ ] implement Proposals for BaseBridgeReceiver

contracts/bridges/polygon/PolygonBridgeReceiver.sol
scott-silver: synced offline. going to remove the initializer (sounds like we don't mind that it's theoretically griefable)

- [ ] remove initializer

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: going to remove `initializer`

contracts/bridges/polygon/PolygonBridgeReceiver.sol
kevincheng96: I clarified this with Scott offline. Basically, the flow for setting up these contracts would not need to differ if we made `initialize` public. It would still be:

1. Deploy the BridgeReceiver
2. Deploy the Timelock with BridgeReceiver address as admin
3. Call BridgerReceiver.initialize with the mainnetTimelock and l2Timelock addresses

The difference is that `initialize` is public and we don't have to pass in an `initializer` address in the constructor (which I think is a lot cleaner, since `BaseBridgeReceiver` can have no constructor args). We will have an `initialized` boolean that flips to true after `initialize()` is called so it can only ever be called once.

In terms of griefing, the risk is minimal. Since all of these steps happen within seconds/minutes of each other in the deploy script, a griefer would have to call `initialize` in the moments between the `BridgeReceiver` being deployed and `initialize()` being called. Even if that happens, the deploy script would just fail when attempting to call `initialize()` (because it has already been initialized by the griefer) and we can just run the deploy script again.

deployments/mumbai/usdc/roots.json
jflatow: We are doing the same thing with USDC and gov for the goerli deploy, not really anything too strange about making it a root to override it... but for something that's *always* required to exist, I think this makes sense

contracts/bridges/BaseBridgeReceiver.sol
kevincheng96: Discussed this offline with Scott. He brought up a good point that we could actually store these actions as proposals instead. That way, proposals keep their atomic, all-or-nothing property. It was kind of an oversight that we initially allowed different actions from the same proposal to be executed individually (selectively and in any order).

contracts/vendor/manifest.json
kevincheng96: Might be a question for 👨‍🍳 🍯

jflatow: Probably fine, not sure how much it matters and to who

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: updated; `exectureTransaction(/.. ../)` is now `executeProposal(proposalId)`

added Proposal struct type, a `state(proposalId)` function, Proposal events and Proposal errors.

contracts/bridges/BaseBridgeReceiver.sol
jflatow: `GovTimelock` and `LocalTimelock`?

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: removed

contracts/bridges/polygon/PolygonBridgeReceiver.sol
scott-silver: removed initializer

deployments/mumbai/usdc/deploy.ts
```diff
@@ -0,0 +1,192 @@
...
+  // Deploy L2 Timelock
+  const l2Timelock = await deploymentManager.deploy(
+    'timelock',
+    'vendor/Timelock.sol',
+    [
+      bridgeReceiver.address,        // admin
+      2 * secondsPerDay,             // delay
+      14 * secondsPerDay,            // grace period
+      2 * secondsPerDay,             // minimum delay
+      30 * secondsPerDay             // maxiumum delay
```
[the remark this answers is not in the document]
scott-silver: 10 minutes?

scenario/utils.ts
scott-silver: synced offline; yeah, you'd think so, but apparently Ethers does some retroactive searching for events because setting the listener after the event has already occurred still returns the desired event 🤷

scenario/utils.ts
scott-silver: updated

scenario/GovernanceScenario.ts
scott-silver: if i'm not mistaken, `context.getProposer` would return you the signer for Mumbai, but here we need the signer for the governance chain, Goerli

scenario/utils.ts
scott-silver: updated

deployments/mumbai/usdc/deploy.ts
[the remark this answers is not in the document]
kevincheng96: Fine by me 👍

scenario/GovernanceScenario.ts
kevincheng96: Ah I see, that's right

plugins/scenario/utils/index.ts
scott-silver: my bad; updated

deployments/mumbai/usdc/configuration.json
scott-silver: sounds good. updated

deployments/mumbai/usdc/configuration.json
scott-silver: reduced to 100,000

contracts/bridges/BaseBridgeReceiver.sol
kevincheng96: Yeah that works!

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,134 @@
...
+    event Initialized(address indexed l2Timelock, address indexed mainnetTimelock);
+    event NewL2Timelock(address indexed oldL2Timelock, address indexed newL2Timelock);
+    event NewMainnetTimelock(address indexed oldMainnetTimelock, address indexed newMainnetTimelock);
+    event ProposalCreated(address indexed messageSender, uint id, address[] targets, uint[] values, string[] signatures, bytes[] calldatas, uint eta);
+    event ProposalExecuted(uint id);
+
+    address public mainnetTimelock;
+    address public l2Timelock;
+    bool public initialized;
+
```
kevincheng96, line 22: nit: extra line

deployments/mumbai/usdc/configuration.json
```diff
@@ -0,0 +1,59 @@
...
+    "borrowKink": 0.8,
+    "borrowSlopeLow": 0.035,
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000011574074074074073e15",
+    "baseBorrowSpeed": "0.0011458333333333333e15",
+    "baseMinForRewards": "100000e6"
```
kevincheng96, line 23: Tbh, even 100k feels a bit high since the faucet drips 10k at a time right? Maybe just set it as 10k?

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+import "../ITimelock.sol";
+
+contract BaseBridgeReceiver {
```
kevincheng96, line 6: Would be nice to add some natspecs to these contracts. You can also just leave a todo for us to fill out later

contracts/bridges/BaseBridgeReceiver.sol
kevincheng96: Correct, just bringing it up since `GovernorBravo` has this check. But I'm not entirely sure why the check is need. Maybe 🍯 can provide some context?

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,134 @@
...
+            targets: targets,
+            values: values,
+            signatures: signatures,
+            calldatas: calldatas,
+            eta: eta,
+            executed: false
+        });
+
+        proposals[proposal.id] = proposal;
+        emit ProposalCreated(messageSender, proposal.id, targets, values, signatures, calldatas, eta);
```
kevincheng96, line 110: Or alternatively, the second argument of the event can just be the `Proposal` struct.

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -0,0 +1,134 @@
...
+        proposal.executed = true;
+        for (uint i = 0; i < proposal.targets.length; i++) {
+            ITimelock(l2Timelock).executeTransaction(proposal.targets[i], proposal.values[i], proposal.signatures[i], proposal.calldatas[i], proposal.eta);
+        }
+        emit ProposalExecuted(proposalId);
+    }
+
+    function state(uint proposalId) public view returns (ProposalState) {
+        if (proposalId > proposalCount || proposalId == 0) revert InvalidProposalId();
+        Proposal storage proposal = proposals[proposalId];
```
kevincheng96, line 125: Can this be `Proposal memory` instead? We aren't applying any state changes to this variable in this function

kevincheng96 (commented):
Nice, overall the new changes LGTM! Just have a few comments and open questions.

deployments/mumbai/usdc/deploy.ts
scott-silver: i updated `deploymentManager.existing` to allow you to specify a network other than "mainnet", and now we pull in `fxChild` like so and then add it to roots:

```
const fxChild = await deploymentManager.existing('fxChild', FX_CHILD, 'mumbai');
```

deleted the hard-coded mumbai roots

deployments/mumbai/usdc/configuration.json
scott-silver: sounds good. updated

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: we're setting `proposal.executed = true;` on the next line. that would require it to be storage, right?

scott-silver: yep, updated

deployments/mumbai/usdc/deploy.ts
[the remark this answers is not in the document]
scott-silver: updated

plugins/scenario/worker/Worker.ts
scott-silver: just encountered the need for this to be a redeploy. was adding something to the Goerli deploy.ts and needed the newly-added root to exist when running Mumbai scenarios.

I've added back in the runDeployScript for the auxiliary deployment

deployments/goerli/usdc/roots.json
scott-silver: removed hard-coded `fxRoot` from Goerli roots; using `.existing` to pull it in as part of the Goerli deploy script

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: going to add this check

scenario/AllowBySigScenario.ts
scott-silver: Cherry-picked your change, Jared: https://github.com/compound-finance/comet/pull/569/commits/a0cb7876c48e2bbbad88f02dd5ef20a8c4f6dfa3

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: updated

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: added

plugins/scenario/worker/Worker.ts
jflatow: I feel like this is more likely to lead to an error than prevent one, and in your case you should just run the deploy script changes for goerli so you get the address you need in roots/aliases for real

scenario/utils.ts
```diff
@@ -254,3 +266,183 @@ export async function fetchLogs(
...
+      { gasPrice: 0 }
+    )
+  ).wait();
+  const proposeEvent = proposeTxn.events.find(event => event.event === 'ProposalCreated');
+  const [id, , , , , , startBlock, endBlock] = proposeEvent.args;
+
+  await executeOpenProposal(dm, { id, startBlock, endBlock });
+}
+
+async function relayMumbaiMessage(
```
jflatow, line 355: I wonder if we should have files for each set of bridge-specific stuff 🤔

jflatow (approved):
This is great work 💪  Probably needs a rebase and commented on one more thing, but think we could basically merge this any time now to build upon (and eventually with the real deployment PR the one having the actual roots.json)

deployments/goerli/usdc/roots.json
```diff
@@ -8,6 +8,5 @@
     "comet": "0x44Ac0af6F39c98d59Ed6D16404210656dB912Ffd",
     "configurator": "0xdE0e9466F406Eb0922746861Bc09785bED912C80",
     "rewards": "0xef9e070044d62C38D2e316146dDe92AD02CF2c2c",
-    "bulker": "0x2677D9ca7477A81A394C324B97F3D7907C6727Fb",
-    "fxRoot": "0x3d1d3e34f7fb6d26245e6640e1c50710efff15ba"
```
kevincheng96: If `deploy` is run again, it will add `fxRoot` back into `roots.json` right?

scott-silver: yep, it should

kevincheng96 (approved):
Looks great! A massive undertaking 💪

scenario/utils.ts
kevincheng96: Yeah that would be a cleaner way to organize things, since each `relayXMessage` is going to be pretty large.

scott-silver: right now, moving `relayMumbaiMessage` out of `utils.ts` creates a circular dependency with some of the helper functions.

i'll move some of these things around. going to do it in a follow-up PR though

PR #563, opened 2022-09-14
Attempt to fix how we import cloned contracts and set verify args

jflatow:
Also add some more options to control deploy task:
 `--no-deploy`: skip actually deploying anything
 `--no-verify-impl`: don't verify the Comet impl


tasks/deployment_manager/task.ts
```diff
@@ -48,10 +48,12 @@ async function runMigration<T>(
 
 task('deploy', 'Deploys market')
   .addFlag('simulate', 'only simulates the blockchain effects')
+  .addFlag('noDeploy', 'skip the actual deploy step')
```
kevincheng96, line 51: IMO, it's more intuitive if we create a new task in this file called `verify` that only runs the verify process (the `deploy` task will still call verify as well). Running `deploy` with a `noDeploy` flag feels kind of unintuitive.

kevincheng96 (commented):
Verify fixes look promising!

tasks/deployment_manager/task.ts
kevincheng96: Yeah sure thing, a separate PR is fine

plugins/import/import.ts
```diff
@@ -78,21 +88,68 @@ async function getContractCreationCode(network: string, address: string) {
   return matches[0][1];
 }
 
+function parseSources({ source, contract, optimized, optimizationRuns }: EtherscanData) {
```
hayesgm, line 91: This is crazy, but hazaah!

hayesgm (approved):

plugins/import/import.ts
```diff
@@ -78,21 +88,68 @@ async function getContractCreationCode(network: string, address: string) {
...
+    return {
+      language: 'Solidity',
+      settings: {
+        optimizer: {
+          enabled: optimized,
+          runs: optimizationRuns,
+        }
+      },
+      sources: {
+        [`contracts/${contract}.sol`]: {
```
scott-silver, line 119: should this be:

````
`contracts/${contract}.sol:${contract}`
```

?

scott-silver (approved):

plugins/deployment_manager/ManualVerify.ts
```diff
@@ -54,14 +55,15 @@ export async function manualVerifyContract(
     settings.optimizer.runs = 1000000;
   }
 
-  let request = toVerifyRequest({
+  const args = await encodeArguments(contractMetadata.abi, sourceName, contractFQN, deployArgs);
```
kevincheng96, line 58: Hmm, I'm a bit confused about this part. Shouldn't we just be encoding the `deployArgs`?

kevincheng96: Actually nevermind, just saw that this is a `hardhat-etherscan` function that takes these specific args

kevincheng96 (approved):
👍

plugins/deployment_manager/ManualVerify.ts
jflatow: Isn't that what `encodeArguments` does?

jflatow: https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-etherscan/src/ABIEncoder.ts#L10

plugins/import/import.ts
jflatow: Not according to etherscan, I think

plugins/deployment_manager/ManualVerify.ts
jflatow: > Actually nevermind, just saw that this is a `hardhat-etherscan` function that takes these specific args

Ah this comment wasn't showing before 👍

jflatow:
Merged by #565

PR #565, opened 2022-09-14
Goerli deployment with prior roots

jflatow:
This will supersede #558, which does not use the pre-existing gov/asset contracts


hayesgm (approved):
lgtm

scott-silver (approved):
🙌

PR #566, opened 2022-09-15
Test enacted CI

kevincheng96:

PR #567, opened 2022-09-15
Add back `enacted` step and programmatically set it after running migrations

kevincheng96:
This PR adds back an optional `enacted` step in our migrations and programmatically sets it whenever a migration is run with `enact` on. We are bringing `enacted` back so we have a way to prevent scenarios from running the same migration again once it has already been enacted on-chain.

Example commits of GHA adding `enacted` after running a migration:
https://github.com/compound-finance/comet/pull/566/commits/04aab559697102d09d580fb688bfa6012cc13555
https://github.com/compound-finance/comet/pull/566/commits/16b3daf885609aae3ff48c604941408315500153

### DIFF
--- a/plugins/deployment_manager/Enacted.ts
+++ b/plugins/deployment_manager/Enacted.ts
@@ -1,0 +1,136 @@
+import { DeploymentManager } from './DeploymentManager';
+import * as ts from 'typescript';
+import * as fs from 'fs/promises';
+import { Migration } from './Migration';
+
+export async function writeEnacted<T>(migration: Migration<T>, deploymentManager: DeploymentManager, writeToFile: boolean = true): Promise<string> {
+  const network = deploymentManager.network;
+  const deployment = deploymentManager.deployment;
+  const migrationPath = `./deployments/${network}/${deployment}/migrations/${migration.name}.ts`;
+  const program = ts.createProgram([migrationPath], { allowJs: true });
+  const sourceFile = program.getSourceFile(migrationPath);
+  const newSourceCode = addEnactedToMigration(sourceFile);
+
+  if (writeToFile) {
+    await fs.writeFile(migrationPath, newSourceCode);
+  }
+
+  return newSourceCode;
+}
+
+export function addEnactedToMigration(sourceFile: ts.SourceFile): string {
+  // XXX The following method directly modifying the AST seems less error-prone,
+  // but unfortunately messes with the original formatting of the file
+  // const enactedMethodDeclaration = ts.factory.createMethodDeclaration(
+  //   [],
+  //   [ts.factory.createModifier(ts.SyntaxKind.AsyncKeyword)],
+  //   undefined,
+  //   'enacted',
+  //   undefined,
+  //   [],
+  //   [
+  //     ts.factory.createParameterDeclaration(
+  //       [],
+  //       [],
+  //       undefined,
+  //       'deploymentManager',
+  //     )
+  //   ],
+  //   undefined,
+  //   ts.factory.createBlock([
+  //     ts.factory.createReturnStatement(ts.factory.createTrue())
+  //   ])
+  // ts.factory.createVariableDeclarationList(
+  //   [ts.factory.createVariableDeclaration(
+  //     ts.factory.createIdentifier("testVar"),
+  //     undefined,
+  //     ts.factory.createKeywordTypeNode(ts.SyntaxKind.StringKeyword),
+  //     ts.factory.createStringLiteral("test")
+  //   )],
+  //   ts.NodeFlags.Const
+  // );
+
+  // const transformerFactory: ts.TransformerFactory<ts.Node> = (
+  //   context: ts.TransformationContext
+  // ) => {
+  //   return (rootNode) => {
+  //     function visit(node: ts.Node): ts.Node {
+  //       node = ts.visitEachChild(node, visit, context);
+  //       if (ts.isExportAssignment(node)) {
+  //         console.log('IS EXPORT ASSIGNMENT')
+  //         console.log(node)
+  //         const callExpression = node.expression as ts.CallExpression;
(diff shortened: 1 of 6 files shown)


plugins/deployment_manager/Enacted.ts
```diff
@@ -0,0 +1,65 @@
...
+      // Prepend a comma
+      insertPos = enact.end;
+      code = ',' + code;
+    }
+    newSourceCode = sourceFileText.substring(0, insertPos)
+      + code
+      + sourceFileText.substring(insertPos);
+  }
+
+  return newSourceCode;
```
jflatow, line 64: We sure we want to edit this file?

tasks/deployment_manager/task.ts
```diff
@@ -152,5 +153,9 @@ task('migrate', 'Runs migration')
       }
 
       await runMigration(dm, prepare, enact, migration, overwrite);
+
+      if (!simulate && enact) {
```
jflatow, line 157: Does this depend on `overwrite` at all?

jflatow (commented):
Nice work

plugins/deployment_manager/Enacted.ts
kevincheng96: I did consider the alternative of having a separate config file to track which migrations have been enacted, but that's an extra file per deployment to manage/commit/read from so I'm leaning more towards this current approach. I think it's easier to interact with migrations when all of the info is in the same migration script.

That being said, curious to hear your thoughts on the approaches.

tasks/deployment_manager/task.ts
kevincheng96: Should it? It seems like `overwrite` is only for controlling writing the migration artifacts, so I'm hesitant to repurpose it for this as well. WDYT about introducing a new flag for writing `enacted` to the migration script? It does seem like we are getting a bit crowded with all the different flags though.

jflatow: 🤔  `!noEnacted && enact` might be nice

kevincheng96: I added a `setEnacted` flag that is always set in the `Enact Migration` task.

plugins/deployment_manager/Enacted.ts
jflatow: This seems like it could be a little fickle/annoying (old wisdom would say not to mix human and machine editing to the same file), but perhaps it will be good

plugins/deployment_manager/test/EnactedTest.ts
```diff
@@ -0,0 +1,101 @@
...
+  },
+
+  async enacted(deploymentManager: DeploymentManager): Promise<boolean> {
+    return true;
+  },
+});
+`;
+
+// XXX add tests for trailing comment
+describe.only('Enacted', () => {
```
jflatow, line 81: Should this still be `.only`?

scenario/constraints/MigrationConstraint.ts
```diff
@@ -62,7 +69,7 @@ export class VerifyMigrationConstraint<T extends CometContext, R extends Require
       async function (ctx: T): Promise<T> {
         for (const migration of ctx.migrations) {
           // XXX does verify get the 'gov' deployment manager as well as the 'local' one?
-          if (migration.actions.verify) {
+          if (migration.actions.verify && !(await isEnacted(migration.actions, ctx.deploymentManager))) {
```
jflatow, line 72: Kinda interesting that we wouldn't want to verify once enacted, makes sense

tasks/deployment_manager/task.ts
```diff
@@ -152,5 +154,9 @@ task('migrate', 'Runs migration')
       }
 
       await runMigration(dm, prepare, enact, migration, overwrite);
+
+      if (enact && setEnacted) {
```
jflatow, line 158: Was suggesting `noEnact` so it would default enable, you think it makes more sense to default disable?

kevincheng96: Good point, converted to `noEnacted`

tasks/deployment_manager/task.ts
kevincheng96: Changed this flag to `noEnacted`

plugins/deployment_manager/Enacted.ts
kevincheng96: Synced offline with Jared. Definitely understand the hesitation with machine editing human-written code, so I'll make sure the tests for this editing operation are comprehensive. The approach I've taken with the current implementation also does not modify the formatting of the source code anymore, so there should always only be a 3-4 line diff in the edit.

plugins/deployment_manager/test/EnactedTest.ts
kevincheng96: Good catch. Removed after adding some more unit tests

scenario/constraints/MigrationConstraint.ts
```diff
@@ -29,15 +34,17 @@ export class MigrationConstraint<T extends CometContext, R extends Requirements>
...
-        // Otherwise the scenario could be running the same proposal twice.
         debug(`${label} Running scenario with migrations: ${JSON.stringify(migrationList.map((m) => m.name))}`);
         for (const migration of migrationList) {
           const artifact = await migration.actions.prepare(ctx.deploymentManager);
           debug(`${label} Prepared migration ${migration.name}.\n  Artifact\n-------\n\n${JSON.stringify(artifact, null, 2)}\n-------\n`);
           // XXX enact will take the 'gov' deployment manager instead of the 'local' one
-          await migration.actions.enact(ctx.deploymentManager, artifact);
-          debug(`${label} Enacted migration ${migration.name}`);
+          if (await isEnacted(migration.actions, ctx.deploymentManager)) {
+            debug(`${label} Migration ${migration.name} has already been enacted`);
```
jflatow, line 43: Nice 👍

jflatow (approved):

jflatow:
I suppose the CI failure/stall is transient?

kevincheng96:
> I suppose the CI failure/stall is transient?

Yep, just transient. They are passing now.

PR #568, opened 2022-09-15
Cross-chain scenarios

scott-silver:
PR that adds the ability to write scenarios that bridge messages from a governance chain (.e.g. Goerli) to another chain (e.g. Mumbai). Implements a test in which a proposal on Goerli updates the `delay` on the Mumbai deployment.

Some notable parts of this change:

- moves construction of the hre from Worker to World
- adds optional `auxiliaryBase` field to scenario bases, used to create second deployment manager on the World
- when constructing World, optionally create `auxiliaryDeploymentManager` for bases with a defined `auxiliaryBase`
- creates independent `setNextBaseFeeToZero`, `setNextBlockTimestamp`, `mineBlocks`, `executeOpenProposal`, and `fastGovernanceExecute` helpers that take a deploymentManager as an argument
- creates a new `fastL2GovernanceExecute` helper that executes a proposal on a governance chain and relays to the bridged chain
- bumps the default scenario stall timeout to from 180_000 -> 240_000

Some refactoring of duplicate references to deploymentManager and hre on Context and World

- removes `CometContext.deploymentManager`; replaces calls to `context.deploymentManager` with `context.world.deploymentManager`
- removes `world.hre`; replaces calls to `world.hre` with `world.deploymentManager.hre`


scenario/utils.ts
```diff
@@ -254,3 +267,199 @@ export async function fetchLogs(
...
+async function relayMumbaiMessage(
+  governanceDeploymentManager: DeploymentManager,
+  bridgeDeploymentManager: DeploymentManager,
+) {
+  const STATE_SENDER = '0xeaa852323826c71cd7920c3b4c007184234c3945';
+  const MUMBAI_RECEIVER_ADDRESSS = '0x0000000000000000000000000000000000001001';
+  const EVENT_LISTENER_TIMEOUT = 60000;
+
+  const deploymentTag = `${bridgeDeploymentManager.network}/${bridgeDeploymentManager.deployment}`;
+  const l2Timelock = await bridgeDeploymentManager.contract('timelock');
```
scott-silver, line 360: rename to `timelock`

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -79,4 +79,14 @@ contract BaseBridgeReceiver {
 
         emit ProcessMessage(messageSender, targets, values, signatures, calldatas, eta);
     }
+
+    function executeTransaction(
```
scott-silver, line 83: open question: should this function be callable by anyone?

if not, then who should be allowed to call it?

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -79,4 +79,14 @@ contract BaseBridgeReceiver {
...
+        address target,
```
kevincheng96, line 84: Wondering if we just make it easier to call `execute`. Instead of passing all these args again, maybe we just provide the hash of the action or some unique id. This would require the `BaseBridgeReceiver` to store a mapping of these actions, but makes it way easier for anyone to call `execute`.

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -79,4 +79,14 @@ contract BaseBridgeReceiver {
...
+        uint value,
+        string memory signature,
+        bytes memory data,
+        uint eta
+    ) public payable returns (bytes memory) {
```
kevincheng96, line 89: `external` instead of `public`?

contracts/bridges/BaseBridgeReceiver.sol
```diff
@@ -79,4 +79,14 @@ contract BaseBridgeReceiver {
...
+        ITimelock(l2Timelock).executeTransaction(target, value, signature, data, eta);
```
kevincheng96, line 90: You need to use `return` here right?

deployments/mumbai/usdc/deploy.ts
```diff
@@ -23,40 +23,40 @@ async function deployContracts(deploymentManager: DeploymentManager, deploySpec:
...
-  const FX_CHILD = "0xCf73231F28B7331BBe3124B907840A94851f9f11"; //
-  const MAINNET_TIMELOCK = "0x6d903f6003cca6255d85cca4d3b5e5146dc33925";
-
-  // Initialize PolygonBridgeReceiver
-  trace(`Initializing PolygonBridgeReceiver`);
-  await polygonBridgeReceiver.initialize(
-    MAINNET_TIMELOCK,       // mainnet timelock
-    l2Timelock.address,     // l2 timelock
-    FX_CHILD                // fxChild
+  const GOERLI_TIMELOCK = "0x339B2D3bf0406DF82f8fa7B0d855a3f47562d8D7";
```
kevincheng96, line 52: Would be neat if we also refactored `deploy` to have multiple deployment managers so we can pull this address dynamically.

plugins/scenario/utils/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  await dm.hre.network.provider.send('hardhat_setNextBlockBaseFeePerGas', ['0x0']);
+}
+
+export async function mineBlocks(dm: DeploymentManager, blocks: number) {
+  await dm.hre.network.provider.send('hardhat_mine', [`0x${blocks.toString(16)}`]);
+}
+
+export async function setNextBlockTimestamp(dm: DeploymentManager, timestamp: number) {
+  await dm.hre.ethers.provider.send('evm_setNextBlockTimestamp', [timestamp]);
+}
```
kevincheng96, line 22: These could live directly in `scenario/utils.ts`, since they aren't used by the plug-in.

kevincheng96, line 22: These could live directly in scenario/utils.ts, since they aren't used by the plug-in.

kevincheng96, line 22: These could live directly in `scenario/utils.ts`, since they aren't used by the plug-in.

kevincheng96, line 22: These could live directly in `scenario/utils.ts`, since they aren't used by the plug-in.

scenario/GovernanceScenario.ts
```diff
@@ -1,86 +1,157 @@
...
-  );
-
-  const CometModified = await dm.hre.ethers.getContractFactory('CometModified');
-  const modifiedComet = CometModified.attach(comet.address).connect(signer.signer);
-
-  // Call new functions on Comet
-  await modifiedComet.initialize(constants.AddressZero);
-  expect(await modifiedComet.newFunction()).to.be.equal(101n);
-});
+const FX_ROOT_GOERLI = '0x3d1d3E34f7fB6D26245E6640E1c50710eFFf15bA';
```
kevincheng96, line 8: Might be better to have these exist in `aliases` by adding a relation for fetching this.

plugins/scenario/Runner.ts
```diff
@@ -61,7 +61,7 @@ export class Runner<T, U, R> {
 
     // initialize the context and take a snapshot of it
     let context = await scenario.initializer(world);
-    let contextSnapshot = await world._snapshot();
+    let contextSnapshot = this.worldSnapshot;
```
jflatow, line 64: These can only be used once, right?

scenario/GovernanceScenario.ts
```diff
@@ -1,86 +1,157 @@
...
+      governanceDeploymentManager,
+      world.deploymentManager,
+      proposer,
+      [FX_ROOT_GOERLI],
+      [0],
+      ["sendMessageToChild(address,bytes)"],
+      [sendMessageToChildCalldata]
+    );
+
+    expect(await timelock.delay()).to.eq(newTimelockDelay);
```
kevincheng96, line 54: 🙌

plugins/scenario/World.ts
```diff
@@ -1,52 +1,84 @@
...
-    })) as string;
+    }) as string;
+    let auxiliarySnapshot: string;
+    if (this.auxiliaryDeploymentManager) {
+      this.snapshotAuxiliaryDeploymentManager = this.auxiliaryDeploymentManager.fork();
+      auxiliarySnapshot = await this.auxiliaryDeploymentManager.hre.network.provider.request({
+        method: 'evm_snapshot',
+        params: [],
+      }) as string;
+    }
```
jflatow, line 61: Oh hm 🤔

scenario/utils.ts
```diff
@@ -254,3 +267,199 @@ export async function fetchLogs(
...
+  const blockNow = await dm.hre.ethers.provider.getBlockNumber();
+  const blocksUntilStart = startBlock - blockNow;
+  const blocksUntilEnd = endBlock - Math.max(startBlock, blockNow);
+
+  if (blocksUntilStart > 0) {
+    await mineBlocks(dm, blocksUntilStart);
+  }
+
+  if (blocksUntilEnd > 0) {
+    for (const whale of COMP_WHALES) {
```
kevincheng96, line 290: `COMP_WHALES` would be chain dependent, so let's leave a XXX to properly handle this

scenario/utils.ts
```diff
@@ -254,3 +267,199 @@ export async function fetchLogs(
...
+
+  await executeOpenProposal(dm, { id, startBlock, endBlock });
+}
+
+async function relayMumbaiMessage(
+  governanceDeploymentManager: DeploymentManager,
+  bridgeDeploymentManager: DeploymentManager,
+) {
+  const STATE_SENDER = '0xeaa852323826c71cd7920c3b4c007184234c3945';
+  const MUMBAI_RECEIVER_ADDRESSS = '0x0000000000000000000000000000000000001001';
```
kevincheng96, line 356: Could these be dervied via `aliases` as well?

scenario/utils.ts
```diff
@@ -254,3 +267,199 @@ export async function fetchLogs(
...
+    target,
+    value,
+    signature,
+    data,
+    eta,
+    { gasPrice: 0 }
+  );
+}
+
+export async function fastL2GovernanceExecute(
```
kevincheng96, line 439: Nice

kevincheng96 (commented):
Nice work, super cool that we are able to start testing cross-chain proposals end-to-end!

plugins/scenario/Runner.ts
kevincheng96: Yeah

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: updated

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: yep, good catch. updated

deployments/mumbai/usdc/deploy.ts
scott-silver: how would you specify what the second deployment manager should be? via command line maybe:

```
yarn hardhat deploy --network mumbai --deployment usdc --parentNetwork goerli --parentDeployment usdc
```

plugins/scenario/utils/index.ts
scott-silver: they're still used by CometContext, but we could update the all calls to `ctx.mineBlock` to use `mineBlocks` instead

contracts/bridges/BaseBridgeReceiver.sol
scott-silver: I'm open to it. the nice thing about reusing the same arguments are that

- you don't have to store anything on BaseBridgeReceiver
- you can pull all the required arguments directly off of the Timelock's `QueueTransaction` event

plugins/scenario/World.ts
scott-silver: which part are you thinking about?

scenario/GovernanceScenario.ts
scott-silver: deleted `FX_ROOT_GOERLI`; instead I added `fxChild` as a root and then pull the `fxRoot` address from `fxChild`

scenario/utils.ts
scott-silver: yes for `STATE_SENDER`.

I added `fxRoot` to Goerli roots, with a relation to go from fxRoot -> stateSender

`MUMBAI_RECEIVER_ADDRESS` isn't crawlable, however. it's just an address that is hard-coded into `fxChild`, unfortunately

scenario/utils.ts
scott-silver: good call; added

scenario/utils.ts
```diff
@@ -319,11 +319,7 @@ export async function fastGovernanceExecute(
   signatures: string[],
   calldatas: string[]
 ) {
-  const governor = await dm.contract('governor');
-
-  if (!governor) {
-    throw new Error("cannot create governance proposal without governor");
-  }
+  const governor = await dm.getContractOrThrow('governor');
```
jflatow, line 322: Can this just be the behavior or `contract()`?

scott-silver: not without further refactoring: `clone`, `deploy` and `existing` all use `this.contract` in a way that would break if it threw for missing contracts. they check for an existing contract and take different actions if it's not there

plugins/scenario/World.ts
jflatow: Just thinking that I see why the change I commented on above was made (worldSnapshot)

plugins/scenario/utils/index.ts
kevincheng96: Oh I just meant we could move these from the `plugins` directory to the `scenarios` directory, since these are unrelated to the plugin itself.

deployments/mumbai/usdc/deploy.ts
kevincheng96: Hmm good point. Might be a bit too complicated to pass it in via command line, so maybe we should just keep it simple for now and leave that to a future improvement.

contracts/bridges/BaseBridgeReceiver.sol
kevincheng96: That's true, it should be pullable from the `QueueTransaction` event.

PR #569, opened 2022-09-16
Re-deploy Goerli with other USDC

jflatow:


scott-silver:
how come?

jflatow:
> how come?

Folks working on Treasury v3 Adaptor requested we use this other USDC since apparently its the only one supported by Circle now on goerli

scott-silver:
> "USDC": "0x07865c6E87B9F70255377e024ace6630C1Eaa37F",

this looks like a contract that Circle deployed, right? not a clone of that contract that we've deployed?

does this mean that we won't be able to use Fauceteer on Goerli? since we won't be a minter on that instance of USDC?

deployments/goerli/usdc/deploy.ts
```diff
@@ -24,32 +24,10 @@ async function deployContracts(deploymentManager: DeploymentManager, deploySpec:
   // Deploy governance contracts
   const { COMP, fauceteer, timelock } = await cloneGov(deploymentManager);
 
-  const usdcProxyAdmin = await deploymentManager.deploy('USDC:admin', 'vendor/proxy/transparent/ProxyAdmin.sol', []);
   const usdcImpl = await deploymentManager.clone('USDC:implementation', clone.usdcImpl, []);
   const usdcProxy = await deploymentManager.clone('USDC', clone.usdcProxy, [usdcImpl.address]);
-  const usdcProxyAdminSlot = '0x10d6a54a4754c8869d6886b5f5d7fbfa5b4522237ea5c60d11bc4e7a1ff9390b';
   const USDC = usdcImpl.attach(usdcProxy.address);
```
scott-silver, line 29: > const USDC = usdcImpl.attach(usdcProxy.address);

we should be able to delete this entirely, right? we're no longer cloning from mainnet USDC, we're adding an already-deployed USDC to `roots.json`

jflatow:
> > "USDC": "0x07865c6E87B9F70255377e024ace6630C1Eaa37F",
> 
> this looks like a contract that Circle deployed, right? not a clone of that contract that we've deployed?
> 
> does this mean that we won't be able to use Fauceteer on Goerli? since we won't be a minter on that instance of USDC?

Adam is getting a large amount which we can give to the fauceteer

deployments/goerli/usdc/deploy.ts
```diff
@@ -24,32 +24,6 @@ async function deployContracts(deploymentManager: DeploymentManager, deploySpec:
   // Deploy governance contracts
   const { COMP, fauceteer, timelock } = await cloneGov(deploymentManager);
 
-  const usdcProxyAdmin = await deploymentManager.deploy('USDC:admin', 'vendor/proxy/transparent/ProxyAdmin.sol', []);
-  const usdcImpl = await deploymentManager.clone('USDC:implementation', clone.usdcImpl, []);
```
kevincheng96, line 28: You can also remove `usdcImpl` and `usdcProxy` from `clone`.

deployments/goerli/usdc/roots.json
```diff
@@ -1,12 +1,12 @@
 {
     "COMP": "0x3587b2F7E0E2D6166d6C14230e7Fe160252B0ba4",
-    "USDC": "0x8bF5b658bdF0388E8b482ED51B14aef58f90abfD",
+    "USDC": "0x07865c6E87B9F70255377e024ace6630C1Eaa37F",
```
kevincheng96, line 3: I think you'll also need to add `baseTokenAddress: "0x07865c6E87B9F70255377e024ace6630C1Eaa37F"` to `configuration.json` now that we aren't cloning USDC.

jflatow: Hmm, this should mean current deployment is broken in a weird way, wonder if Treasury will hit it before I redeploy...

jflatow: Actually, no, its just referencing by name as it should be https://github.com/compound-finance/comet/blob/main/deployments/goerli/usdc/configuration.json#L4

deployments/goerli/usdc/deploy.ts
scott-silver: he's referring to lines 6 and 7:

```
const clone = {
  usdcImpl: '0xa2327a938Febf5FEC13baCFb16Ae10EcBc4cbDCF',
  usdcProxy: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48',
```

deployments/goerli/usdc/roots.json
kevincheng96: Right, but referencing by name shouldn't work anymore because we aren't cloning `USDC`, so there's no contract address that maps to `USDC` in the cache. That's why I think we'll need to add `baseTokenAddress` to the configuration. Does the deploy script actually work if you wipe out your cache?

jflatow: That's why it's a root?

jflatow: Ugh, roots should be aliases, good point.

jflatow: Ok, should be fixed in the next 2 commits (patch and then re-deploy), looks right on etherscan: https://goerli.etherscan.io/address/0x3EE77595A8459e93C2888b13aDB354017B198188#readProxyContract

deployments/goerli/usdc/deploy.ts
jflatow: Fixed

deployments/goerli/usdc/roots.json
kevincheng96: So I see you addressed the roots/aliases issue, but that's actually not the suggestion I was making. We have the ability to specify token addresses directly in `configuration.json` and I think that's where these hardcoded addresses should live. IMO, deployment scripts shouldn't require there to be pre-existing roots to work. A developer should be able to run a deployment script with a `configuration.json` file and have it just work, creating a `roots.json` in the process. Let's maybe sync on this tmr during parking lot?

deployments/goerli/usdc/deploy.ts
```diff
@@ -21,37 +13,14 @@ async function deployContracts(deploymentManager: DeploymentManager, deploySpec:
...
-  await deploymentManager.idempotent(
-    async () => !sameAddress(await ethers.provider.getStorageAt(usdcProxy.address, usdcProxyAdminSlot), usdcProxyAdmin.address),
-    async () => {
-      trace(`Changing admin of USDC proxy to ${usdcProxyAdmin.address}`);
-      trace(await wait(usdcProxy.connect(signer).changeAdmin(usdcProxyAdmin.address)));
+  // Declare existing assets as aliases
+  const COMP = await deploymentManager.existing('COMP', '0x3587b2F7E0E2D6166d6C14230e7Fe160252B0ba4');
+  const USDC = await deploymentManager.existing('USDC', '0x07865c6E87B9F70255377e024ace6630C1Eaa37F');
+  const WBTC = await deploymentManager.existing('WBTC', '0xAAD4992D949f9214458594dF92B44165Fb84dC19');
+  const WETH = await deploymentManager.existing('WETH', '0x42a71137C09AE83D8d05974960fd607d40033499');
```
kevincheng96, line 20: Should these be declared before `await mintTokens(deploymentManager);`?

kevincheng96 (approved):
Overall looks good! Thanks for making the changes. 

Just one final question about the ordering of declaring existing tokens and calling `mintTokens`.

deployments/goerli/usdc/deploy.ts
jflatow: `COMP` is tied to `cloneGov`, the others could be moved but kinda nice to have them all together?

kevincheng96: Actually I misread, for some reason I thought `mintTokens` was being called before these declarations

PR #570, opened 2022-09-16
Scenario for adding new asset

kevincheng96:


scenario/utils.ts
```diff
@@ -254,3 +254,8 @@ export async function fetchLogs(
     return contract.queryFilter(filter, fromBlock, toBlock);
   }
 }
+
+export function extractCalldata(data: string): string {
+  // Remove the first 4 bytes (function selector) of the transaction data
+  return '0x' + data.slice(10);
+}
```
jflatow, line 261: https://github.com/compound-finance/comet/blob/main/src/deploy/index.ts#L74-L77

jflatow (commented):
💪

scenario/utils.ts
kevincheng96: Oh cool, was wondering where it went. Will use this one instead

kevincheng96: Done

jflatow (approved):
LGTM

PR #571, opened 2022-09-16
Fix scenario timestamp issue

kevincheng96:
This should fix the `Timestamp X is lower than or equal to previous block's timestamp Y` errors when running `executeOpenProposal`. Scenarios on goerli are now failing for other reasons (a set of passed proposals with wrong inputs that fail when being executed).


scenario/context/CometContext.ts
```diff
@@ -279,7 +280,8 @@ export class CometContext {
     }
 
     const proposal = await governor.proposals(id);
-    await this.setNextBlockTimestamp(proposal.eta.toNumber() + 1);
+    const nextBlockTimestamp = max(proposal.eta.toNumber(), await this.world.timestamp());
+    await this.setNextBlockTimestamp(nextBlockTimestamp + 1);
```
jflatow, line 284: 👍

jflatow (approved):

PR #572, opened 2022-09-16
Emit ERC-20 transfer (mint) event on absorb

kevincheng96:
When an absorbed account ends up with a positive balance, we need to emit a ERC-20 transfer event to indicate that some amount of `Comet` tokens are minted.

### DIFF
--- a/contracts/Comet.sol
+++ b/contracts/Comet.sol
@@ -1181,9 +1181,6 @@ contract Comet is CometMainInterface {
 
         int104 newPrincipal = principalValue(newBalance);
         updateBasePrincipal(account, accountUser, newPrincipal);
-        if (newPrincipal > 0) {
-            emit Transfer(address(0), account, presentValueSupply(baseSupplyIndex, unsigned104(newPrincipal)));
-        }
 
         // reset assetsIn
         userBasic[account].assetsIn = 0;
@@ -1198,6 +1195,10 @@ contract Comet is CometMainInterface {
         uint256 basePaidOut = unsigned256(newBalance - oldBalance);
         uint256 valueOfBasePaidOut = mulPrice(basePaidOut, basePrice, uint64(baseScale));
         emit AbsorbDebt(absorber, account, basePaidOut, valueOfBasePaidOut);
+
+        if (newPrincipal > 0) {
+            emit Transfer(address(0), account, presentValueSupply(baseSupplyIndex, unsigned104(newPrincipal)));
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


contracts/Comet.sol
```diff
@@ -1181,6 +1181,9 @@ contract Comet is CometMainInterface {
 
         int104 newPrincipal = principalValue(newBalance);
         updateBasePrincipal(account, accountUser, newPrincipal);
+        if (newBalance > 0) {
+            emit Transfer(address(0), account, uint256(newBalance));
```
kevincheng96, line 1185: Wondering if the amount should actually be `presentValueSupply(baseSupplyIndex, newPrincipal)` to protect against any rounding/truncation?

kevincheng96: Yeah, thinking about it more, we'll need to do that instead of directly using `newBalance` as the amount to protect against rounding down.

kevincheng96: Modified

jflatow (commented):
Nice!

contracts/Comet.sol
```diff
@@ -1181,6 +1181,9 @@ contract Comet is CometMainInterface {
 
         int104 newPrincipal = principalValue(newBalance);
         updateBasePrincipal(account, accountUser, newPrincipal);
+        if (newPrincipal > 0) {
+            emit Transfer(address(0), account, presentValueSupply(baseSupplyIndex, unsigned104(newPrincipal)));
+        }
```
jflatow, line 1186: Looks good 💪

jflatow: Should we move it lower down?

kevincheng96: Sure, we could. Should we put it at the very end, after `emit AbsorbDebt`?

I put it here because the transfer is actually happening in `updateBasePrincipal`.

kevincheng96: Done

kevincheng96:
All scenarios that check a custom error are failing because the local `Comet` contract now differs from what's on-chain. Usually, we just set `upgrade: true` to circumvent this, but I'll spend some time to see if there' a better way.

kevincheng96:
> All scenarios that check a custom error are failing because the local `Comet` contract now differs from what's on-chain. Usually, we just set `upgrade: true` to circumvent this, but I'll spend some time to see if there's a better way.

The custom error scenarios have been patched by 🍯  in #594 and all scenarios are now passing on this branch after the rebase.

jflatow:
Merged #599

PR #574, opened 2022-09-19
Separate whales for testnet

jflatow:
For prior testnets, we completely controlled the COMP allocation, so we were able to delegate to the top mainnet whale when cloning gov. For new goerli, we are leveraging an existing gov installation, so we need to know about some external COMP whales. The current mechanism is sort of a compromise to keep complexity down and prevent scens from being even slower. We shouldn't necessarily need any sets of whales beyond 'mainnet' and 'testnet'.

### DIFF
--- a/scenario/context/CometContext.ts
+++ b/scenario/context/CometContext.ts
@@ -67,8 +67,12 @@ export class CometContext {
     this.assets = {};
   }
 
+  async getCompWhales(): Promise<string[]> {
+    return COMP_WHALES[this.world.base.name === 'mainnet' ? 'mainnet' : 'testnet'];
+  }
+
   async getProposer(): Promise<SignerWithAddress> {
-    return this.world.impersonateAddress(COMP_WHALES[0]);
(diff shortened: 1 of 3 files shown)


kevincheng96 (approved):
👍

jflatow:
> This is great!
> 
> The change does assume that Goerli will be our governance testnet. But that's true for Polygon, Arbitrum, and Optimism so it's probably a safe assumption for now.

If we deploy another gov clone somewhere it would ideally either work like the previous testnets, or we could give some COMP to the deployer address we've been using, so I think its actually a bit weaker assumption currently

kevincheng96:
Hmm, we seem to be seeing the same issue for `fuji` and `kovan` now instead of `goerli`.

kevincheng96:
Ah, I think it's because we already delegated votes to the mainnet whale addresses on `fuji` and `kovan`, so we just have to redelegate to the testnet whale for things to work properly again.

kevincheng96:
I re-delegated COMP votes to the testnet whale on `fuji` and `kovan` so hopefully it should work now.

PR #575, opened 2022-09-20
Cross-chain migrations

scott-silver:
Updating the migration flow to reflect the changes for L2s/bridged chains.

In migrations, the `prepare` step will receive a `deploymentManager` for the current chain.

The `enact` step will receive a `governanceDeploymentManager` for the governing chain (most likely the mainnet deployment or the Goerli deployment).

Requires extending `hreForBase` to return non-forked HREs (the current implementation assumes that the HRE you're constructing should be a fork).

The migration task determines whether a given deployment is a bridged deployment using the `scenario.bases` config, which is somewhat strange but is currently the only place that we store whether a given deployment is governed by a different chain.

### DIFF
--- a/tasks/deployment_manager/task.ts
+++ b/tasks/deployment_manager/task.ts
@@ -15,6 +15,7 @@ function getForkEnv(env: HardhatRuntimeEnvironment): HardhatRuntimeEnvironment {
 
 async function runMigration<T>(
   deploymentManager: DeploymentManager,
+  governanceDeploymentManager: DeploymentManager,
   prepare: boolean,
   enact: boolean,
   migration: Migration<T>,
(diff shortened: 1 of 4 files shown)


tasks/deployment_manager/task.ts
```diff
@@ -152,6 +153,27 @@ task('migrate', 'Runs migration')
       );
       await dm.spider();
 
+      let governanceDm: DeploymentManager;
+      const base = env.config.scenario.bases.find(b => b.network === network && b.deployment === deployment);
```
scott-silver, line 157: somewhat strange to pul this information from scenario bases, although `env.config.scenario.bases` is already a dependency of the `runMigration` task

jflatow: Yea... bases should probably be part of deployment manager, but its just a config namespace I guess? We could revisit separately

kevincheng96 (approved):
👍

jflatow (approved):

PR #577, opened 2022-09-22
Revive old migrations

kevincheng96:
🧟 

Now that we have `enacted` back (#567), we can bring back old migrations without breaking our scenarios. Having old migrations exist directly in the repo is nice because we can directly refer to them instead of having to look through the commit history.


jflatow (approved):
💪

PR #578, opened 2022-09-22
Optimize scenario CI run

kevincheng96:
This PR decreases the run time of our CI runs from ~30min to ~5-10min by splitting up the single scenario job into separate jobs per base. This also means that flaky scenarios for a single base can be retried individually without having to re-run scenarios for all other bases. I also removed some jobs that look redundant.

This change does mean that each base will now have their own scenario report, but that's not a big deal because they can all still be accessed pretty easily ([example run](https://github.com/compound-finance/comet/pull/578/checks?check_run_id=8488711261)).


.github/workflows/run-scenarios.yaml
```diff
@@ -4,6 +4,9 @@ on:
   pull_request:
 jobs:
   run-scenarios:
+    strategy:
+      matrix:
+        bases: [ development, mainnet, goerli, kovan, fuji ]
```
jflatow, line 9: Nice, I think we could even drop kovan now?

.github/workflows/scenario-report.yaml
```diff
@@ -1,16 +0,0 @@
...
-jobs:
-  scenario-report:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: dorny/test-reporter@v1
-      with:
-        artifact: scenario-results        # Artifact name
-        name: Scenario Tests              # Name of the check run which will be created
-        path: 'scenario-results.json'     # Path to test results (inside artifact .zip)
-        reporter: mocha-json              # Format of test results
```
jflatow, line 16: So we didn't need these?

jflatow (approved):
Awesome 🙌

.github/workflows/scenario-report.yaml
kevincheng96: I don't think we do, since we are actually running the same exact step at the end of `run-scenarios.yaml` (and similarly, in `run-unit-tests.yaml`).

.github/workflows/run-scenarios.yaml
kevincheng96: Sure, I'll remove it.

kevincheng96: Done

scott-silver (approved):
🥳

PR #580, opened 2022-09-22
Proposal to add LINK to Goerli

kevincheng96:
This adds a new migration script for adding LINK as a collateral asset to Goerli.

### DIFF
--- a/scenario/SupplyScenario.ts
+++ b/scenario/SupplyScenario.ts
@@ -77,6 +77,9 @@ for (let i = 0; i < MAX_ASSETS; i++) {
   scenario(
     `Comet#supply > collateral asset ${i}`,
     {
+      // XXX Unfortunately, the filtering step happens before solutions are run, so this will filter out
+      // hypothetical assets added during the migration/proposal constraint because those assets don't exist
+      // yet
       filter: async (ctx) => await isValidAssetIndex(ctx, i) && await isTriviallySourceable(ctx, i, amountToSupply),
       tokenBalances: {
         albert: { [`$asset${i}`]: amountToSupply },
--- a/scenario/constraints/ProposalConstraint.ts
+++ b/scenario/constraints/ProposalConstraint.ts
@@ -60,6 +60,10 @@ export class ProposalConstraint<T extends CometContext, R extends Requirements> 
           throw(err);
         }
       }
+
+      // Re-set the assets in case they were updated via a proposal
+      await ctx.setAssets();
+
       return ctx;
     };
   }
--- a/deployments/goerli/usdc/migrations/1663870047_add_link.ts
+++ b/deployments/goerli/usdc/migrations/1663870047_add_link.ts
@@ -1,0 +1,101 @@
+import { DeploymentManager, migration } from '../../../../plugins/deployment_manager';
+import { calldata, exp, proposal, wait } from '../../../../src/deploy';
+
+import { expect } from 'chai';
+
+const clone = {
+  link: '0x514910771af9ca656af840dff83e8264ecf986ca',
+};
+
+const LINK_PRICE_FEED = '0x48731cF7e84dc94C5f84577882c14Be11a5B7456';
+
+export default migration('1663870047_add_link', {
+  async prepare(deploymentManager: DeploymentManager) {
+    return {};
+  },
+
+  async enact(deploymentManager: DeploymentManager) {
+    const trace = deploymentManager.tracer();
+    const signer = await deploymentManager.getSigner();
+
+    const {
+      governor,
+      comet,
+      configurator,
+      cometAdmin,
+      fauceteer,
+    } = await deploymentManager.getContracts();
+
+    // Clone LINK and send half of total supply to the fauceteer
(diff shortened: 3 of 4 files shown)


scenario/SupplyScenario.ts
```diff
@@ -75,6 +75,9 @@ for (let i = 0; i < MAX_ASSETS; i++) {
   scenario(
     `Comet#supply > collateral asset ${i}`,
     {
+      // XXX Unfortunately, the filtering step happens before solutions are run, so this will filter out
+      // hypothetical assets added during the migration/proposal constraint because those assets don't exist
+      // yet
```
kevincheng96, line 80: Any ideas on how we should address this? In the runner, solutions are generated *before* any solutions (e.g. executing on-chain proposals in the ProposalConstraint) are run, so the FilterConstraint can only filter based off of the on-chain state before solutions are run.

kevincheng96:
Interesting, the `has assets` scenario is failing on Goerli because the proposal constraint has added a new asset so there is a length mismatch: 

```
[goerli] ❌ has assets [1 run]: Error AssertionError: expected [ Array(4) ] to have the same members as [ Array(3) ]
    at Scenario.property (/home/runner/work/comet/comet/scenario/CometScenario.ts:17:54)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at runNextTicks (node:internal/process/task_queues:65:3)
    at listOnTimeout (node:internal/timers:528:9)
    at processTimers (node:internal/timers:502:7)
    at async Runner.run (/home/runner/work/comet/comet/plugins/scenario/Runner.ts:90:26)
    at async /home/runner/work/comet/comet/plugins/scenario/worker/Worker.ts:90:22
```

It's comparing the assets from Comet with the assets from `CometContext`. Should `CometContext.assets` be updated after the proposal gets executed? I think we explicitly said that constraints should not touch context.

jflatow:
> Interesting, the `has assets` scenario is failing on Goerli because the proposal constraint has added a new asset so there is a length mismatch:
> 
> ```
> [goerli] ❌ has assets [1 run]: Error AssertionError: expected [ Array(4) ] to have the same members as [ Array(3) ]
>     at Scenario.property (/home/runner/work/comet/comet/scenario/CometScenario.ts:17:54)
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (node:internal/process/task_queues:96:5)
>     at runNextTicks (node:internal/process/task_queues:65:3)
>     at listOnTimeout (node:internal/timers:528:9)
>     at processTimers (node:internal/timers:502:7)
>     at async Runner.run (/home/runner/work/comet/comet/plugins/scenario/Runner.ts:90:26)
>     at async /home/runner/work/comet/comet/plugins/scenario/worker/Worker.ts:90:22
> ```
> 
> It's comparing the assets from Comet with the assets from `CometContext`. Should `CometContext.assets` be updated after the proposal gets executed? I think we explicitly said that constraints should not touch context.

I think constraints can definitely touch context? Properties don't need to modify it

scenario/SupplyScenario.ts
jflatow: We could have an `AssetConstraint` to add it 🤔

jflatow: Ah nevermind, that doesn't help the filter

kevincheng96:
> I think constraints can definitely touch context? Properties don't need to modify it

Ok yeah, I think I misunderstood this [comment](https://github.com/compound-finance/comet/blob/1a663de8ecd40b0f8f3267612ae8c4cafd72b2dc/plugins/scenario/Runner.ts#L67-L68) in `Runner.ts`. It's stating that the `solve` function should not touch context, not that the solution itself cannot.

scenario/SupplyScenario.ts
jflatow: I think the solution is what we said we should prob do anyway which is put these in a single scenario?

kevincheng96: I tried to make that change a few week backs, but it has its own set of challenges. We need to conditionally source collateral assets in these scenarios, which we can't do if we use only a single scenario. We'll have to alter the way balance and token balance constraints work so they can conditionally source assets instead of relying on a filter constraint. But adding conditional logic to those constraints also feels kind of off.

scenario/constraints/ProposalConstraint.ts
```diff
@@ -60,6 +60,10 @@ export class ProposalConstraint<T extends CometContext, R extends Requirements>
           throw(err);
         }
       }
+
+      // Re-set the assets in case they were updated via a proposal
+      await ctx.setAssets();
```
jflatow, line 65: Good point 🎣

jflatow (approved):

kevincheng96:
I enacted the migration locally because we are seeing some `HeadersTimeoutError`s in CI when running enact. Here's the transaction that made the proposal: https://goerli.etherscan.io/tx/0xbfee0b4028f3e34a0258c927c83be16607b0b1a4c1811c9866209203e7e94bdc

I voted for, queued, and executed the proposal successfully so LINK is now a supported asset on Goerli.

PR #581, opened 2022-09-22
Disable fail-fast in matrix configuration

kevincheng96:
This means that one matrix job failing will no longer cancel all the other running matrix jobs.


jflatow (commented):
👍

scott-silver (approved):

PR #585, opened 2022-09-23
pull hardhat-cover from master branch

scott-silver:
So we can get patch: https://github.com/compound-finance/hardhat-cover/commit/c9064e8bf04d3ae34773adbfee70cade741078ec

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5382,7 +5382,7 @@ hardhat-contract-sizer@^2.4.0:
 
 hardhat-cover@compound-finance/hardhat-cover:
   version "1.0.0"
-  resolved "https://codeload.github.com/compound-finance/hardhat-cover/tar.gz/13f93694d92ba78a846af7ecd9f5347dbb2bd5b4"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "fast-glob": "^3.2.7",
     "hardhat": "https://github.com/jflatow/hardhat/releases/download/viaIR/hardhat-v2.9.9.tgz",
     "hardhat-contract-sizer": "^2.4.0",
-    "hardhat-cover": "compound-finance/hardhat-cover",
+    "hardhat-cover": "compound-finance/hardhat-cover#master",
```
jflatow, line 76: Isn't this redundant? Isn't it better to just rely on whatever default branch is (i.e. master)

scott-silver: yep, no need to update the `package.json`. used `yarn hardhat upgrade hardat-cover` instead to just bump the lock file

kevincheng96 (approved):

PR #586, opened 2022-09-23
move relayMumbaiMessage, other scenario utils into files

scott-silver:
Follow-up from https://github.com/compound-finance/comet/pull/561#discussion_r977022147


kevincheng96 (commented):
LGTM, but seeing a lot of `Cannot find module` errors in scenarios

scott-silver:
imports should be updated now

kevincheng96 (approved):

PR #587, opened 2022-09-23
Mainnet WETH Deployment

jflatow:
~~WIP~~ Ready for review now.

Changes made in preparation for launching a ~~mainnet/ETH~~`mainnet/WETH` deployment.

- adds `mainnet-weth` scenario base
- adds configuration for `mainnet/weth` deploy with `wstETH` and `cbETH` as collateral assets
- adds a `WstETHPriceFeed.sol` contract that calculates the `wstETH / USD` price using the `stETH / USD` price feed and the value of `wstETH:stETH`
- updates scenarios to work with ETH-based markets, where possible
- adds a `matchesDeployment` filter to allow for scenarios that specifically target (or avoid) the `mainnet-weth` deploy
- adds a `getRewardToken` fn to context, to allow retrieving the reward token even when it is not a collateral asset

Remaining todos before we can launch:

- [x] set the supply caps back to 0
- [x] replace the `cbETH` price feed with the actual Chainlink `cbETH / USD` price feed (when it's available)
- [x] decide on actual configuration params
- [x] get OZ audit on `WstETHPriceFeed.sol`
- [x] launch a new version of the Liquidator Bot for `mainnet/eth`
- [x] deploy new Bulker (#611)


deployments/mainnet/eth/configuration.json
```diff
@@ -0,0 +1,49 @@
...
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0e15",
+    "baseBorrowSpeed": "0e15",
+    "baseMinForRewards": "1000000e6"
+  },
+  "rewardTokenAddress": "0xc00e94cb662c3520282e6f5717214004a7f26888",
+  "assets": {
```
kevincheng96, line 29: Thoughts on also supporting a stablecoin like USDC to better facilitate the shorting use-case?

kevincheng96: Since with the current configuration, the only use-case is levering up on staked ETH.

.github/workflows/deploy-market.yaml
```diff
@@ -32,7 +33,7 @@ jobs:
       - name: Seacrest
         uses: hayesgm/seacrest@v1
         with:
-          ethereum_url: "${{ fromJSON('{\"fuji\":\"https://api.avax-test.network/ext/bc/C/rpc\",\"kovan\":\"https://kovan.infura.io/v3/$INFURA_KEY\",\"mainnet\":\"https://mainnet.infura.io/v3/$INFURA_KEY\",\"goerli\":\"https://goerli.infura.io/v3/$INFURA_KEY\",\"mumbai\":\"https://polygon-mumbai.infura.io/v3/$INFURA_KEY\"}')[inputs.network] }}"
+          ethereum_url: "${{ fromJSON('{\"fuji\":\"https://api.avax-test.network/ext/bc/C/rpc\",\"kovan\":\"https://kovan.infura.io/v3/$INFURA_KEY\",\"mainnet\":\"https://mainnet.infura.io/v3/$INFURA_KEY\",\"mainnet-eth\":\"https://mainnet.infura.io/v3/$INFURA_KEY\",\"goerli\":\"https://goerli.infura.io/v3/$INFURA_KEY\",\"mumbai\":\"https://polygon-mumbai.infura.io/v3/$INFURA_KEY\"}')[inputs.network] }}"
```
scott-silver, line 36: added the `mainnet-eth` deployment to the Github Actions.

for the `deploy-market` action, `mainnet` uses `https://mainnet.infura.io/v3/$INFURA_KEY`, but for `prepare-migration` and `enact-migration` it uses `https://mainnet-eth.compound.finance\`

wondering if there is a reason for the difference? or should we pick one and use them across the actions

deployments/mainnet/eth/configuration.json
scott-silver: Sounds like we're sticking with `cbETH` and `wstETH` for the launch

.github/workflows/deploy-market.yaml
scott-silver: updated

deployments/mainnet/eth/configuration.json
```diff
@@ -0,0 +1,49 @@
...
+    "cbETH": {
+      "address": "0xBe9895146f7AF43049ca1c1AE358B0541Ea49704",
+      "priceFeed": "0xCfE54B5cD566aB89272946F602D76Ea879CAb4a8",
+      "decimals": "18",
+      "borrowCF": 0.80,
+      "liquidateCF": 0.85,
+      "liquidationFactor": 0.95,
+      "supplyCap": "500e18"
```
scott-silver, line 37: bumped the supply caps to get scenarios running

- [ ] reset supplyCaps to 0

.github/workflows/deploy-market.yaml
```diff
@@ -32,7 +33,7 @@ jobs:
       - name: Seacrest
         uses: hayesgm/seacrest@v1
         with:
-          ethereum_url: "${{ fromJSON('{\"fuji\":\"https://api.avax-test.network/ext/bc/C/rpc\",\"kovan\":\"https://kovan.infura.io/v3/$INFURA_KEY\",\"mainnet\":\"https://mainnet.infura.io/v3/$INFURA_KEY\",\"goerli\":\"https://goerli.infura.io/v3/$INFURA_KEY\",\"mumbai\":\"https://polygon-mumbai.infura.io/v3/$INFURA_KEY\"}')[inputs.network] }}"
```
Cumop21, line 35: jflatow/mainnet-eth

Cumop21 (approved):

scenario/SupplyScenario.ts
```diff
@@ -250,7 +250,7 @@ scenario(
         asset: baseAsset.address,
         amount: 100n * scale,
       })
-    ).to.be.revertedWith('ERC20: transfer amount exceeds allowance');
+    ).to.be.reverted;
```
scott-silver, line 253: `WETH` reverts without reason strings, unfortunately ☹️

scenario/RewardsScenario.ts
```diff
@@ -156,8 +156,8 @@ scenario(
     const baseScale = (await comet.baseScale()).toBigInt();
     const toBorrow = 1_000n * baseScale;
 
-    const [rewardTokenAddress, rescaleFactor] = await rewards.rewardConfig(comet.address);
-    const rewardToken = context.getAssetByAddress(rewardTokenAddress);
+    const [_, rescaleFactor] = await rewards.rewardConfig(comet.address);
+    const rewardToken = await context.getRewardToken();
```
scott-silver, line 160: we could have `context.getRewardToken` return both the `rewardToken` instance and the `rescaleFactor`, which would eliminate the need to call `rewards.rewardConfig` on line 159

jflatow: `getRewardConfig`?

scenario/SupplyScenario.ts
jflatow: This kinda sucks, another case where I think we should at least preserve the error information in a comment in case we find a way to restore strength later

.github/workflows/deploy-market.yaml
```diff
@@ -9,6 +9,7 @@ on:
           - fuji
           - kovan
           - mainnet
+          - mainnet-eth
```
jflatow, line 12: Can we get rid of this now?

deployments/mainnet/eth/deploy.ts
```diff
@@ -0,0 +1,45 @@
+import { Deployed, DeploymentManager } from '../../../plugins/deployment_manager';
+import { DeploySpec, deployComet } from '../../../src/deploy';
+import { getConfiguration, NetworkConfiguration } from '../../../src/deploy/NetworkConfiguration';
+
+export default async function deploy(deploymentManager: DeploymentManager, deploySpec: DeploySpec): Promise<Deployed> {
+  // XXX configuration params for assets are just placeholders
```
jflatow, line 6: We should ask Gauntlet to start thinking about params too

deployments/mainnet/eth/deploy.ts
```diff
@@ -0,0 +1,45 @@
...
+  // Deploy all Comet-related contracts
+  const deployed = await deployComet(
+    deploymentManager,
+    deploySpec,
+    { assetConfigs }
+  );
+  const { comet } = deployed;
+
+  // Get a handle to the WETH contract for bulker
+  const WETH = await deploymentManager.existing('WETH', await comet.baseToken());
```
jflatow, line 35: Is it not in spider yet? (e.g. `deploymentManager.contract('WETH')`)

deployments/mainnet/eth/deploy.ts
```diff
@@ -0,0 +1,45 @@
...
+  let { assetConfigs } = await getConfiguration(deploymentManager);
+  assetConfigs = (assetConfigs || []).map(ac => {
+    return ac.asset === wstETHaddress ? { ...ac, ...{ priceFeed: wstETHPriceFeed.address } } : ac;
+  });
+
+  // Deploy all Comet-related contracts
+  const deployed = await deployComet(
+    deploymentManager,
+    deploySpec,
+    { assetConfigs }
```
jflatow, line 30: Something seems off here, we shouldn't have to do all this... 🤔

jflatow (commented):
Looking great, couple comments/questions 💪

deployments/mainnet/eth/deploy.ts
scott-silver: The additional setup is so that we can deploy the `wstETHPriceFeed` contract and then pass it in as an argument to `deployComet`, overriding the value from `configuration.json`.

We normally don't have to do this since all of our price feeds are already deployed.

One other option would be to deploy the price feed manually in advance, and just add that address to the configuration file.

deployments/mainnet/eth/deploy.ts
scott-silver: Started a conversation in the Gauntlet channel

.github/workflows/deploy-market.yaml
scott-silver: yep, removed

deployments/mainnet/eth/deploy.ts
scott-silver: we don't even need to do that. we can just pass in `await comet.baseToken()` to the Bulker constructor. updated

scenario/SupplyScenario.ts
scott-silver: sure, added the error messages back as comments.

as long as we're running these scenarios against a WETH-based deployment we'll have to stick with `.reverted`, though, since the WETH contract we're using isn't ever going to return revert strings.

jflatow: maybe one day we'll have network specific revert reasons 🤷

deployments/mainnet/eth/deploy.ts
```diff
@@ -0,0 +1,45 @@
+import { Deployed, DeploymentManager } from '../../../plugins/deployment_manager';
+import { DeploySpec, deployComet } from '../../../src/deploy';
+import { getConfiguration, NetworkConfiguration } from '../../../src/deploy/NetworkConfiguration';
+
+export default async function deploy(deploymentManager: DeploymentManager, deploySpec: DeploySpec): Promise<Deployed> {
+  // XXX configuration params for assets are just placeholders
+
+  // Deploy WstETHPriceFeed
+  const wstETHPriceFeed = await deploymentManager.deploy(
+    'wstETHPriceFeed',
```
jflatow, line 8: The alias should be the same as what spider gets, which I think should be `wstETH:priceFeed`

scenario/SupplyScenario.ts
scott-silver: oh, that's an interesting idea.

we could write a wrapper function that the function to execute and a map of networks to the expected error message. for anything

deployments/mainnet/eth/deploy.ts
jflatow: I think we might be able to get rid of everything except the deploy if we use ```getContractAddress(`${assetName}:priceFeed`, assetConfig.priceFeed)``` here https://github.com/compound-finance/comet/blob/main/src/deploy/NetworkConfiguration.ts#L98 i.e. with the current one as the fallback (and I guess here too for parity https://github.com/compound-finance/comet/blob/main/src/deploy/NetworkConfiguration.ts#L134)

scenario/RewardsScenario.ts
scott-silver: i like it. added a `getRewardConfig` function context.

deployments/mainnet/eth/deploy.ts
```diff
@@ -5,28 +5,16 @@ import { getConfiguration, NetworkConfiguration } from '../../../src/deploy/Netw
 export default async function deploy(deploymentManager: DeploymentManager, deploySpec: DeploySpec): Promise<Deployed> {
   // Deploy WstETHPriceFeed
   const wstETHPriceFeed = await deploymentManager.deploy(
-    'wstETHPriceFeed',
+    'wstETH:priceFeed',
```
jflatow, line 8: 🎉

deployments/mainnet/eth/deploy.ts
scott-silver: that's awesome. made the update

deployments/mainnet/eth/configuration.json
```diff
@@ -0,0 +1,49 @@
...
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0e15",
+    "baseBorrowSpeed": "0e15",
+    "baseMinForRewards": "1000000e6"
+  },
+  "rewardTokenAddress": "0xc00e94cb662c3520282e6f5717214004a7f26888",
+  "assets": {
+    "cbETH": {
+      "address": "0xBe9895146f7AF43049ca1c1AE358B0541Ea49704",
+      "priceFeed": "0xCfE54B5cD566aB89272946F602D76Ea879CAb4a8",
```
jflatow, line 32: XXX change to upcoming cbETH/USD feed

jflatow:
For setting the supply caps from 0, we need a separate migration w/ a proposal. For USDC we also seeded reserves, not sure if we want to try to do that here. Doing rewards after I think makes sense

deployments/mainnet/eth/configuration.json
jflatow: Updated to https://etherscan.io/address/0x67eF3CAF8BeB93149F48e8d20920BEC9b4320510

scenario/BulkerScenario.ts
```diff
@@ -168,3 +169,94 @@ scenario(
...
+
+    // Final expectations
+    const baseIndexScale = (await comet.baseIndexScale()).toBigInt();
+    const baseSupplyIndex = (await comet.totalsBasic()).baseSupplyIndex.toBigInt();
+    const baseTransferred = getExpectedBaseBalance(toTransferBase, baseIndexScale, baseSupplyIndex);
+    expect(await comet.collateralBalanceOf(albert.address, collateralAsset.address)).to.be.equal(toSupplyCollateral);
+    expect(await baseAsset.balanceOf(albert.address)).to.be.equal(toBorrowBase);
+    expect(await comet.balanceOf(betty.address)).to.be.equal(baseTransferred);
+    // NOTE: differs from the equivalent scenario for non-ETH markets
+    expect(await comet.borrowBalanceOf(albert.address)).to.be.equal(toBorrowBase + toTransferBase - (toSupplyEth - toWithdrawEth));
```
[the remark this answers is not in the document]
jflatow: Just to make sure I understand, if the scenario changed the collateral asset from ETH to something else it would be the same right?

jflatow:
- [ ] We also need to make the name consistent everywhere (cWETHv3)

scenario/BulkerScenario.ts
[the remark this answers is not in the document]
scott-silver: The `non-WETH` version of this test has these two assertions instead:

```
expect(await comet.borrowBalanceOf(albert.address)).to.be.equal(toBorrowBase + toTransferBase);
expect(await comet.collateralBalanceOf(albert.address, WETH.address)).to.be.equal(toSupplyEth - toWithdrawEth);
```

So in a USDC deployment, the borrow balance is how much the user borrowed and transferred. As well, the user has a collateral balance of WETH equal to how much ETH they supplied minus how much they borrowed.

But in this case, since the base is WETH, then the borrow balance is how much WETH was borrowed, how much WETH was transferred, how much ETH was borrowed (minus how much ETH was supplied)

scenario/GovernanceScenario.ts
```diff
@@ -143,7 +143,68 @@ scenario('add new asset',
...
+    const borrowAmount = 1000n * (await comet.baseScale()).toBigInt();
+    await dogecoin.connect(albert.signer).approve(comet.address, exp(100, 8));
+    await albert.supplyAsset({ asset: dogecoin.address, amount: exp(100, 8) });
+    await albert.withdrawAsset({ asset: baseAssetAddress, amount: borrowAmount });
+
+    expect(await albert.getCometCollateralBalance(dogecoin.address)).to.be.equal(exp(100, 8));
+    expect(await albert.getCometBaseBalance()).to.be.equal(-borrowAmount);
+  });
+
+scenario('add new asset (mainnet-weth)',
```
kevincheng96, line 155: Curious why we need to separate out this scenario? Shouldn't it just work for both deployments?

scenario/RewardsScenario.ts
```diff
@@ -21,7 +21,7 @@ function calculateRewardsOwed(
 scenario(
   'Comet#rewards > can claim supply rewards for self',
   {
-    filter: async (ctx) => await isRewardSupported(ctx),
+    filter: async (ctx) => await isRewardSupported(ctx) && !matchesDeployment(ctx, [{network: 'mainnet', deployment: 'weth'}]),
```
kevincheng96, line 24: We should add an XXX to remove this filter once rewards are supported on the WETH deployment.

test/wsteth-price-feed.ts
```diff
@@ -0,0 +1,107 @@
...
+  await wstETHPriceFeed.deployed();
+
+  return {
+    simpleWstETH,
+    stETHPriceFeed,
+    wstETHPriceFeed
+  };
+}
+
+const testCases = [
```
kevincheng96, line 29: Weird edge case to maybe test is if the price returned is negative (Chainlink returns `int256` for price). I don't think it really matters since Comet will just revert if it's a negative price, but let's make sure we do return a negative price back.

kevincheng96 (commented):
Looks great so far!

scenario/GovernanceScenario.ts
scott-silver: the original test attempts to borrow 1,000 units of the base asset using 100 assets of DOGE (with a price of $1,000 per DOGE).

that works with a USDC base, since a $1,000 borrow can be collateralized with $100,000 worth of DOGE.

but borrowing 1,000 ETH costs about $1.5M, so the borrow is uncollateralized

scenario/RewardsScenario.ts
scott-silver: i guess alternatively, we could delete `rewardTokenAddress` from the `mainnet/WETH` configuration so that the `isRewardSupported` check works as intended

contracts/WstETHPriceFeed.sol
```diff
@@ -0,0 +1,76 @@
...
+    function latestRoundData() override external view returns (
+        uint80 roundId,
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        (uint80 roundId, int256 stETHPrice, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) = AggregatorV3Interface(stETHtoUSDPriceFeed).latestRoundData();
+        uint256 tokensPerStEth = IWstETH(wstETH).tokensPerStEth();
+        int price = stETHPrice * signed256(wstETHScale) / signed256(tokensPerStEth);
```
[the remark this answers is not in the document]
scott-silver: yep, that works too. updated

scenario/RewardsScenario.ts
scott-silver: deleted `rewardTokenAddress` from the configuration file, and deleted this filter

test/wsteth-price-feed.ts
scott-silver: good idea. added a negative price to the test cases

scenario/GovernanceScenario.ts
kevincheng96: Ah I see, that is indeed annoying. Another approach could be to set the borrow amount to be the `borrowBaseMin` instead of a constant. But this works too.

kevincheng96 (commented):
Cool, I think everything looks good now. Just waiting on params to be finalized

contracts/WstETHPriceFeed.sol
```diff
@@ -70,7 +70,7 @@ contract WstETHPriceFeed is AggregatorV3Interface {
     ) {
         (uint80 roundId, int256 stETHPrice, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) = AggregatorV3Interface(stETHtoUSDPriceFeed).latestRoundData();
         uint256 tokensPerStEth = IWstETH(wstETH).tokensPerStEth();
-        int price = stETHPrice * signed256(wstETHScale) / signed256(tokensPerStEth);
+        int256 price = stETHPrice * int256(wstETHScale) / signed256(tokensPerStEth);
```
jflatow, line 73: Since we are assuming `wstETHScale` is already guaranteed to be < int256.max?

scott-silver: yeah, that's the assumption:

https://github.com/compound-finance/comet/pull/587#discussion_r1009744397

scenario/constraints/SupplyCapConstraint.ts
```diff
@@ -0,0 +1,60 @@
...
+    const supplyCaps = requirements.supplyCaps;
+    if (supplyCaps !== undefined) {
+      const solutions = [];
+      solutions.push(async function barelyMeet(context: T) {
+        const supplyAmountPerAsset = {};
+        for (const [assetName, rawAmount] of Object.entries(supplyCaps)) {
+          const asset = await getAssetFromName(assetName, context);
+          const decimals = await asset.token.decimals();
+          const amount = parseAmount(rawAmount);
+          expect(amount.op).to.equal(ComparisonOp.GTE, `Operation ${amount.op} not supported (yet) by supply cap constraint`);
```
kevincheng96, line 19: What does this line do?

contracts/WstETHPriceFeed.sol
jflatow: Yep 👍 comet has `10**18` max constant and changing that would be a massive breaking change

scenario/GovernanceScenario.ts
jflatow: Shouldn't we just have a PriceConstraint?

scott-silver: that's a clever idea. So you'd specify a price for the base asset and/or the collateral assets

we're having a similar conversation on the [Polygon PR](https://github.com/compound-finance/comet/pull/598#discussion_r1013496694).

Another approach would be to write a helper function returned the amount of a given asset required to cover a borrow of X base units

jflatow: Haha, yep just commented there, we can pick one ;) But I think its more desirable to use literal numbers in properties where we can (👨‍🍳 has long advocated for this and I kind of agree)

scott-silver: #615

deployments/mainnet/weth/configuration.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "Compound WETH",
+  "symbol": "cWETHv3",
+  "baseToken": "WETH",
+  "baseTokenAddress": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
+  "baseTokenPriceFeed": "0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419",
+  "borrowMin": "100e6",
```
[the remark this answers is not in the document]
scott-silver: bumping this so we don't forget: `borrowMin`, `targetReserves`, and `baseMinForRewards` are all 6 decimal values, not 18 decimal values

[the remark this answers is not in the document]
kevincheng96: Here's what I'm thinking for the tracking index values:
```
trackingIndexScale: 1e12 or 1e15
baseMinForRewards: 1_000e18 (is this too high? similar to USDC market, this would be around $1mn)
```

kevjin (approved):

PR #590, opened 2022-09-30
Fix coverage

scott-silver:
`--max-old-space-size=4096` fixes the max heap errors:

image

https://github.com/compound-finance/comet/actions/runs/3116058799/jobs/5053854291

bumping the Mocha timeout prevents failing for timeouts:

image

https://github.com/compound-finance/comet/actions/runs/3116323528/jobs/5078730536


jflatow (approved):
💪 Could even just bump it to the max max, but good if its using less

PR #593, opened 2022-10-04
additional strategy for pulling Contract Creation code

scott-silver:
I posted some of this in Discord; adding here for posterity.

Currently, when we want to grab Contract Creation code, we attempt to scrape the content directly from Etherscan.

We make a request for the page data at `https://etherscan.io/address/0x2260fac5e5542a773aa44fbcfedf7c193bc2c599#code`, for example, and search for the contents of a div with a particular ID.

Recently, Polygonscan put a Cloudflare challenge page in front of their pages, so our attempts to scrape their pages fail with a 503 and some page content that includes the text `polygonscan.com needs to review the security of your connection before proceeding`. This breaks our scenarios/deploys for Mumbai/Polygon.

Sometimes, the Contract Creation code can be pulled via the Etherscan API. For example, the "input" data for the first transaction for WBTC here:

https://api.etherscan.io/api?module=account&action=txlist&address=0x2260fac5e5542a773aa44fbcfedf7c193bc2c599&startblock=0&endblock=99999999&page=1&offset=10&sort=asc

...matches the Contract Creation code here: https://etherscan.io/address/0x2260fac5e5542a773aa44fbcfedf7c193bc2c599#code

It's not true for all contracts, unfortunately. DAI, for example, does not have its Contract Creation listed in the `txlist` endpoint:

https://api.etherscan.io/api?module=account&action=txlist&address=0x6b175474e89094c44da98b954eedeac495271d0f&startblock=0&endblock=99999999&page=1&offset=10&sort=asc

This appears to be because DAI was created as part of an "internal" transaction:

https://etherscan.io/address/0x6b175474e89094c44da98b954eedeac495271d0f#internaltx

It's possible to retrieve that internal transaction via the Etherscan API, but the "input" value (which would give the constructor args) is inexplicably empty:

https://api.etherscan.io/api?module=account&action=txlistinternal&txhash=0x495402df7d45fe36329b0bd94487f49baee62026d50f654600f6771bd2a596ab

For the time-being, we can get away with a combination of these two strategies (sometimes scraping, sometimes pulling from transaction data). But if Etherscan puts Cloudflare in front of their other block explorer instances, we may have to resort to an even more elaborate strategy to reproduce Contract Creation code.

### DIFF
--- a/plugins/import/import.ts
+++ b/plugins/import/import.ts
@@ -117,14 +117,15 @@ async function pullFirstTransactionForContract(network: string, address: string)
     apikey: getEtherscanApiKey(network)
   };
   const url = `${getEtherscanApiUrl(network)}?${paramString(params)}`;
+  const debugUrl = `${getEtherscanApiUrl(network)}?${paramString({ ...params, ...{ apikey: '[API_KEY]'}})}`;
 
-  debug(`Attempting to pull Contract Creation code from first tx at ${url}`);
+  debug(`Attempting to pull Contract Creation code from first tx at ${debugUrl}`);
   const result = await get(url, {});
   const contractCreationCode = result.result[0].input;
   if (!contractCreationCode) {
-    throw new Error(`Unable to find Contract Creation tx at ${url}`);
+    throw new Error(`Unable to find Contract Creation tx at ${debugUrl}`);
   }
-  debug(`Creation Code found in first tx at ${url}`);
+  debug(`Creation Code found in first tx at ${debugUrl}`);
   return contractCreationCode.slice(2);
 }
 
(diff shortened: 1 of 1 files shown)


plugins/import/import.ts
```diff
@@ -73,21 +83,67 @@ async function getEtherscanApiData(network: string, address: string, apiKey: str
...
+    startblock: 0,
+    endblock: 99999999,
+    page: 1,
+    offset: 10,
+    sort: 'asc',
+    apikey: getEtherscanApiKey(network)
+  };
+  const url = `${getEtherscanApiUrl(network)}?${paramString(params)}`;
+
+  debug(`Attempting to pull Contract Creation code from first tx at ${url}`);
```
scott-silver, line 121: this dumps our various API keys into the logs ☹️ . i should probably remove that param before logging

scott-silver: removed

PR #594, opened 2022-10-05
Attempt to patch around custom errors in hardhat

jflatow:
Uses a new cut of hardhat with selectors included in error messages.

~~XXX: currently 3 new test failures due to "incorrect parameters"...~~
Added a second patch to hardhat to fix inferring errors...


kevincheng96 (approved):
Awesome!

PR #595, opened 2022-10-05
Deploy Mumbai

scott-silver:
It turns out that when a Github Action adds a commit to a branch, it doesn't trigger the same CI actions as when you push changes. Long thread about it [here](https://github.com/orgs/community/discussions/25702).

So I had to add a vacuous change to trigger CI checks after the new roots were added.

That seems like an important issue, because it means scenarios aren't run against newly-deployed roots.


jflatow:
> It turns out that when a Github Action adds a commit to a branch, it doesn't trigger the same CI actions as when you push changes. Long thread about it [here](https://github.com/orgs/community/discussions/25702).
> 
> So I had to add a vacuous change to trigger CI checks after the new roots were added.
> 
> That seems like an important issue, because it means scenarios aren't run against newly-deployed roots.

I had actually thought of this as almost a feature since we had already run deploy in scenarios and the workflow must/should have been run on a branch passing CI, but I agree it could be good to run again with real roots before merge

README.md
```diff
@@ -253,4 +253,4 @@ contract (or you can deploy a new instance of it yourself):
 Initiating transactions this way via the public mempool will
 [almost certainly get frontrun](https://youtu.be/UZ-NNd6yjFM), but you might be
 able to use [flashbots](https://docs.flashbots.net/) to mask your transactions
-from frontrunners.
\ No newline at end of file
+from frontrunners.
```
kevincheng96, line 256: nit: should we revert this change?

kevincheng96 (approved):
Awesome, can't wait to try it out!

README.md
scott-silver: sure thing. reverted

scott-silver:
We can merge this set of roots, but we have several pending patches to Comet that will not be included (#596 and the incoming sweep functionality). I have an in-progress change that will make some small changes to the Bulker as well.

We can either merge these roots and then run a migration to upgrade once those patches are merged. Or we can just wait until those patches are merged and then redeploy.

jflatow:
> We can merge this set of roots, but we have several pending patches to Comet that will not be included (#596 and the incoming sweep functionality). I have an in-progress change that will make some small changes to the Bulker as well.
> 
> We can either merge these roots and then run a migration to upgrade once those patches are merged. Or we can just wait until those patches are merged and then redeploy.

Let's wait until all the patches are all ready, one of the goals is to have the same version of Comet everywhere, and its nice to avoid the unnecessary migration

scott-silver:
deleted the old roots, rebased and added new roots: https://github.com/compound-finance/comet/actions/runs/3373881750/jobs/5598995780

PR #596, opened 2022-10-06
Proposal to patch in the absorb Transfer event

jflatow:


jflatow:
This will subsume #572 as well

jflatow:
Part of #599

PR #597, opened 2022-10-07
[draft] attempt to get unit tests running again

scott-silver:


hardhat.config.ts
```diff
@@ -232,7 +232,7 @@ const config: HardhatUserConfig = {
   },
 
   paths: {
-    tests: './{test,plugins/deployment_manager/test}',
+    tests: '.test',
```
jflatow, line 235: Whoa, big breaking change!

jflatow:
> This will get tests running locally and in CI, but has the unfortunate side effect that running `yarn test` and `yarn hardhat test` will run two different sets of tests ☹️

Whoa, know why? Isn't `yarn test` just calling `hardhat test`?

> 
> The tests run but fail with:
> 
> ```
> /home/runner/work/comet/comet/node_modules/hardhat/internal/hardhat-network/stack-traces/solidity-errors.js:112
>     return new SolidityCallSite(sourceReference.sourceName, sourceReference.contract, sourceReference.function !== undefined
> ```
> 
> 🍯 This is the error that your Hardhat patch was intended to fix, correct?
> 
> Does the latest release of your forked Hardhat include the patch?

Yes, at least an error *like* that. We should give https://github.com/NomicFoundation/hardhat/pull/3205 a try...

hardhat.config.ts
scott-silver: you mean it's a breaking change that `'./{test,plugins/deployment_manager/test}'` stopped working with Hardhat?

kevincheng96:
> > This will get tests running locally and in CI, but has the unfortunate side effect that running `yarn test` and `yarn hardhat test` will run two different sets of tests ☹️
> 
> Whoa, know why? Isn't `yarn test` just calling `hardhat test`?

I'm guessing because `yarn test` now runs `hardhat test ./test/*.ts ./test/**/*.ts ./plugins/deployment_manager/test/*.ts`, while `yarn hardhat test` runs `hardhat test` on tests path is specified in `hardhat.config.ts` (which is just `.test` now).

hardhat.config.ts
kevincheng96, line 235: Should this actually be `./test`?

scott-silver: yep 🤦

jflatow: yes

test/rewards-test.ts
```diff
@@ -153,7 +153,8 @@ describe('CometRewards', () => {
       const { rewards } = await makeRewards({ governor, configs: [[comet, COMP]] });
       await expect(
         rewards.setRewardConfig(comet.address, COMP.address)
-      ).to.be.revertedWith(`custom error 'AlreadyConfigured("${comet.address}")`);
+        //).to.be.revertedWith(`custom error 'AlreadyConfigured("${comet.address}")`);
```
scott-silver, line 156: are these commented-out lines left in intentionally?

jflatow: Yeah, since they were stronger assertions before, I left them in case we switch to `revertWithCustomError` or something which can add back those checks

scott-silver:
> We also move the Comet errors into the interface, in order to make things simpler/more consistent for the future, even though its not totally necessary right now.

Does this mean we'll need a governance proposal to merge these changes?

plugins/scenario/utils/hreForBase.ts
```diff
@@ -36,20 +27,7 @@ waffle type extension: https://github.com/nomiclabs/hardhat/blob/master/packages
...
-        args?: any[],
-        overrideOptions?: providers.TransactionRequest
-      ) => Promise<Contract>;
-      solidity: typeof solidity;
-      link: typeof link;
-      deployMockContract: (signer: Signer, abi: any[]) => Promise<MockContract>;
-      createFixtureLoader: typeof createFixtureLoader;
-      loadFixture: typeof loadFixture;
-    };
+    waffle: any;
```
scott-silver, line 30: I suspect we can delete this `waffle` key entirely, since we're no longer attaching waffle matchers to the environment.

scott-silver:
@jflatow I don't think I can approve this since I opened the original PR, but LGTM!

jflatow (approved):

plugins/scenario/utils/hreForBase.ts
jflatow: Maybe informative to leave it though, I'm not going to mess with it

PR #598, opened 2022-10-11
Polygon cUSDCv3 deploy

scott-silver:
Deploy script and updates for a potential Polygon cUSDCv3 market.

- Renames IWETH9.sol to IWrappedNativeToken.sol (since we will use it for WETH, WMATIC and potentially WAVAX)
- Updates the WETH references in Bulker.sol to WrappedNative
- Adds `fxRoot` to the `mainnet/USDC` roots, along with relations to find the `stateSender`
- Polygon cUSDCv3 config
- Polygon cUSDCv3 deploy script
- updates executeOpenProposalAndRelay to only attempt to relay for proposals that target a bridged deployment (otherwise, the `relayMessage` will timeout waiting for a message that never comes)

TODOs

- [x] rename IWrappedNativeToken.sol to IWrappedNativeAsset.sol
- [x] lint
- [x] set `supplyCap`s to 0
- [x] decide on assets
- [x] ~~decide on borrow factors, collateral factors, liquidation factors~~ (set in proposal)
- [x] ~~update `mainnet/USDC` roots~~ [edited roots.json manually]
- [x] update `Polygon/USDC` roots


deployments/polygon/usdc/configuration.json
```diff
@@ -0,0 +1,64 @@
...
+  },
+  "assets": {
+    "WETH": {
+      "address": "0x7ceb23fd6bc0add59e62ac25578270cff1b9f619",
+      "priceFeed": "0xF9680D99D6C9589e2a93a78A04A279e509205945",
+      "decimals": "18",
+      "borrowCF": 0.825,
+      "liquidateCF": 0.895,
+      "liquidationFactor": 0.95,
+      "supplyCap": "500000e18"
```
scott-silver, line 34: added some placeholder `supplyCap`s for scenarios, but we should set these to 0 before merging

deployments/polygon/usdc/configuration.json
```diff
@@ -0,0 +1,64 @@
...
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0e15",
+    "baseBorrowSpeed": "0e15",
+    "baseMinForRewards": "1000000e6"
+  },
+  "assets": {
```
scott-silver, line 26: tossed in some assets (with some arbitrary `borrowCF`, `liquidateCF` and `liquidationFactor` values); additions/removals/adjustments welcome

deployments/polygon/usdc/configuration.json
```diff
@@ -0,0 +1,64 @@
...
+    "WETH": {
+      "address": "0x7ceb23fd6bc0add59e62ac25578270cff1b9f619",
+      "priceFeed": "0xF9680D99D6C9589e2a93a78A04A279e509205945",
```
scott-silver, line 29: this is the `ETH / USD` price feed for Polygon

deployments/polygon/usdc/configuration.json
```diff
@@ -0,0 +1,64 @@
...
+      "decimals": "18",
+      "borrowCF": 0.825,
+      "liquidateCF": 0.895,
+      "liquidationFactor": 0.95,
+      "supplyCap": "500000e18"
+    },
+    "WBTC": {
+      "address": "0x1bfd67037b42cf73acf2047067bd4f2c47d9bfd6",
+      "priceFeed": "0xDE31F8bFBD8c84b5360CFACCa3539B938dd78ae6",
```
scott-silver, line 38: this is the `WBTC / USD` price feed. it is a "Custom" feed, which Chainlink describes as "Feeds built to serve a specific use case and might not be suitable for general use"

there is also a `BTC / USD` feed

deployments/polygon/usdc/configuration.json
```diff
@@ -0,0 +1,73 @@
...
+      "decimals": "8",
+      "borrowCF": 0.70,
+      "liquidateCF": 0.77,
+      "liquidationFactor": 0.95,
+      "supplyCap": "500000e18"
+    },
+    "WMATIC": {
```
scott-silver, line 45: so the plan to launch on Polygon without `WMATIC` as a collateral asset hit a snag: having a functioning Bulker requires passing in a `wrappedNativeAsset` that is a collateral asset in the protocol.

so if you create an instance of Bulker with `WMATIC` as the `wrappedNativeAsset`, but `WMATIC` is not a collateral asset, then attempts to use the `ACTION_SUPPLY_NATIVE_ASSET` and `ACTION_WITHDRAW_NATIVE_ASSET` actions will fail.

our options seem to be:

- launch on Polygon without Bulker; update the UI to work without Bulker
- launch with a Bulker where the `ACTION_SUPPLY_NATIVE_ASSET` and `ACTION_WITHDRAW_NATIVE_ASSET` will revert; update scenarios to exclude the Polygon deployment from scenarios that test that functionality
- launch with WMATIC; add cWMATICv3 later (ending up with both as collateral assets)
- launch cWMATICv3, followed by cUSDCv3 (with cWMATICv3 as collateral asset)

contracts/Bulker.sol
```diff
@@ -14,28 +14,28 @@ interface IClaimable {
 contract Bulker {
     /** General configuration constants **/
     address public immutable admin;
-    address payable public immutable weth;
+    address payable public immutable wrappedNativeAsset;
```
scott-silver, line 17: Changing this value from `weth` to `wrappedNativeAsset` is purely a cosmetic change (the variable might actually refer to a different wrapped asset, like wrapped Matic or wrapped Avax).

It's a nice update, but causes issues when trying to read this value for deployments that use the existing Bulker (trying to read `bulker.wrappedNativeAsset()` from those contracts reverts, since their equivalent is `bulker.weth()`).

Some options to deal with this:

1. deploy new Bulkers for `mainnet` and `goerli`
2. revert these updates to Bulker; leave the references to `weth` as is
3. create a `context.getNativeAssetWrapper` function that returns the value at this storage slot, wether it's called `weth` or `wrappedNativeAsset`

re-deploying seems heavy-handed; 2 or 3 seem preferable

deployments/polygon/usdc/configuration.json
scott-silver: discussed during Parking lot.

it sounds like we're comfortable deploying with a Bulker that has a few actions that revert. when we want to have a Polygon Bulker that is capable of wrapping MATIC via a cWMATICv3 deployment, we'll need to create a new Bulker that takes in another Comet instance as a wrapper asset (instead of a WETH-style wrapper asset).

contracts/Bulker.sol
scott-silver: also resolved in parking lot

deployments/polygon/usdc/configuration.json
scott-silver: agreed. will update

scott-silver: updated

scenario/utils/index.ts
```diff
@@ -257,6 +258,10 @@ export async function isBulkerSupported(ctx: CometContext): Promise<boolean> {
   return bulker == null ? false : true;
 }
 
+export function isInNetwork(ctx: CometContext, networks: string[]): boolean {
```
scott-silver, line 261: we should replace this with the more powerful `matchesDeployment`fn from #587: 

https://github.com/compound-finance/comet/pull/587/files#diff-33630834798399f8c06541165f332a3c60939d28f72c8a664a4e4e116737d02aR265

jflatow: Really wish we could get rid of `matchesDeployment` though and find a way to do things through constraints

deployments/polygon/usdc/configuration.json
jflatow: I guess next steps are to rebase this on supply cap constraint, set these to 0, and take out of 'draft' status?

scenario/utils/index.ts
scott-silver: most often, we need `matchesDeployment`because there's a scenario that assumes a particular relationship between the base asset and a collateral asset.

for example, if you have a scenarios that 1) supplies 1,000 of `$asset0` and then 2) borrows 100 of the base asset.

that works for many of our deployments, because the base asset is a stablecoin and the first asset is worth more than $1/unit.

but you need way more collateral than that when the base asset is WETH.

scott-silver: we could add a helper function that takes in a comet instance and a number of base units and it returns you the amount of the first collateral asset that is needed in order to cover a borrow of that many base units

jflatow: Then we're getting away from using real numbers in scens... I'm starting to think a PriceConstraint, though relatively complicated, might solve a bunch of these problems more elegantly

scott-silver: #615

deployments/polygon/usdc/deploy.ts
```diff
@@ -0,0 +1,74 @@
...
+  const deployed = await deployComet(deploymentManager, deploySpec);
+  const { comet } = deployed;
+
+  // Deploy Bulker
+  const bulker = await deploymentManager.deploy(
+    'bulker',
+    'Bulker.sol',
+    [
+      await comet.governor(),      // admin
+      ethers.constants.AddressZero // weth (zero address, since Polygon deployment does not include WMATIC)
```
jflatow, line 64: We should revisit this?

kevincheng96: That field has changed to `wrappedNativeToken` in `BaseBulker`. So we can pass in the address of `WMATIC` here.

Also need to update the file from `Bulker.sol` to either `BaseBulker.sol` or `PolygonBulker.sol` (which does not exist yet).

deployments/polygon/usdc/configuration.json
scott-silver: updated supply caps

scenario/BulkerScenario.ts
```diff
@@ -10,15 +10,16 @@ scenario(
     filter: async (ctx) => await isBulkerSupported(ctx) && !matchesDeployment(ctx, [{deployment: 'weth'}, {network: 'mumbai'}]),
     supplyCaps: {
       $asset0: 3000,
+      $asset2: 1,
```
scott-silver, line 13: commented on this in #668, but wanted to mention it here; 

This scenario assumes that the wrapped native token collateral asset has a supply cap of at least .01 units.

In the Polygon deploy, that asset is $asset2

PR #599, opened 2022-10-13
Patches and proposal for mainnet USDC market upgrade

jflatow:
This branch includes 2 recommended patches to the market logic, and creates a proposal to apply the updates to mainnet cUSDCv3.

*Emit erc20 transfer (mint) event when absorbed account ends up with a positive balance*
 - Calculate amount by converting from principal instead of directly using present value
 - Move Transfer event down
 - Fix unit test

*Automatically include excess balance in collateral reserves*

- Adds a reserves fn which it checks when collateral is bought

Note: we could add a fn to convert any (prior) collateral balance of the contract into reserves by clearing it out and removing from total. However, collateral reserves are currently 0, and we expect them to remain 0 if liquidation is working properly. If somehow collateral is acquired and held by the protocol (i.e. from an absorb) before this patch gets applied, the protocol will simply hold that balance as a supplier, until governance decides to move it, for which a fn can always be added (at that time) to assist.

Also note that we expect 1 scenario to fail on each of `mainnet`, `goerli`, and `fuji`, for one of two runs, due to testing for the newly patched in event. We also expect the new scenario testing that the entire ERC20 collateral balance is part of reserves fails in the run without the migration.


jflatow:
TODO:
 - [ ] Scenario to transfer collateral to the market without supplying, and buy it back

scott-silver:
Does this subsume #572 ?

jflatow:
> Does this subsume #572 ?

Yes, was holding off to close that one until we merge this one, as its easier to review just the event change from that PR

deployments/fuji/usdc/migrations/1665028496_absorb_transfer_event.ts
```diff
@@ -0,0 +1,62 @@
+import { DeploymentManager, migration } from '../../../../plugins/deployment_manager';
+import { calldata, exp, proposal } from '../../../../src/deploy';
```
scott-silver, line 2: nit: looks like `calldata`, `exp`, `expect` are all unused. same with the `COMP` on line 21

scott-silver: I guess we don't lint the `/deployments/` directory. I'll create a PR to enable that

deployments/fuji/usdc/migrations/1665028496_absorb_transfer_event.ts
```diff
@@ -0,0 +1,62 @@
...
+    const {
+      governor,
+      comet,
+      configurator,
+      rewards,
+      COMP,
+    } = await deploymentManager.getContracts();
+
+    // 1. & 2.
+    //  added a scenario to check for new Transfer event
```
scott-silver, line 60: is this a todo?

deployments/mainnet/usdc/migrations/1665028496_absorb_transfer_event.ts
```diff
@@ -0,0 +1,75 @@
...
+    );
+    trace(txn);
+
+    const event = (await txn.wait()).events.find(event => event.event === 'ProposalCreated');
+    const [proposalId] = event.args;
+    trace(`Created proposal ${proposalId}.`);
+  },
+
+  async verify(deploymentManager: DeploymentManager) {
+    const {
```
scott-silver, line 61: nit: everything excep `COMP` is unused

deployments/mainnet/usdc/migrations/1665028496_absorb_transfer_event.ts
```diff
@@ -0,0 +1,75 @@
...
+      },
+
+      // 2. Deploy and upgrade to a new version of Comet
+      {
+        contract: cometAdmin,
+        signature: 'deployAndUpgradeTo(address,address)',
+        args: [configurator.address, comet.address],
+      },
+
+      // 3. Transfer COMP to Andrew
```
scott-silver, line 42: 🙌 that's awesome

scenario/LiquidationScenario.ts
```diff
@@ -161,6 +161,37 @@ scenario(
...
+
+    await world.increaseTime(
+      await timeUntilUnderwater({
+        comet,
+        actor: albert,
+      })
+    );
+
+    const ab0 = await betty.absorb({ absorber: betty.address, accounts: [albert.address] });
+    expect(ab0.events[2].event).to.be.equal('Transfer');
```
scott-silver, line 188: looks like `ab0.events` can be null.

you could update this to: `ab0.events?.[2]?.event` if you wanted the expectations to fall when there were no events (instead of failing with a JS error about not finding `event` on `undefined`)

test/absorb-test.ts
```diff
@@ -355,14 +356,15 @@ describe('absorb', function () {
     const { comet, tokens, users: [absorber, underwater], priceFeeds } = protocol;
     const { COMP, WBTC, WETH } = tokens;
 
-    const debt = 1n - (exp(41000, 6) + exp(3000, 6) + exp(175, 6));
+    const finalDebt = 1n;
+    const startingDebt = finalDebt - (exp(41000, 6) + exp(3000, 6) + exp(175, 6));
```
scott-silver, line 360: what are the 41K, 3K and 175 values here?

deployments/fuji/usdc/migrations/1665028496_absorb_transfer_event.ts
jflatow: No its there

scenario/LiquidationScenario.ts
jflatow: This is the test we want ;) See the note at the top

test/absorb-test.ts
jflatow: These are from Kevin's branch, he updated this test to be more logical while he was at it - I believe those are coming from the prices?

deployments/fuji/usdc/migrations/1665028496_absorb_transfer_event.ts
scott-silver: is there a reason to call `getContracts()` then?

deployments/fuji/usdc/migrations/1665028496_absorb_transfer_event.ts
scott-silver: #601

scenario/LiquidationScenario.ts
jflatow: Good call 🦅

deployments/fuji/usdc/migrations/1665028496_absorb_transfer_event.ts
jflatow: I think I was going to check that the address had changed, but I removed this and updated the scen name / description

contracts/Comet.sol
```diff
@@ -1190,10 +1202,11 @@ contract Comet is CometMainInterface {
 
         uint collateralAmount = quoteCollateral(asset, baseAmount);
         if (collateralAmount < minAmount) revert TooMuchSlippage();
+        if (collateralAmount > getCollateralReserves(asset)) revert InsufficientReserves();
 
         // Note: Pre-transfer hook can re-enter buyCollateral with a stale collateral ERC20 balance.
         //       This is a problem if quoteCollateral derives its discount from the collateral ERC20 balance.
-        withdrawCollateral(address(this), recipient, asset, safe128(collateralAmount));
+        doTransferOut(asset, recipient, safe128(collateralAmount));
```
kevincheng96, line 1209: I think this change actually makes the protocol even more susceptible to a pre-transfer hook attack. A malicious token can bypass the `if (collateralAmount > getCollateralReserves(asset)) revert InsufficientReserves();` to buy more than the collateral reserves with re-entrancy.

This is because the accounting check now happens separately and before the `doTransferOut`, while the old code did the accounting check and `doTransferOut` together in `withdrawCollateral`.

jflatow: Malicious token can buy more of itself than reserves? Interesting, we should think this through

kevincheng96: Discussed offline. Thankfully, only the `msg.sender` (`Comet`) can register a pre-transfer hook for a ERC777 token. BUT, this is still vulnerable to a malicious token, so we should probably update the documentation.

jflatow: We think this isn't susceptible issues from ERC777 since only sender can register pre-transfer hooks in that standard. For some questionable asset in the future which allows arbitrary pre-transfer behavior, it should not be listed, but at least the risk is isolated to the questionable asset

contracts/Comet.sol
```diff
@@ -477,7 +477,15 @@ contract Comet is CometMainInterface {
     }
 
     /**
-     * @notice Gets the total amount of protocol reserves, denominated in the number of base tokens
+     * @notice Gets the total balance of protocol collateral reserves for an asset
```
kevincheng96, line 480: Worth documenting that we're intending for this to revert if `balanceOf(this) < totalSupplyAsset`?

jflatow: Yep

scenario/LiquidationScenario.ts
```diff
@@ -161,6 +161,37 @@ scenario(
...
+
+    await world.increaseTime(
+      await timeUntilUnderwater({
+        comet,
+        actor: albert,
+      })
+    );
+
+    const ab0 = await betty.absorb({ absorber: betty.address, accounts: [albert.address] });
+    expect(ab0.events?.[2]?.event).to.be.equal('Transfer');
```
kevincheng96, line 188: Cool, lets also check that this is a `mint` event as opposed to any random ERC20 transfer event:

```
expect(event(ab0, 2)).to.be.deep.equal({
      Transfer: {
        amount: baseBalance,
        from: ethers.constants.AddressZero,
        to: albert.address,
      }
    });
```

test/buy-collateral-test.ts
```diff
@@ -321,12 +393,12 @@ describe('buyCollateral', function () {
 
       // set collateral balance
       const t0 = Object.assign({}, await normalComet.totalsCollateral(normalWETH.address), {
-        totalSupplyAsset: exp(100, 18),
+        totalSupplyAsset: exp(5, 18),
```
kevincheng96, line 396: Curious what these changes are for?

kevincheng96:
Overall, this looks good. Only suggestions are to update the documentation in two places:
1. `getCollateralReserves` can revert under some cases, which is non-critical
2. A malicious token with a pre-transfer hook can buy users' collateral

test/buy-collateral-test.ts
jflatow: These are necessary bc you set up collateral reserves now differently than before. This is the amount of collateral that belongs to users excluding reserves now

kevincheng96: In that case, shouldn't these (`totalSupplyAsset`) stay at 100e18 while lines 421-430 change from `setCollateralBalance of Comet for WETH to 5e18` to `subtract 5e18 from totalSupplyAsset for WETH`?

jflatow: Hm, that's a good point, I guess the `setCollateralBalance` is supposed to be simulating absorb here, not a `supplyCollateral`... but I guess I don't really understand this part of the test. I can just delete the setting of totals collateral and collateral balance for the market, it doesn't change any of whats being tested, afaik

scenario/LiquidationScenario.ts
jflatow: This doesn't handle the pre-upgrade failure cases as gracefully, with an expectation failure like we made the other change to support, I think the presence of Transfer event should actually be sufficient, although could add a check for each of the properties if you really think its important?

kevincheng96 (approved):
👍

PR #600, opened 2022-10-14
WstETHPriceFeed + tests

scott-silver:
calculated price feed for `wstETH / USD`, using the `stETH / USD` price feed and the `wstETH`'s `tokensPerStEth` function.

Implements the Chainlink `AggregatorV3Interface`, though it only implements the `latestRoundData` function that Comet uses.


contracts/WSTETHPriceFeed.sol
```diff
@@ -0,0 +1,73 @@
...
+     * @notice Unimplemented function required to fulfill AggregatorV3Interface; always reverts
+     **/
+    function getRoundData(uint80 _roundId) override external view returns (
+        uint80 roundId,
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        revert NotImplemented();
```
scott-silver, line 49: `getRoundData` is required for the AggregatorV3Interface, but it seems potentially harmful to return a value for this function since we don't actually know what the `tokensPerStEth` value was at the time that of a given `stETH / USD` price

contracts/WSTETHPriceFeed.sol
```diff
@@ -0,0 +1,73 @@
...
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        (,int256 stETHPrice, , , ) = AggregatorV3Interface(stETHtoUSDPriceFeed).latestRoundData();
+        uint8 wstDecimals = IWstETH(wstETH).decimals();
+        uint256 tokensPerStEth = IWstETH(wstETH).tokensPerStEth();
+        int price = stETHPrice * signed256(tokensPerStEth) / signed256(10 ** wstDecimals);
+        return (0, price, 0, 0, 0);
```
scott-silver, line 71: I've opted to return 0 for `roundId`, `startedAt`, `updatedAt` and `answeredInRound`

Alternatively, we could pass through the values for the `stETH / USD` latest round data

test/wsteth-price-feed.ts
```diff
@@ -0,0 +1,77 @@
...
+  return {
+    simpleWstETH,
+    stETHpriceFeed,
+    wstETHPriceFeed
+  };
+}
+
+const testCases = [
+  {
+    stEthPrice: exp(1000, 8),
```
scott-silver, line 33: just added some easily verifiable price test cases, e.g.:

if the price of `stETH / USD` is $1000, and one `wstETH` is worth .2 `stETH`, then the `wstETH / USD` price should be $200.

suggestions for additional test cases welcome. might be worth doing some much larger and small numbers, though at some point we're just testing that Solidity's multiplication works as expected.

contracts/WSTETHPriceFeed.sol
kevincheng96: I think it's better to pass through the values in case we ever decide to check for staleness in the future.

contracts/WSTETHPriceFeed.sol
```diff
@@ -0,0 +1,73 @@
...
+        uint80 roundId,
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        (,int256 stETHPrice, , , ) = AggregatorV3Interface(stETHtoUSDPriceFeed).latestRoundData();
+        uint8 wstDecimals = IWstETH(wstETH).decimals();
+        uint256 tokensPerStEth = IWstETH(wstETH).tokensPerStEth();
+        int price = stETHPrice * signed256(tokensPerStEth) / signed256(10 ** wstDecimals);
```
kevincheng96, line 70: I think this should be `stETHPrice / tokensPerStEth` instead.

test/wsteth-price-feed.ts
kevincheng96: It's actually the other way around. If `tokensPerStEth=0.2`, it means that one `wstETH` is worth `5 stETH`. So the price of `wstETH` should be $5000.

scott-silver: The [description](https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#readContract) of the `tokensPerEth` function is:

`Get amount of wstETH for one stETH`

The value right now is `916820579837578359`, so `916820579837578359` / 1e18 = 0.9168205798375784.

My reading of that is that it means that one `wstETH` is currently worth ~.91 `stETH1`

scott-silver: There's a different function, `stEthPerToken` which I believe would return the value you're describing: `Get amount of stETH for one wstETH`

contracts/WSTETHPriceFeed.sol
scott-silver: see https://github.com/compound-finance/comet/pull/600#discussion_r997477595

test/wsteth-price-feed.ts
scott-silver: updated

contracts/WSTETHPriceFeed.sol
scott-silver: updated

contracts/WSTETHPriceFeed.sol
```diff
@@ -0,0 +1,73 @@
...
+     **/
+    function latestRoundData() override external view returns (
+        uint80 roundId,
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        (,int256 stETHPrice, , , ) = AggregatorV3Interface(stETHtoUSDPriceFeed).latestRoundData();
+        uint8 wstDecimals = IWstETH(wstETH).decimals();
```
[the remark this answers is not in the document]
scott-silver: updated

contracts/WstETHPriceFeed.sol
```diff
@@ -0,0 +1,76 @@
...
+    function latestRoundData() override external view returns (
+        uint80 roundId,
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        (uint80 roundId, int256 stETHPrice, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) = AggregatorV3Interface(stETHtoUSDPriceFeed).latestRoundData();
+        uint256 tokensPerStEth = IWstETH(wstETH).tokensPerStEth();
+        int price = stETHPrice * signed256(wstETHScale) / signed256(tokensPerStEth);
```
scott-silver, line 73: ⛑️ Updated the logic of this line; I believe this is correct, but let me know if you agree.

The price is now:

```
stETHPrice * wstETHScale / tokensPerEth
```

e.g.:

```
1300e8 * 1e18 / .91e18 == ~1428e8
```

test/wsteth-price-feed.ts
```diff
@@ -0,0 +1,80 @@
...
+
+describe('wstETH price feed', function () {
+  describe('latestRoundData', function () {
+    for (const { stEthPrice, tokensPerStEth, result } of testCases) {
+      it(`stEthPrice (${stEthPrice}), tokensPerStEth (${tokensPerStEth}) -> ${result}`, async () => {
+        const { wstETHPriceFeed } = await makeWstETH({ stEthPrice, tokensPerStEth });
+        const latestRoundData = await wstETHPriceFeed.latestRoundData();
+        const price = latestRoundData.answer.toBigInt();
+
+        expect(price).to.eq(result);
```
kevincheng96, line 65: Should we also validate timestamp, roundId, etc?

kevincheng96 (approved):
LGTM!

Cumop21 (approved):

contracts/WstETHPriceFeed.sol
```diff
@@ -0,0 +1,76 @@
...
+    /// @notice Chainlink stETH / USD price feed
+    address public immutable stETHtoUSDPriceFeed;
+
+    /// @notice WstETH contract address
+    address public immutable wstETH;
+
+    /// @notice scale for WstETH contract
+    uint public immutable wstETHScale;
+
+    constructor(address stETHtoUSDPriceFeed_, address wstETH_) {
```
jflatow, line 32: So basically we don't need the wstETH/stETH price feed? Interesting that it exists

scott-silver: Is there one for mainnet?

I only see `wstETH / stETH` for [Optimism](https://data.chain.link/optimism/mainnet/crypto-eth/wsteth-steth%20exchangerate) and [Arbitrum](https://data.chain.link/arbitrum/mainnet/crypto-eth/wsteth-steth%20exchangerate).

Looks like they're returning the same value as [wstETH.stEthPerToken](https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#readContract)

test/wsteth-price-feed.ts
scott-silver: sure, added a test for the additional return values

Cumop21 (approved):

contracts/WstETHPriceFeed.sol
jflatow: You're right I don't see Ethereum mainnet

test/absorb-test.ts
```diff
@@ -64,7 +64,7 @@ describe('absorb', function () {
     expect(lU1.numAbsorbed).to.be.equal(0);
     expect(lU1.approxSpend).to.be.equal(0);
 
-    const usdcPrice = await priceFeeds['USDC'].price();
+    const [_, usdcPrice] = await priceFeeds['USDC'].latestRoundData();
```
jflatow, line 67: So we were accessing the stored price directly before? I guess `latestRoundData` is actually more correct since that's what we use in Comet and the official interface?

scott-silver: yeah, i had made `price` a public variable, which was a mistake.

all the variables are now internal and you have to access them via `latestRoundData` or `getRoundData`, which are the functions defined by the `AggregatorV3Interface`

PR #605, opened 2022-10-25
Add a script to help testnet gov and improve interface for scripts

jflatow:
May need some tweaks (still running now, fuji voting period is loong ⏳ )


jflatow:
Yay it works, finished enacting `fuji`

PR #607, opened 2022-10-26
Add a publish task

jflatow:
This is useful for verifying contracts deployed e.g. from `prepare` that didn't get verified (like the CometFactory on a Comet upgrade), although we should probably change the prepare task to change the verification strategy, its still useful to have this option


hayesgm (approved):
really nice to have

PR #611, opened 2022-11-01
Mainnet WETH Bulker

scott-silver:
Creating a new Bulker that is able to wrap/unwrap `stETH`.

Includes the following changes:

- moves the Bulker into a `bulkers/` subdirectory, renaming it to `BaseBulker`
- creates a new `MainnetBulker` that inherits from `BaseBulker` and adds `supplyStEthTo` and `withdrawStEthTo` functions
- converts the bulker action constants to strings/bytes32 to decrease the likelihood that bulkers will create actions with conflicting values
- adds a `handleAction` function that contracts inheriting from `BaseBulker` can implement; by default it just reverts
- scenarios for wrapping and unwrapping via the `MainnetBulker`

TODO:
- [x] fix unit tests
- [x] add scenario for BaseBulker that checks the UnhandledAction() error

### DIFF
--- a/contracts/bulkers/BaseBulker.sol
+++ b/contracts/bulkers/BaseBulker.sol
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.15;
 
-import "./CometInterface.sol";
-import "./ERC20.sol";
-import "./IWETH9.sol";
+import "../CometInterface.sol";
+import "../ERC20.sol";
+import "../IWETH9.sol";
 
 interface IClaimable {
     function claim(address comet, address src, bool shouldAccrue) external;
@@ -11,23 +11,24 @@ interface IClaimable {
     function claimTo(address comet, address src, address to, bool shouldAccrue) external;
 }
 
-contract Bulker {
+contract BaseBulker {
     /** General configuration constants **/
     address public immutable admin;
     address payable public immutable weth;
 
     /** Actions **/
-    uint public constant ACTION_SUPPLY_ASSET = 1;
-    uint public constant ACTION_SUPPLY_ETH = 2;
-    uint public constant ACTION_TRANSFER_ASSET = 3;
-    uint public constant ACTION_WITHDRAW_ASSET = 4;
-    uint public constant ACTION_WITHDRAW_ETH = 5;
-    uint public constant ACTION_CLAIM_REWARD = 6;
+    bytes32 public constant ACTION_SUPPLY_ASSET = 'ACTION_SUPPLY_ASSET';
+    bytes32 public constant ACTION_SUPPLY_ETH = 'ACTION_SUPPLY_ETH';
+    bytes32 public constant ACTION_TRANSFER_ASSET = 'ACTION_TRANSFER_ASSET';
+    bytes32 public constant ACTION_WITHDRAW_ASSET = 'ACTION_WITHDRAW_ASSET';
+    bytes32 public constant ACTION_WITHDRAW_ETH = 'ACTION_WITHDRAW_ETH';
+    bytes32 public constant ACTION_CLAIM_REWARD = 'ACTION_CLAIM_REWARD';
 
     /** Custom errors **/
     error InvalidArgument();
     error FailedToSendEther();
     error Unauthorized();
+    error UnhandledAction();
 
     constructor(address admin_, address payable weth_) {
         admin = admin_;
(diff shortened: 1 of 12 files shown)


contracts/bulkers/BaseBulker.sol
```diff
@@ -1,33 +1,34 @@
...
+import "../IWETH9.sol";
 
 interface IClaimable {
     function claim(address comet, address src, bool shouldAccrue) external;
 
     function claimTo(address comet, address src, address to, bool shouldAccrue) external;
 }
 
-contract Bulker {
+contract BaseBulker {
```
kevincheng96, line 14: Can't comment on line 35 so going to comment here instead. Should we go ahead and make the base even more generic by renaming `weth` to `wrappedNativeToken`?

contracts/bulkers/MainnetBulker.sol
```diff
@@ -0,0 +1,60 @@
...
+    bytes32 public constant ACTION_SUPPLY_STETH = "ACTION_SUPPLY_STETH";
+    bytes32 public constant ACTION_WITHDRAW_STETH = "ACTION_WITHDRAW_STETH";
+
+    constructor(
+        address admin_,
+        address payable weth_,
+        address payable steth_,
+        address payable wsteth_
+    ) BaseBulker(admin_, weth_) {
+        steth = steth_;
```
kevincheng96, line 20: You can derive this from `wsteth.stETH()`.

kevincheng96 (commented):
Nice, the contract looks good so far! Just a few small comments

scenario/BulkerScenario.ts
```diff
@@ -332,3 +332,24 @@ scenario(
...
+  async ({ comet, actors }) => {
+    const { betty } = actors;
+
+    const supplyGalacticCreditsCalldata = utils.defaultAbiCoder.encode(
+      ['address', 'address', 'uint'],
+      [comet.address, betty.address, exp(1, 18)]
+    );
+    const calldata = [supplyGalacticCreditsCalldata];
+    const actions = [
+      ethers.utils.formatBytes32String('ACTION_SUPPLY_GALACTIC_CREDITS')
```
kevincheng96, line 348: 👽: Comet doesn't support Galactic Credits?!

contracts/bulkers/BaseBulker.sol
scott-silver: Yeah, it's a good question. `wrappedNativeToken` will be the more accurate name once we're using the bulker for wrapped Matic and wrapped Avax.

The downside is that it will break the existing interface. With these changes, the existing, deployed Bulkers and the new MainnetBulker have the same interface (the MainnetBulker just adds a couple additional actions).

But if we update the constants to be `ACTION_SUPPLY_NATIVE_TOKEN` instead of `ACTION_SUPPLY_ETH`, then we'll need to either wrap deployment-specific scenarios or replace the deployed bulkers.

scenario/BulkerScenario.ts
scott-silver: it's an outrage! just wait until the Emperor hears about this

contracts/bulkers/MainnetBulker.sol
scott-silver: good call; updated (and made the `steth` and `wsteth` address not `payable`)

contracts/bulkers/BaseBulker.sol
kevincheng96: IMO it's fine to temporarily have both interfaces in the repo so scenarios can still use the old Bulker interface until we complete the migration on-chain. The cost of not making this change is that we'll either need a separate `BaseBulker` for the other chains or the bulker on other chains will incorrectly refer to `WMATIC` or `WAVAX` as `WETH`.

contracts/bulkers/MainnetBulker.sol
kevincheng96: Good catch, didn't catch the `payable` on my end

contracts/bulkers/IWstETH.sol
```diff
@@ -0,0 +1,23 @@
...
+    function unwrap(uint256 _wstETHAmount) external returns (uint256);
+
+    function receive() external payable;
+
+    function getWstETHByStETH(uint256 _stETHAmount) external view returns (uint256);
+    function getStETHByWstETH(uint256 _wstETHAmount) external view returns (uint256);
+
+    function stEthPerToken() external view returns (uint256);
+    function tokensPerStEth() external view returns (uint256);
+}
```
jflatow, line 23: Maybe this should live in the parent dir?

jflatow (approved):
Awesome

contracts/bulkers/IWstETH.sol
scott-silver: works for me. moved

contracts/bulkers/BaseBulker.sol
scott-silver: sounds good. updated all references to `eth`/`weth` to be `nativeToken`/`wrappedNativeToken`

kevincheng96 (approved):
Looks good 💪

contracts/bulkers/MainnetBulker.sol
```diff
@@ -2,23 +2,22 @@
...
     bytes32 public constant ACTION_WITHDRAW_STETH = "ACTION_WITHDRAW_STETH";
 
     constructor(
         address admin_,
-        address payable weth_,
-        address payable steth_,
-        address payable wsteth_
-    ) BaseBulker(admin_, weth_) {
-        steth = steth_;
+        address payable wrappedNativeToken_,
```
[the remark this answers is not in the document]
scott-silver: yep, you're right. updated

Siroph79 (commented):
test/helpers.ts

contracts/bulkers/MainnetBulker.sol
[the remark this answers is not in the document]
Siroph79: Really? Let try it too

PR #612, opened 2022-11-02
[draft] break bases into groups

scott-silver:


.github/workflows/run-scenarios.yaml
```diff
@@ -7,7 +7,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        bases: [ development, mainnet, goerli, fuji, mumbai ]
+        bases: [
```
scott-silver, line 10: grouped in no particular order.

PR #613, opened 2022-11-03
[Draft] Optimism deployment

scott-silver:


scenario/utils/relayOptimismMessage.ts
```diff
@@ -0,0 +1,81 @@
...
+an L1 address to its offset, L2 equivalent.
+
+https://optimistic.etherscan.io/address/0x4200000000000000000000000000000000000007#code
+*/
+function applyL1ToL2Alias(address: string) {
+  const offset = BigInt('0x1111000000000000000000000000000000001111');
+  return `0x${(BigInt(address) + offset).toString(16)}`;
+}
+
+export default async function relayOptimismMessage(
```
scott-silver, line 19: In order to test the Optimism proposal (that we will eventually write), we need to add token bridging to this function  like the Polygon relayer.

deployments/optimism/usdc/deploy.ts
```diff
@@ -0,0 +1,80 @@
...
+  );
+
+  // Deploy Comet
+  const deployed = await deployComet(deploymentManager, deploySpec);
+  const { comet } = deployed;
+
+  // Deploy Bulker
+  const bulker = await deploymentManager.deploy(
+    'bulker',
+    'bulkers/BaseBulker.sol',
```
kevincheng96, line 67: We might want to use the `MainnetTimelock` to have `wstETH` support since I think Lido is making a push on Optimism. We could also switch the Bulker when/if we do a ETH deploy.

scenario/utils/isBridgeProposal.ts
```diff
@@ -15,6 +15,12 @@ export async function isBridgeProposal(
       const { targets } = await governor.getActions(openProposal.id);
       return targets.map((x: string) => x.toLowerCase()).includes(fxRoot.address.toLowerCase());
     }
+    case 'optimism': {
+      const governor = await governanceDeploymentManager.getContractOrThrow('governor');
+      const optimismL1CrossDomainMessenger = await governanceDeploymentManager.getContractOrThrow('optimismL1CrossDomainMessenger');
+      const { targets } = await governor.getActions(openProposal.id);
+      return targets.includes(optimismL1CrossDomainMessenger.address);
+    }
```
kevincheng96, line 23: We should also support token bridging, which would call into the [`L1StandardBridge`](https://community.optimism.io/docs/developers/bridge/standard-bridge).

scenario/utils/relayOptimismMessage.ts
```diff
@@ -0,0 +1,81 @@
...
+    };
+
+    governanceDeploymentManager.hre.ethers.provider.on(filter, (log) => {
+      resolve(log);
+    });
+
+    setTimeout(() => {
+      reject(new Error('OptimismL1CrossDomainMessenger.SentMessage event listener timed out'));
+    }, EVENT_LISTENER_TIMEOUT);
+  });
```
kevincheng96, line 45: We can use `ethers.provider.getLogs` to avoid the complexities of setting up our own listener with a timeout. Similar to how we do it for relaying Polygon messages.

scenario/utils/relayOptimismMessage.ts
kevincheng96, line 19: Nice!

kevincheng96 (commented):
Awesome work on this!

deployments/mainnet/usdc/deploy.ts
```diff
@@ -12,6 +12,12 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
   // Mainnet -> Polygon bridge contract
   const fxRoot = await deploymentManager.existing('fxRoot', '0xfe5e5D361b2ad62c541bAb87C45a0B9B018389a2');
 
+  // Mainnet -> Optimism bridge contract
+  const optimismL1CrossDomainMessenger = await deploymentManager.existing(
+    'optimismL1CrossDomainMessenger',
+    '0x25ace71c97B33Cc4729CF772ae268934F7ab5fA1'
```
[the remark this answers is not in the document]
scott-silver: I believe this address came from the Optimism repo:

https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts/deployments/mainnet

scenario/utils/relayOptimismMessage.ts
scott-silver: agreed; the Polygon version is much improved over this iteration. we should rework this to incorporate those improvements

kevincheng96:
Here's our PR for adding COMP to Optimism (only to Optimism-goerli for now, we still have to deploy the token on Optimism mainnet): https://github.com/ethereum-optimism/ethereum-optimism.github.io/pull/294

scenario/GovernanceScenario.ts
```diff
@@ -332,4 +332,54 @@ async function fastL1ToPolygonGovernanceExecute(
...
+        [timelock.address],
+        [0],
+        ['setDelay(uint256)'],
+        [utils.defaultAbiCoder.encode(['uint'], [newTimelockDelay])]
+      ]
+    );
+
+    const sendMessageCalldata = utils.defaultAbiCoder.encode(
+      ['address', 'bytes', 'uint32'],
+      [bridgeReceiver.address, l2ProposalData, 1000000]
```
kevincheng96, line 365: We need to be careful about calculating the `gasLimit` param and make sure that it's sufficient for the L2 txn. Also, I wonder what happens if the `gasLimit` we set is too low. Logically, the L2 transaction would fail but I don't see it covered in the docs.

deployments/optimism-goerli/usdc/configuration.json
```diff
@@ -0,0 +1,56 @@
...
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000011574074074074073e15",
+    "baseBorrowSpeed": "0.0011458333333333333e15",
+    "baseMinForRewards": "10000e6"
+  },
+  "rewardTokenAddress": "0x6AF3cb766D6cd37449bfD321D961A61B0515c1BC",
+  "assets": {
+    "OP": {
+      "priceFeed": "0xE882831E58eec48B7f304482771F67e6b846733D",
```
cwang25, line 30: Trying to populating infos into c3-api to start supporting optimism. But it seems we're missing OP's address here 🤔 ? Is it '0x4200000000000000000000000000000000000042' ? 
(Just checked on testnet, it seems it's also this address, but wanted to double check :) )

kevincheng96: Yes "0x4200000000000000000000000000000000000042" is the official address for OP on both Optimism testnet and mainnet. However, on testnet, we've deployed our own OP token because there doesn't seem to be a public faucet for it and it would make using the testnet easier if we controlled the token supply. Our testnet OP token address is "0x32719c5e2e35909C8f42167315445e349e749E05".

Btw, if you are unsure which addresses we use on each chain for each deployment, you can use Spider to programmatically pull in all the relevant addresses for a deployment. See https://github.com/compound-finance/comet/tree/main#spider

cwang25: Got it thank you so much!! 🙏

deployments/optimism-goerli/usdc/deploy.ts
```diff
@@ -0,0 +1,135 @@
+import { Deployed, DeploymentManager } from '../../../plugins/deployment_manager';
+import { DeploySpec, deployComet, exp, wait } from '../../../src/deploy';
+
+const clone = {
+  op: '0x4200000000000000000000000000000000000042'
+};
+
+const secondsPerDay = 24 * 60 * 60;
+
+const GOERLI_TIMELOCK = '0x8Fa336EB4bF58Cfc508dEA1B0aeC7336f55B1399';
```
cwang25, line 10: Curious if local timelock on optimism-goerli should be 'https://goerli-optimism.etherscan.io/address/0x4c8011724f2ece7b6ce4b13c90fe5036c9e1eb2d' instead of 0x8Fa336EB4bF58Cfc508dEA1B0aeC7336f55B1399 ? 🤔 

Went to check 0x8Fa336EB4bF58Cfc508dEA1B0aeC7336f55B1399, but it got empty address.

kevincheng96: So this is the Ethereum Goerli timelock, not the Optimism Goerli timelock 😃

We hardcode it here because it is provided as a constructor arg to the `BridgeReceiver`, which needs to know what L1 contract messages should be coming from.

PR #614, opened 2022-11-04
Impersonate flag for migrate task

ajb413:
Add an impersonate flag for migration simulations so developers can test their new migration script.

### DIFF
--- a/plugins/scenario/World.ts
+++ b/plugins/scenario/World.ts
@@ -84,11 +84,7 @@ export class World {
   }
 
   async impersonateAddress(address: string, value?: bigint): Promise<SignerWithAddress> {
-    if (value) {
-      const signer = await this.deploymentManager.getSigner();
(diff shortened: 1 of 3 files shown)


jflatow (commented):
Badass 💪  I think this is exactly right, just want to tweak how you set the first signer

ajb413:
> Badass 💪 I think this is exactly right, just want to tweak how you set the first signer

LMK what you have in mind specifically.

jflatow:
> > Badass 💪 I think this is exactly right, just want to tweak how you set the first signer
> 
> LMK what you have in mind specifically.

Added in a commit, but I also think we should maybe move/refactor the `World` `impersonateAddress` fn and use it to give this one a grant of ETH as well, I will follow up later today with this

kevincheng96 (approved):
LGTM. But looks like scenarios are failing.

PR #615, opened 2022-11-04
PriceConstraint

scott-silver:


scenario/context/CometContext.ts
```diff
@@ -143,6 +144,36 @@ export class CometContext {
     return this;
   }
 
+  async updatePriceFeeds(newPrices: Record<string, number>) {
+    const comet = await this.getComet();
+    const baseToken = await comet.baseToken();
+
+    const newPriceFeeds: Record<string, string> = {};
+    for (const asset in newPrices) {
+      // XXX use the correct alias
```
jflatow, line 153: What alias?

scenario/constraints/PriceConstraint.ts
```diff
@@ -0,0 +1,43 @@
...
+  async solve(requirements: R, _initialContext: T) {
+    return async function (ctx: T): Promise<T> {
+      const prices = requirements.prices;
+      if (prices !== undefined) {
+        const assetPriceMap = {};
+        for (const [assetAlias, price] of Object.entries(prices)) {
+          const cometAsset = await getAssetFromName(assetAlias, ctx);
+          assetPriceMap[cometAsset.address] = price;
+        }
+        await ctx.updatePriceFeeds(assetPriceMap);
```
jflatow, line 17: Ah interesting, I was imagining we might do something crazy like try to manipulate the prices in different ways depending on the price feed type, but this is a nice idea

jflatow (commented):
💪  Thanks for getting this going!

scenario/context/CometContext.ts
scott-silver: like `weth:priceFeed`

scenario/constraints/PriceConstraint.ts
scott-silver: how else could you manipulate the price, in a way that would actually cause the Chainlink feeds to get updated?

scenario/context/CometContext.ts
jflatow: isn't that what it's doing now?

scenario/constraints/PriceConstraint.ts
jflatow: Maybe impersonate and post?

kevincheng96 (dismissed):
Awesome, this looks super useful!

scenario/constraints/PriceConstraint.ts
kevincheng96: This is a really creative approach and, surprisingly, might be the most straightforward one as well. I would've thought posting prices would be easier, but doesn't seem like there's a easy way to do that with Chainlink's contracts AFAICT.

scenario/context/CometContext.ts
scott-silver: no, right now it's using the asset address, not the alias (e.g. `0x123...:priceFeed`), which is pretty useless

scenario/constraints/PriceConstraint.ts
scott-silver: what do you think, Jared? are you alright with this approach or should I try to come up with something different?

jflatow: I think its a great idea, but I'm a little hesistant/concerned that it changes the contract setup for the test rather than mutating real contracts, I think I'd like to look into why pretending to be Chainlink is so hard

kevincheng96: Agreed, it's worth spending some time to see if we can do it directly via Chainlink. I only briefly looked at their contracts.

jflatow (approved):

PR #616, opened 2022-11-09
Liquidation bot improvements

scott-silver:


scripts/liquidation_bot/deploy.ts
```diff
@@ -41,9 +41,27 @@ async function main() {
...
+      [
+        '0xc00e94Cb662C3520282E6f5717214004A7f26888', // COMP
+        '0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599', // WBTC
+        '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2', // WETH
+        '0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984', // UNI
+        '0x514910771AF9Ca656af840dff83E8264EcF986CA', // LINK
+      ],
+      [
+        true,
+        true,
```
scott-silver, line 53: By setting this to `true` for `WBTC`, we should now be using this pool for selling `WBTC`:

https://info.uniswap.org/#/pools/0xcbcdf9626bc03e24f779434178a73a0b4bad62ed

scott-silver: With the new setters, we can update these configs, as well

kevincheng96 (approved):
LGTM!

scripts/liquidation_bot/deploy.ts
kevincheng96: Configs look good as well

PR #617, opened 2022-11-09
Look for BuyCollateral opportunities even if no accounts are absorbable

kevincheng96:
Currently, the bot will only `BuyCollateral` if there are accounts to be absorbed. This change will make it so the bot will also look for `BuyCollateral` opportunities separately.

### DIFF
--- a/scripts/liquidation_bot/index.ts
+++ b/scripts/liquidation_bot/index.ts
@@ -12,7 +12,7 @@ import {
 } from './liquidateUnderwaterBorrowers';
 
 const loopDelay = 5000;
-const updateAssetsDelay = 86_400_000; // 1 day
+const loopsUntilUpdateAssets = 1000;
 let assets: Asset[] = [];
 
 async function main() {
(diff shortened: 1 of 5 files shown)


test/liquidation/liquidation-bot-script.ts
```diff
@@ -1,8 +1,8 @@
 import { expect, exp } from '../helpers';
-import liquidateUnderwaterBorrowers from '../../scripts/liquidation_bot/liquidateUnderwaterBorrowers';
-import makeLiquidatableProtocol, { forkMainnet, resetHardhatNetwork } from './makeLiquidatableProtocol';
+import { arbitragePurchaseableCollateral, getAssets, hasPurchaseableCollateral, liquidateUnderwaterBorrowers } from '../../scripts/liquidation_bot/liquidateUnderwaterBorrowers';
+import { forkMainnet, makeProtocol, makeLiquidatableProtocol, resetHardhatNetwork } from './makeLiquidatableProtocol';
 
-describe('Liquidation Bot', function () {
+describe.only('Liquidation Bot', function () {
```
scott-silver, line 5: leftover `.only`

scott-silver (approved):
🙌

scripts/liquidation_bot/index.ts
```diff
@@ -45,19 +52,34 @@ async function main() {
   ) as Liquidator;
 
   let lastBlockNumber: number;
-
+  let loopsUntilUpdateAssets = updateAssetsDelay / loopDelay;
```
scott-silver, line 55: we probably shouldn't reuse `loopDelay`, given that `loopDelay` is used in a very different manner.

why not just set `loopUntilUpdateAssets` to some arbitrary value like 100, or 1,000? so every 100 loops we'll check for new assets

PR #618, opened 2022-11-10
Add Flashbots support to the liquidation bot

kevincheng96:
Transactions can be sent to Flashbots by specifying `USE_FLASHBOTS=true` in env variables.


scott-silver (approved):
💪

PR #619, opened 2022-11-10
Add googleCloudLog, improve logging of info; alert on failed liquidation

scott-silver:


scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -22,15 +23,18 @@ async function attemptLiquidation(
...
       pairToken: daiPool.tokenAddress,
       poolFee: daiPool.poolFee
     });
-    console.log(`Successfully liquidated ${targetAddresses}`);
+    googleCloudLog(LogSeverity.INFO, `Successfully liquidated ${targetAddresses} via ${liquidator.address}`);
   } catch (e) {
-    console.log(`Failed to liquidate ${targetAddresses}`);
-    console.log(e.message);
+    googleCloudLog(
+      LogSeverity.ALERT,
```
kevincheng96, line 35: Does this actually alert someone right now?

scott-silver: not yet, but it means we can construct a Policy that alerts if there are more than X number of `ALERT` level logs within a period of time

kevincheng96: Nice

kevincheng96 (approved):
👍

PR #620, opened 2022-11-13
add Sushi Swap support to Liquidator.sol

scott-silver:


scenario/LiquidationBotScenario.ts
```diff
@@ -0,0 +1,115 @@
...
+const RECIPIENT = '0xe8F0c9059b8Db5B863d48dB8e8C1A09f97D3B991';
+const UNISWAP_ROUTER = '0xe592427a0aece92de3edee1f18e0157c05861564';
+const SUSHISWAP_ROUTER = '0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9F';
+
+enum Exchange {
+  Uniswap,
+  SushiSwap
+}
+
+scenario.only(
```
scott-silver: delete `.only`

scenario/LiquidationBotScenario.ts
```diff
@@ -0,0 +1,120 @@
...
+    },
+    async ({ comet, actors }, _context, world) => {
+      const { albert, betty } = actors;
+
+      // const liquidator = await world.deploymentManager.existing(
+      //   'liquidator',
+      //   LIQUIDATOR_ADDRESS
+      // ) as Liquidator;
+
+      const liquidator = await world.deploymentManager.deploy(
```
scott-silver, line 51: Deploying in a scenario is obviously not ideal.

I think we avoided adding the Liquidator contract to `roots.json` when we first developed it since we felt like it was not actually a part of Comet.

I'm thinking we should revisit that idea, though. It seems like having a functioning Liquidation contract is more critical than we'd originally thought

contracts/liquidator/Liquidator.sol
```diff
@@ -186,19 +233,19 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
+        uint256 amountOut = uniswapRouter.exactInputSingle(
             ISwapRouter.ExactInputSingleParams({
                 tokenIn: swapToken,
                 tokenOut: baseToken,
                 fee: poolFee,
                 recipient: address(this),
                 deadline: block.timestamp,
                 amountIn: swapAmount,
-                amountOutMinimum: 0,
+                amountOutMinimum: amountOutMin,
```
scott-silver, line 248: specifying an `amountOutMinimum` helps throw an error at the preferable place in the code, since we'll get get errors specifying that we didn't receive enough in return

hayesgm: cool

hayesgm (commented):
This is amazing. Great upgrade.

contracts/liquidator/Liquidator.sol
```diff
@@ -126,51 +154,70 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
+        PoolConfig memory config = poolConfigs[asset];
         // If asset is not found, proceed with default pool config
         if (config.fee == 0) {
-            return UniswapPoolConfig({
+            return PoolConfig({
                 fee: DEFAULT_POOL_FEE,
-                isLowLiquidity: false
+                isLowLiquidity: false,
+                exchange: Exchange.Uniswap
             });
```
kevincheng96, line 186: IMO, we should probably just revert instead of swapping through some pool that will likely either revert or eat up our profits.

kevincheng96: But we don't have to make that change in this PR.

kevincheng96: Actually, your `amountOutMin` change probably takes care of this issue anyways

scenario/LiquidationBotScenario.ts
```diff
@@ -0,0 +1,120 @@
...
+
+      expect(await comet.isLiquidatable(albert.address)).to.be.true;
+      expect(await comet.collateralBalanceOf(albert.address, collateralAssetAddress)).to.be.greaterThan(0);
+
+      await liquidator.connect(betty.signer).initFlash({
+        accounts: [albert.address],
+        pairToken: daiPool.tokenAddress,
+        poolFee: daiPool.poolFee
+      });
+
```
kevincheng96, line 115: We should also check that the liquidation was profitable (maybe either checking some event that conveys the profit or the token balance of the recipient).

scenario/LiquidationScenario.ts
```diff
@@ -25,7 +25,7 @@ invariant:
 isolating for timeElapsed:
 timeElapsed = -liquidationMargin / (baseBalanceOf * price / baseScale) / (borrowRate / factorScale);
 */
-async function timeUntilUnderwater({ comet, actor, fudgeFactor = 0n }: { comet: CometInterface, actor: CometActor, fudgeFactor?: bigint }): Promise<number> {
+export async function timeUntilUnderwater({ comet, actor, fudgeFactor = 0n }: { comet: CometInterface, actor: CometActor, fudgeFactor?: bigint }): Promise<number> {
```
kevincheng96, line 28: Move to util?

kevincheng96 (approved):
Awesome work, this looks great!

scenario/LiquidationBotScenario.ts
scott-silver: added a test that the balance of the recipient address is increased

kevincheng96 (approved):
Looks good!

scenario/LiquidationScenario.ts
scott-silver: moved

PR #621, opened 2022-11-14
check for useFlashbots before down-casing

scott-silver:
`useFlashbots` is undefined if you don't define a `USE_FLASHBOTS` environment variable; we should check for its existence before calling `.toLowerCase` on it (which will throw an error otherwise)


kevincheng96 (approved):
Good catch!

PR #622, opened 2022-11-15
Liquidate via Flashbots bundles

kevincheng96:
This PR implements liquidating positions via the private mempool (to prevent front-running) by sending bundles to Flashbots. Both the Flashbots and non-Flashbots paths have been tested E2E on Goerli.

### DIFF
--- a/scripts/liquidation_bot/index.ts
+++ b/scripts/liquidation_bot/index.ts
@@ -26,7 +26,7 @@ async function main() {
   if (!deployment) {
     throw new Error('missing required env variable: DEPLOYMENT');
   }
-  if (!ethPk) {
+  if (useFlashbots && !ethPk) {
     throw new Error('missing required env variable: ETH_PK');
   }
 
--- a/scripts/liquidation_bot/index.ts
+++ b/scripts/liquidation_bot/index.ts
@@ -11,7 +11,7 @@ import {
   Asset
 } from './liquidateUnderwaterBorrowers';
 import { FlashbotsBundleProvider } from '@flashbots/ethers-provider-bundle';
-import { providers, Wallet } from 'ethers';
+import { Signer, Wallet } from 'ethers';
 import googleCloudLog, { LogSeverity } from './googleCloudLog';
 
 const loopDelay = 5000;
@@ -19,12 +19,15 @@ const loopsUntilUpdateAssets = 1000;
 let assets: Asset[] = [];
 
 async function main() {
-  let { DEPLOYMENT: deployment, LIQUIDATOR_ADDRESS: liquidatorAddress, USE_FLASHBOTS: useFlashbots } = process.env;
+  let { DEPLOYMENT: deployment, LIQUIDATOR_ADDRESS: liquidatorAddress, USE_FLASHBOTS: useFlashbots, ETH_PK: ethPk } = process.env;
   if (!liquidatorAddress) {
     throw new Error('missing required env variable: LIQUIDATOR_ADDRESS');
   }
   if (!deployment) {
     throw new Error('missing required env variable: DEPLOYMENT');
+  }
+  if (!ethPk) {
+    throw new Error('missing required env variable: ETH_PK');
   }
 
   const network = hre.network.name;
@@ -45,12 +48,12 @@ async function main() {
   );
   await dm.spider();
 
-  const signer = await dm.getSigner();
   const contracts = await dm.contracts();
   const comet = contracts.get('comet') as CometInterface;
 
   // Flashbots provider requires passing in a standard provider
   let flashbotsProvider: FlashbotsBundleProvider;
+  let signer: Signer;
   if (useFlashbots && useFlashbots.toLowerCase() === 'true') {
     // XXX use a designated auth signer
     // `authSigner` is an Ethereum private key that does NOT store funds and is NOT your bot's primary key.
@@ -60,12 +63,12 @@ async function main() {
 
     if (network === 'mainnet') {
       flashbotsProvider = await FlashbotsBundleProvider.create(
-        signer.provider! as providers.BaseProvider, // a normal ethers.js provider, to perform gas estimations and nonce lookups
+        hre.ethers.provider, // a normal ethers.js provider, to perform gas estimations and nonce lookups
         authSigner, // ethers.js signer wallet, only for signing request payloads, not transactions
       );
     } else if (network === 'goerli') {
       flashbotsProvider = await FlashbotsBundleProvider.create(
-        signer.provider! as providers.BaseProvider, // a normal ethers.js provider, to perform gas estimations and nonce lookups
+        hre.ethers.provider, // a normal ethers.js provider, to perform gas estimations and nonce lookups
         authSigner, // ethers.js signer wallet, only for signing request payloads, not transactions
         'https://relay-goerli.flashbots.net',
         'goerli'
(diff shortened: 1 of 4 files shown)


scott-silver:
> Both the Flashbots and non-Flashbots paths have been tested E2E on Goerli.

You mentioned that, when using Flashbots, the transaction will fail silently (it just won't get picked up by Flashbots and after some time we'll assume that it has failed).

What about the non-Flashbots case? Are transaction failures immediate?

scripts/liquidation_bot/index.ts
```diff
@@ -11,21 +11,24 @@ import {
...
 async function main() {
-  let { DEPLOYMENT: deployment, LIQUIDATOR_ADDRESS: liquidatorAddress, USE_FLASHBOTS: useFlashbots } = process.env;
+  let { DEPLOYMENT: deployment, LIQUIDATOR_ADDRESS: liquidatorAddress, USE_FLASHBOTS: useFlashbots, ETH_PK: ethPk } = process.env;
   if (!liquidatorAddress) {
     throw new Error('missing required env variable: LIQUIDATOR_ADDRESS');
   }
   if (!deployment) {
     throw new Error('missing required env variable: DEPLOYMENT');
   }
+  if (!ethPk) {
```
scott-silver, line 29: `ethPk` is only required if `useFlashbots` is true, right?

kevincheng96: > You mentioned that, when using Flashbots, the transaction will fail silently (it just won't get picked up by Flashbots and after some time we'll assume that it has failed).
> 
> What about the non-Flashbots case? Are transaction failures immediate?

Actually, I figured out a way to block on the Flashbots txn and get a response on block inclusion or not. So we can now add alerting if a txn does not get included X amount of times (but probably better to just alert if there's purchaseable collateral available for X consecutive blocks).

Transaction failures are never silent. If a txn would fail, Hardhat would always convey that.

kevincheng96:
> You mentioned that, when using Flashbots, the transaction will fail silently (it just won't get picked up by Flashbots and after some time we'll assume that it has failed).

Actually, I figured out a way to block on the Flashbots txn and get a response on block inclusion or not. So we can now add alerting if a txn does not get included X amount of times (but probably better to just alert if there's purchaseable collateral available for X consecutive blocks).

> What about the non-Flashbots case? Are transaction failures immediate?

Transaction failures are never silent. If a txn would fail, the gas estimation should always fail and we would be notified in the logs.

scripts/liquidation_bot/index.ts
kevincheng96: Yeah that's correct. I can change this check.

scripts/liquidation_bot/sendTransaction.ts
```diff
@@ -0,0 +1,64 @@
...
+      success = false;
+      googleCloudLog(LogSeverity.INFO, 'Block passed without inclusion');
+    } else if (resolution === FlashbotsBundleResolution.AccountNonceTooHigh) {
+      success = false;
+      googleCloudLog(LogSeverity.INFO, 'Account nonce too high');
+    }
+  } else {
+    success = false;
+    googleCloudLog(LogSeverity.ALERT, `Error while sending Flashbots bundle: ${bundleReceipt.error}`);
+  }
```
kevincheng96, line 45: @scott-silver We check for Flashbots block inclusion here. But note that this is blocking, which is why I brought up parallelization.

scripts/liquidation_bot/index.ts
scott-silver: it's up to you; the Pulumi configs all contain an `ETH_PK` right now, so it's not a huge deal

scripts/liquidation_bot/sendTransaction.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (resolution === FlashbotsBundleResolution.BundleIncluded) {
+      success = true;
+      googleCloudLog(LogSeverity.INFO, 'Bundle included!');
+    } else if (resolution === FlashbotsBundleResolution.BlockPassedWithoutInclusion) {
+      // XXX alert if too many attempts are not included in a block
+      success = false;
+      googleCloudLog(LogSeverity.INFO, 'Block passed without inclusion');
+    } else if (resolution === FlashbotsBundleResolution.AccountNonceTooHigh) {
+      success = false;
+      googleCloudLog(LogSeverity.INFO, 'Account nonce too high');
```
scott-silver, line 40: should we alert on this? `LogSeverity.ALERT`?

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -19,49 +20,45 @@ export interface Asset {
...
-    googleCloudLog(LogSeverity.INFO, 'Sending a public txn');
-    await signerWithFlashbots.signer.sendTransaction(txn);
+const flashLoanPools = {
+  'mainnet': {
+    // DAI pool
+    tokenAddress: '0x6B175474E89094C44Da98b954EedeAC495271d0F',
+    poolFee: 100
+  },
+  'goerli': {
+    // WETH pool
```
scott-silver, line 30: does it matter that we use `WETH` both as the pair token to initiate the flash loan and also as an intermediary step in swapping?

scripts/liquidation_bot/sendTransaction.ts
kevincheng96: Sure, done.

scripts/liquidation_bot/index.ts
kevincheng96: I updated the check because it's strictly only needed when `useFlashbots` is true.

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
scott-silver: once we merge #623 we can also set a conservative `maxCollateralToPurchase` value if that will help use a different pool

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -19,49 +20,45 @@ export interface Asset {
...
+      gasLimit: Math.ceil(1.1 * (await liquidator.estimateGas.initFlash(calldata)).toNumber()),
+      gasPrice: Math.ceil(1.1 * (await hre.ethers.provider.getGasPrice()).toNumber()),
     });
-    await sendTxn(txn, signerWithFlashbots);
-    googleCloudLog(LogSeverity.INFO, `Successfully liquidated ${targetAddresses} via ${liquidator.address}`);
+    const success = await sendTxn(txn, signerWithFlashbots);
+    if (success) {
+      googleCloudLog(LogSeverity.INFO, `Successfully liquidated ${targetAddresses} via ${liquidator.address}`);
+    } else {
+      googleCloudLog(LogSeverity.ALERT, `Failed to liquidate ${targetAddresses} via ${liquidator.address}`);
```
scott-silver, line 60: is it possible to add a stack trace or error message for a failed transaction?

scott-silver: maybe it's already being logged on ln 44 of `sendTransaction`?

kevincheng96: Flashbots won't include transactions that revert, so `googleCloudLog(LogSeverity.INFO, 'Block passed without inclusion')` would get logged. Transactions that revert deterministically won't even make it to that point because gas estimation fails when constructing the txn, so you just end up with the alert log in the catch block below.

scott-silver (approved):

PR #623, opened 2022-11-16
add maxCollateralToPurchase to Liquidator.sol

scott-silver:
adds a `maxCollateralToPurchase` field to the Liquidator's pool config. This lets us specify a maximum amount of the given collateral asset to purchase, so that we don't try to purchase all the liquidity in a given pool, suffer a ton of slippage, and then fail to make a profit/pay back the flash loan.

Added a bunch of scenarios to flex this behavior. Scenarios have given us a general sense of the depth of the current Uniswap/SushiSwap pools that we're using; but of course, these things are subject to change as real-world conditions evolve.

### DIFF
--- a/.github/workflows/run-forge-tests.yaml
+++ b/.github/workflows/run-forge-tests.yaml
@@ -21,7 +21,7 @@ jobs:
         run: forge install
 
       - name: Run tests
-        run: forge test -vvv
+        run: forge test -vvv --via-ir
         env:
           ETHERSCAN_KEY: ${{ secrets.ETHERSCAN_KEY }}
           SNOWTRACE_KEY: ${{ secrets.SNOWTRACE_KEY }}
--- a/contracts/liquidator/Liquidator.sol
+++ b/contracts/liquidator/Liquidator.sol
@@ -29,14 +29,17 @@ interface IUniswapV2Router {
  */
 contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, PeripheryPayments {
     /** Errors */
+    error InsufficientAmountOut(address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut, uint256 amountOutMin, Exchange exchange, uint24 fee);
     error InsufficientBalance(uint256 available, uint256 required);
     error InvalidExchange();
+    error InvalidPoolConfig();
     error Unauthorized();
 
     /** Events **/
     event Absorb(address indexed initiator, address[] accounts);
+    event AbsorbedWithoutBuyingCollateral();
     event Pay(address indexed token, address indexed payer, address indexed recipient, uint256 value);
-    event Swap(address indexed tokenIn, address indexed tokenOut, Exchange exchange, uint24 fee, uint256 amountIn, uint256 amountOut);
+    event Swap(address indexed tokenIn, address indexed tokenOut, uint256 amountIn, uint256 amountOut, Exchange exchange, uint24 fee);
 
     /** Structs needed for Uniswap flash swap **/
     struct FlashParams {
@@ -58,10 +61,13 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
         SushiSwap
     }
 
+    // XXX make this less gassy; rearrange fields
     struct PoolConfig {
         bool isLowLiquidity;
         uint24 fee;
         Exchange exchange;
+        uint256 maxCollateralToPurchase;
+        bool isSet;
     }
 
     /** Liquidator configuration constants **/
@@ -108,6 +114,8 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
      * @param _assets The list of assets for pool configs
      * @param _lowLiquidityPools The list of boolean indicators if pool is low liquidity and requires ETH swap
      * @param _poolFees The list of given poolFees for assets
+     * @param _exchanges The list of given exchanges for assets
+     * @param _maxCollateralToPurchases The list of given maxSafeExchanges for assets
      **/
     constructor(
         address _recipient,
@@ -120,7 +128,8 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
         address[] memory _assets,
         bool[] memory _lowLiquidityPools,
         uint24[] memory _poolFees,
-        Exchange[] memory _exchanges
+        Exchange[] memory _exchanges,
+        uint256[] memory _maxCollateralToPurchases
     ) PeripheryImmutableState(_factory, _WETH9) {
         require(_assets.length == _lowLiquidityPools.length, "Wrong data");
         require(_assets.length == _poolFees.length, "Wrong data");
@@ -139,10 +148,13 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
             bool lowLiquidity = _lowLiquidityPools[i];
             uint24 poolFee = _poolFees[i];
             Exchange exchange = _exchanges[i];
+            uint256 maxCollateralToPurchase = _maxCollateralToPurchases[i];
             poolConfigs[asset] = PoolConfig({
                 isLowLiquidity: lowLiquidity,
                 fee: poolFee,
-                exchange: exchange
+                exchange: exchange,
+                maxCollateralToPurchase: maxCollateralToPurchase,
+                isSet: true
             });
         }
     }
@@ -157,7 +169,8 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
         address[] memory _assets,
         bool[] memory _lowLiquidityPools,
         uint24[] memory _poolFees,
-        Exchange[] memory _exchanges
+        Exchange[] memory _exchanges,
+        uint256[] memory _maxCollateralToPurchases
     ) external {
         if (msg.sender != admin) revert Unauthorized();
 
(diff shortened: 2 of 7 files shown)


scripts/liquidation_bot/deploy.ts
```diff
@@ -71,6 +75,13 @@ async function main() {
         Exchange.Uniswap,   // WETH
         Exchange.Uniswap,   // UNI
         Exchange.Uniswap    // LINK
+      ],
+      [
+        BigNumber.from(exp(500,18)),    // COMP
```
scott-silver, line 80: These are the actual values that we'll deploy with.

If we want to be conservative, we could even cut these in half.

contracts/liquidator/Liquidator.sol
```diff
@@ -108,6 +114,8 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
      * @param _assets The list of assets for pool configs
      * @param _lowLiquidityPools The list of boolean indicators if pool is low liquidity and requires ETH swap
      * @param _poolFees The list of given poolFees for assets
+     * @param _exchanges The list of given exchanges for assets
+     * @param _maxCollateralToPurchases The list of given maxSafeExchanges for assets
```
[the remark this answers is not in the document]
kevincheng96: +1 sounds more correct

contracts/liquidator/Liquidator.sol
```diff
@@ -29,14 +29,17 @@ interface IUniswapV2Router {
...
     /** Errors */
+    error InsufficientAmountOut(address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut, uint256 amountOutMin, Exchange exchange, uint24 fee);
     error InsufficientBalance(uint256 available, uint256 required);
     error InvalidExchange();
+    error InvalidPoolConfig();
     error Unauthorized();
 
     /** Events **/
     event Absorb(address indexed initiator, address[] accounts);
+    event AbsorbedWithoutBuyingCollateral();
```
kevincheng96, line 40: nit: we've been using verbs for events so `AbsorbWithoutBuyingCollateral` would be more consistent

contracts/liquidator/Liquidator.sol
```diff
@@ -247,11 +260,17 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
                 deadline: block.timestamp,
                 amountIn: swapAmount,
-                amountOutMinimum: amountOutMin,
+                amountOutMinimum: 0,
                 sqrtPriceLimitX96: 0
             })
         );
-        emit Swap(swapToken, baseToken, Exchange.Uniswap, poolFee, swapAmount, amountOut);
+
+        if (amountOut < amountOutMin) {
```
kevincheng96, line 268: Let's leave a note here to explain that we defer the `amountOutMin` check to get better error messaging.

contracts/liquidator/Liquidator.sol
```diff
@@ -284,14 +303,19 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
...
-            amountOutMin,
+            0,
             path,
             address(this),
             block.timestamp
         );
         uint256 amountOut = amounts[amounts.length - 1];
 
-        emit Swap(swapToken, baseToken, Exchange.SushiSwap, 3000, swapAmount, amountOut);
+        if (amountOut < amountOutMin) {
```
kevincheng96, line 313: Same note here.

scripts/liquidation_bot/deploy.ts
kevincheng96: LINK seems a bit high, we might want to lower it a bit to be safe.

scenario/LiquidationBotScenario.ts
```diff
@@ -18,46 +22,188 @@ enum Exchange {
...
+          comet.address, // _comet
+          UNISWAP_V3_FACTORY_ADDRESS, // _factory
+          WETH9, // _WETH9
+          10e6, // _liquidationThreshold,
+          [
+            '0xc00e94Cb662C3520282E6f5717214004A7f26888', // COMP
+            '0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599', // WBTC
+            '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2', // WETH
+            '0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984', // UNI
+            '0x514910771AF9Ca656af840dff83E8264EcF986CA', // LINK
```
kevincheng96, line 94: You can pull these from `assets` on line 76.

contracts/liquidator/Liquidator.sol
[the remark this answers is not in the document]
scott-silver: updated

scripts/liquidation_bot/deploy.ts
scott-silver: Cranked them all down (except for COMP, which was already relatively low). Let me know what you think

contracts/liquidator/Liquidator.sol
scott-silver: good call. updated

contracts/liquidator/Liquidator.sol
scott-silver: added note

contracts/liquidator/Liquidator.sol
scott-silver: added note

scenario/LiquidationBotScenario.ts
```diff
@@ -137,13 +292,382 @@ for (let i = 0; i < MAX_ASSETS; i++) {
         poolFee: daiPool.poolFee
       });
 
+      // confirm that Albert position has been abosrbed
       expect(await comet.isLiquidatable(albert.address)).to.be.false;
       expect(await comet.collateralBalanceOf(albert.address, collateralAssetAddress)).to.eq(0);
 
-      // XXX confirm that protocol reserves have increased
-
```
kevincheng96, line 144: We should also confirm that not all the collateral has been purchased.

scenario/LiquidationBotScenario.ts
```diff
@@ -18,46 +22,188 @@ enum Exchange {
...
+    ' == 10000',
+    // UNI
+    ' == 250000',
+    // LINK
+    ' == 500000',
   ];
+
   scenario(
-    `LiquidationBot > liquidates an underwater position for $asset${i}`,
+    `LiquidationBot > liquidates large position of $asset${i}, by setting maxCollateralToPurchase`,
```
kevincheng96, line 196: nit: "partially liquidates large position ..."

scenario/LiquidationBotScenario.ts
```diff
@@ -137,13 +292,382 @@ for (let i = 0; i < MAX_ASSETS; i++) {
...
+        ],
+        [
+          constants.MaxUint256, // COMP
+          constants.MaxUint256, // WBTC
+          constants.MaxUint256, // WETH
+          constants.MaxUint256, // UNI
+          constants.MaxUint256, // LINK
+        ]
+      ]
+    );
```
kevincheng96, line 370: Could use a helper function for this since it's being called in every scenario.

kevincheng96: I guess it might be slightly hard to define one since we're tweaking certain parts of the config in each scenario

scenario/LiquidationBotScenario.ts
scott-silver: updated

scripts/liquidation_bot/deploy.ts
kevincheng96: I think all the other numbers looked fine, so I actually prefer for them to stay the same (other than LINK). IMO, we don't want to be overly cautious with these numbers because it severely hurts the economics of liquidations, both in terms of gas costs and the fact that it takes much longer to liquidate (worst price for both protocol and the liquidator).

kevincheng96 (approved):
Awesome getting these scenarios up and running 💪

scenario/LiquidationBotScenario.ts
scott-silver: updated

scenario/LiquidationBotScenario.ts
scott-silver: added a check that there are at least 10 units of the collateral asset after the partial liquidation

.github/workflows/run-forge-tests.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
         run: forge install
 
       - name: Run tests
-        run: forge test -vvv
+        run: forge test -vvv --via-ir
```
scott-silver, line 24: This PR has been hitting the following failure when forge compiles the contracts:

image

```
CompilerError: Stack too deep. Try compiling with `--via-ir` (cli) or the equivalent `viaIR: true` (standard JSON) while enabling the optimizer. Otherwise, try removing local variables. When compiling inline assembly: Variable headStart is 1 slot(s) too deep inside the stack.
```

Compiling with `--via-ir` seems to work fine, though.

scenario/LiquidationBotScenario.ts
scott-silver: yeah, it would only save us a few lines that don't change across tests

scenario/ConstraintScenario.ts
```diff
@@ -94,7 +94,9 @@ scenario(
   }
 );
 
-scenario(
+// XXX unable to source sufficient $asset0 for mainnet, goerli to borrow up to
+// 75% utilization
+scenario.skip(
```
scott-silver, line 99: This scenario is failing on `main` for `mainnet` and `goerli`

PR #625, opened 2022-11-18
Price feeds for WETH deployment + Bulker changes for OZ audit

kevincheng96:
This PR implements and modifies price feeds to support the upcoming WETH deployment. The favored plan so far is to use ETH-denominated price feeds as opposed to USD price feeds, but stick with using 8 decimals for prices to avoid having to change the `Comet` and `Configurator` implementations. 

This would require: 
- A new wrapper price feed (`ScalingPriceFeed.sol`) that scales prices up or down to 8 decimals
- A new `ConstantPriceFeed` that always returns 1e8 for the `WETH` base asset, since should always hold a 1:1 value with ETH
- Modifications to the `WstETHPriceFeed` to return prices in terms of ETH instead of USD

This is an alternative approach to #626, which is a more complex change but could be a better long-term solution.

*Note: This PR also now contains the changes from #634 and #635, which address some suggestions made by OZ for their audit of `WstETHPriceFeed` and `Bulker`.*

### DIFF
--- a/contracts/ConstantPriceFeed.sol
+++ b/contracts/ConstantPriceFeed.sol
@@ -1,12 +1,9 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.15;
 
-import "./vendor/@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
+import "./IPriceFeed.sol";
 
-contract ConstantPriceFeed is AggregatorV3Interface {
-    /** Custom errors **/
-    error NotImplemented();
-
+contract ConstantPriceFeed is IPriceFeed {
     /// @notice Version of the price feed
     uint public constant override version = 1;
 
@@ -26,13 +23,6 @@ contract ConstantPriceFeed is AggregatorV3Interface {
     constructor(uint8 decimals_, int256 constantPrice_) {
         decimals = decimals_;
         constantPrice = constantPrice_;
-    }
-
-    /**
-     * @notice Unimplemented function required to fulfill AggregatorV3Interface; always reverts
-     **/
-    function getRoundData(uint80 _roundId) override external pure returns (uint80, int256, uint256, uint256, uint80) {
-        revert NotImplemented();
     }
 
     /**
--- a/contracts/ConstantPriceFeed.sol
+++ b/contracts/ConstantPriceFeed.sol
@@ -32,13 +32,7 @@ contract ConstantPriceFeed is AggregatorV3Interface {
     /**
      * @notice Unimplemented function required to fulfill AggregatorV3Interface; always reverts
      **/
-    function getRoundData(uint80 _roundId) override external view returns (
-        uint80 roundId,
-        int256 answer,
-        uint256 startedAt,
-        uint256 updatedAt,
-        uint80 answeredInRound
-    ) {
+    function getRoundData(uint80 _roundId) override external pure returns (uint80, int256, uint256, uint256, uint80) {
         revert NotImplemented();
     }
 
--- a/contracts/ConstantPriceFeed.sol
+++ b/contracts/ConstantPriceFeed.sol
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+import "./vendor/@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
+
+contract ConstantPriceFeed is AggregatorV3Interface {
+    /** Custom errors **/
+    error InvalidInt256();
+    error NotImplemented();
+
+    /// @notice Version of the price feed
+    uint public constant override version = 1;
+
+    /// @notice Description of the price feed
+    string public constant description = "Constant price feed";
+
+    /// @notice Number of decimals for returned prices
+    uint8 public immutable override decimals;
+
+    /// @notice The constant price
+    int private immutable CONSTANT_PRICE;
+
+    /**
+     * @notice Construct a new scaling price feed
+     * @param decimals_ The number of decimals for the returned prices
+     **/
+    constructor(uint8 decimals_) {
+        decimals = decimals_;
+        CONSTANT_PRICE = signed256(10 ** decimals_);
+    }
+
+    /**
+     * @notice Unimplemented function required to fulfill AggregatorV3Interface; always reverts
+     **/
+    function getRoundData(uint80 _roundId) override external view returns (
+        uint80 roundId,
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        revert NotImplemented();
+    }
+
+    /**
+     * @notice Price for the latest round
+     * @return roundId Round id from the underlying price feed
+     * @return answer Latest price for the asset (will always be a constant price)
+     * @return startedAt Timestamp when the round was started; passed on from underlying price feed
+     * @return updatedAt Timestamp when the round was last updated; passed on from underlying price feed
+     * @return answeredInRound Round id in which the answer was computed; passed on from underlying price feed
+     **/
+    function latestRoundData() external view returns (
+        uint80 roundId,
+        int256 answer,
+        uint256 startedAt,
+        uint256 updatedAt,
+        uint80 answeredInRound
+    ) {
+        return (0, CONSTANT_PRICE, block.timestamp, block.timestamp, 0);
+    }
+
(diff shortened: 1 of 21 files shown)


contracts/WstETHPriceFeed.sol
```diff
@@ -2,35 +2,36 @@
...
 
+    /// @notice Version of the price feed
     uint public constant override version = 1;
 
-    /// @notice Scale for returned prices
+    /// @notice Description of the price feed
+    string public constant override description = "Custom price feed for wstETH / ETH";
+
+    /// @notice Number of decimals for returned prices
     uint8 public override decimals = 8;
```
kevincheng96, line 18: Rather than hardcoding this, we could also make this an immutable set via the constructor.

contracts/ScalingPriceFeed.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    function signed256(uint256 n) internal pure returns (int256) {
+        if (n > uint256(type(int256).max)) revert InvalidInt256();
+        return int256(n);
+    }
+
+    function scalePrice(int256 price) internal view returns (int256) {
+        int256 scaledPrice;
+        if (shouldUpscale) {
+            scaledPrice = price * rescaleFactor;
```
scott-silver, line 94: Are we ever going to upscale prices?

It seems like the only use case for this contract is truncating an 18 decimal price feed to 8 decimals.

Would it be simpler/more efficient to have `scalePrice` always return `price / rescaleFactor`?

contracts/ConstantPriceFeed.sol
```diff
@@ -0,0 +1,61 @@
...
+    /// @notice The constant price
+    int private immutable CONSTANT_PRICE;
+
+    /**
+     * @notice Construct a new scaling price feed
+     * @param decimals_ The number of decimals for the returned prices
+     **/
+    constructor(uint8 decimals_) {
+        decimals = decimals_;
+        CONSTANT_PRICE = signed256(10 ** decimals_);
```
jflatow, line 29: Kinda wonder if we should just pass this in, or a factor... or otherwise maybe rename `UnitPriceFeed`, but arbitrary constant seems more flexible

contracts/ConstantPriceFeed.sol
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * @notice Unimplemented function required to fulfill AggregatorV3Interface; always reverts
+     **/
+    function getRoundData(uint80 _roundId) override external pure returns (uint80, int256, uint256, uint256, uint80) {
+        revert NotImplemented();
```
jflatow, line 36: I wonder if it makes more sense to just define our own *actual* interface at this point, otherwise this begs the question a little of the purpose of the interface since we are basically just tricking the compiler

contracts/WstETHPriceFeed.sol
jflatow: Probably, maybe, since we are passing it in the others?

contracts/ConstantPriceFeed.sol
kevincheng96: Yeah good point, passing in an arbitrary constant sounds good. I'll make that change.

contracts/ScalingPriceFeed.sol
kevincheng96: Yeah maybe simpler is better. But my goal was to make this flexible *in case* there was ever a need to scale up. There's not much extra gas costs to it, so it just depends on how much we value the extra flexibility. 

Personally, I rather have the flexibility (hardcoding price feed decimals has made the WETH deployment trickier, for example), but not a super strong opinion if we think it's not worth it.

contracts/ConstantPriceFeed.sol
kevincheng96: Do you mean a separate interface only for this price feed? That works, but `Comet` code does use the `AggregatorV3Interface`. Either way, calling this specific function selector on this price feed would revert, but at least this price feed currently conforms to the interface that `Comet` expects.

jflatow: I mean change the interface Comet actually uses (which is a subset of aggregator), and then it wouldn't depend on vendor anymore (solidity change, but shouldn't affect the bytecode), and we could reference that here

contracts/WstETHPriceFeed.sol
kevincheng96: Done

contracts/ConstantPriceFeed.sol
kevincheng96: Done

contracts/ConstantPriceFeed.sol
kevincheng96: I see, that makes sense. We do lose the ability to verify it against the vendor repo using Vendoza though, if you are okay with that.

jflatow:
The description of this PR is out of date now, right? This is now a PR to do truncated USD prices, not ETH

contracts/ConstantPriceFeed.sol
jflatow: What are we actually verifying in this case?

kevincheng96: We are verifying the `AggregatorV3Interface` we use in this repo against the interface defined in the Chainlink repo. More of an extra sanity check.

kevincheng96:
> The description of this PR is out of date now, right? This is now a PR to do truncated USD prices, not ETH

The description should still be accurate since this PR is for ETH-denominated prices, not USD. The `ScalingPriceFeed` can be used for any price denomination, but the changes to `WstETHPriceFeed` make it denominated in ETH rather than USD.

contracts/ConstantPriceFeed.sol
jflatow: But does that add anything? We use the interface we use right, and we sim against the actual contracts, whether or not its defined in Chainlink repo doesn't even give us extra sanity, right?

kevincheng96: Practically, it offers very little extra sanity as you said because of scenarios. But scenarios also don't have 100% coverage of all edge-cases, so extra sanity checks isn't a bad thing.

That being said, I'm totally okay with not having this interface be verified by Vendoza (I wasn't defending it, rather just pointing out we lose its checks). 

However, I do want to discuss whether or not changing `Comet`'s `AggregatorV3Interface` is a worthwhile change. The `ConstantPriceFeed` is really the only price feed that lacks a functioning `getRoundData`. All other price feeds that Comet interacts with, including the other 2 custom feeds in this PR, will have a functioning `getRoundData`. We could make the `ConstantPriceFeed` use a different interface, but I'm not sure it's worthwhile to change the interface that `Comet` uses just because of this one price feed for the WETH deployment. IMO, it seems more correct to be using the full canonical Chainlink interface.

jflatow: But we never *use* `getRoundData` even if it's there, right? It's weird to me that we would have an interface that has functions that we don't rely on, and then revert in impl bc we know we don't actually use the fn for anything. I don't feel strongly but maybe should discuss live, I don't see what's more correct about it

kevincheng96: Synced offline. I created a separate price feed interface without `getRoundData` to be used by `Comet` and extended by these wrapper price feed contracts. This new interface prevents us from accidentally calling an unsupported function (e.g. `getRoundData` for `ConstantPriceFeed`) in any future changes.

kevincheng96:
Updated this PR with a new `IPriceFeed` interface to be used by the wrapper price feeds and `Comet`. Discussion can be found here: https://github.com/compound-finance/comet/pull/625#discussion_r1029839340

deployments/mainnet/weth/deploy.ts
```diff
@@ -10,8 +10,29 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
...
+      exp(1, 8)                                      // constantPrice
+    ]
+  );
+
+  // Deploy scaling price feed for cbETH
+  const cbETHScalingPriceFeed = await deploymentManager.deploy(
+    'cbETH:priceFeed',
+    'ScalingPriceFeed.sol',
+    [
+      '0x86392dC19c0b719886221c78AB11eb8Cf5c52812', // XXX using stETH / ETH price feed until cbETH / ETH is deployed
```
jflatow, line 34: We should have cbETH / ETH now?

kevincheng96: Good catch, just updated it.

jflatow (approved):
💪

PR #626, opened 2022-11-21
Move priceFeedDecimals to Comet constructor

kevincheng96:
This PR modifies the `Comet` and `CometExt` contracts to allow `priceFeedDecimals` to be set in the constructor. This is to support Comet deployments where the price feeds may have other decimals amounts, such as 18 for X/ETH price feeds.

The contract changes are as follows:
- Add `priceFeedDecimals` as a new field in `Configuration`, which is passed to the `Comet` constructor and is also stored in the `Configurator`
- Remove any `priceScale` dependencies in the `CometExt` (by removing it from `CometCore`) since `priceScale` is now directly dependent on what's passed to the `Comet` constructor
- Add method and event for setting `priceFeedDecimals` in the `Configurator`

I also moved some of the price feed changes over from #625 (e.g. constant price feed, modified wstETH price feed) to get the full deployment and scenarios working.

### DIFF
--- a/deployments/mainnet/weth/configuration.json
+++ b/deployments/mainnet/weth/configuration.json
@@ -3,12 +3,12 @@
   "symbol": "cWETHv3",
   "baseToken": "WETH",
   "baseTokenAddress": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
-  "baseTokenPriceFeed": "0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419",
   "borrowMin": "100e6",
   "governor": "0x6d903f6003cca6255d85cca4d3b5e5146dc33925",
   "pauseGuardian": "0xbbf3f1421d886e9b2c5d716b5192ac998af2012c",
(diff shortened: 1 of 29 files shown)


deployments/mainnet/weth/configuration.json
```diff
@@ -28,7 +28,7 @@
   "assets": {
     "cbETH": {
       "address": "0xBe9895146f7AF43049ca1c1AE358B0541Ea49704",
-      "priceFeed": "0x67eF3CAF8BeB93149F48e8d20920BEC9b4320510",
+      "priceFeed": "0x86392dC19c0b719886221c78AB11eb8Cf5c52812",
```
kevincheng96, line 31: Note: This is not the correct price feed. I'm just using a placeholder one with 18 decimals until cbETH/ETH is deployed.

jflatow (commented):
Ok so we discussed and we intend to go with #625

kevincheng96:
Yep, going to close this one.

PR #627, opened 2022-11-21
sync hasPurchasableCollateral definitions

scott-silver:
This is a fix for a potential issue with the way that the TypeScript code and the Liquidator.sol code determine whether Comet has a sufficient amount of collateral for sale.

Approximately once per block, the TypeScript code calls `hasPurchaseableCollateral` to see if there is any collateral for sale. It returns true if there is any asset whose collateral reserve value (in USDC) is greater than $100. If that's true, then we fire off a transaction to the Liquidator contract (with an empty list of addresses to absorb).

The liquidator contract then tries to calculate how much of each asset it should buy from the protocol. If the value is less than `liquidationThreshold`, the contract will decide to buy 0. The transaction goes through but it does nothing and costs some amount of gas.

So if there is $500 worth of an asset for sale, `hasPurchaseableCollateral` will return true. but if the `liquidationThreshold` is $1,000, then when you go to buy that collateral, the Liquidator will say to buy $0, and just continue.

That transaction will happen nearly every block, until you've drained all the ETH from the wallet that the Liquidation Bot is using. At which point, you'll start getting errors when you try to initiate a transaction.

This patch exposes a `hasPurchasableCollateral` function on the Liquidator.sol contract. This means that both the TypeScript code and the Solidity code will have the same definition of whether there is collateral worth purchasing (and they will stay in sync even when `liquidationThreshold` is altered in the future).

### DIFF
--- a/contracts/liquidator/Liquidator.sol
+++ b/contracts/liquidator/Liquidator.sol
@@ -408,6 +408,33 @@ contract Liquidator is IUniswapV3FlashCallback, PeripheryImmutableState, Periphe
         return a <= b ? a : b;
     }
 
+    function hasPurchasableCollateral() external view returns (bool) {
+        uint8 numAssets = comet.numAssets();
+        for (uint8 i = 0; i < numAssets; i++) {
+            address asset = comet.getAssetInfo(i).asset;
+            uint256 assetBaseAmount = baseAmountForCollateral(asset);
+
+            if (assetBaseAmount > 0) {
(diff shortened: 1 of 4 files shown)


jflatow (approved):
Nice improvement 💪

kevincheng96 (approved):
Great work on this 👍

scott-silver:
made unnecessary by #642

PR #632, opened 2022-12-07
[draft] DEX aggregator liquidation bot

scott-silver:
_Extremely_ rough version of a 1inch-enabled liquidation bot. Lots of changes to make still, but want to get some visibility on the decisions I'm making and thoughts on the remaining issues.


contracts/liquidator/LiquidatorV2.sol
```diff
@@ -0,0 +1,235 @@
...
+        address[] calldata liquidatableAccounts,
+        address[] calldata assets,
+        // uint256[] calldata assetBaseAmounts,
+        address[] calldata swapTargets,
+        bytes[] calldata swapTransactions
+    ) external {
+        comet.absorb(address(this), liquidatableAccounts);
+        emit Absorb(msg.sender, liquidatableAccounts);
+
+        address poolToken0 = 0x6B175474E89094C44Da98b954EedeAC495271d0F; // XXX DAI
```
scott-silver, line 127: - [ ] remove hard-coded DAI reference
- [ ] build pool on contract construction
- [ ] potentially allow the pool to be re-set

contracts/liquidator/LiquidatorV2.sol
```diff
@@ -0,0 +1,235 @@
...
+    event BuyAndSwap(
+        address indexed tokenIn,
+        address indexed tokenOut,
+        uint256 baseAmountPaid,
+        uint256 assetBalance,
+        uint256 amountOut
+    );
+
+    /// @notice Compound Comet protocol
+    CometInterface public immutable comet;
```
scott-silver, line 37: alternatively, we could take in a comet instance as an argument, similar to the way Bulker does

contracts/liquidator/LiquidatorV2.sol
```diff
@@ -0,0 +1,235 @@
...
+
+        address[] swapTargets;
+        bytes[] swapTransactions;
+    }
+
+    // XXX name?
+    function absorbAndArbitrage(
+        address[] calldata liquidatableAccounts,
+        address[] calldata assets,
+        // uint256[] calldata assetBaseAmounts,
```
scott-silver, line 120: - [ ] delete

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,284 @@
...
+        // expect the balance of the recipient to have increased
+        assertGt(ERC20(usdc).balanceOf(liquidator_eoa), initialRecipientBalance);
+
+        // expect the protocol reserves to have increased
+        assertGt(CometInterface(comet).getReserves(), initialReserves);
+
+        // XXX make sure that you're making > 1% of the value of the swap
+    }
+
+    function testLargeWbtcSwap() public {
```
scott-silver, line 164: these five `textLarge${asset}Swap` tests are fairly repetitive. originally, i wrote this as a single test that looped over each asset and then performed the same actions/checked the same expecatations.

the advantage of having them as separate tests is that it's easier to get a readable stack trace for failures when they are separate. when you have all 5 in the same test and there is a failure, the stack trace is enormous and filled with a number of successful operations.

still undecided on which is preferable.

scenario/LiquidationBotV2Scenario.ts
```diff
@@ -0,0 +1,116 @@
+import { scenario } from './context/CometContext';
```
scott-silver, line 1: - [ ] delete file

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -20,54 +22,264 @@ export interface Asset {
...
-    tokenAddress: '0x6B175474E89094C44Da98b954EedeAC495271d0F',
-    poolFee: 100
-  },
-  'goerli': {
-    // WETH pool
-    tokenAddress: '0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6',
-    poolFee: 3000
-  }
-};
+const protocols = [
```
scott-silver, line 25: - [ ] delete

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -20,54 +22,264 @@ export interface Asset {
...
+
+  console.log({
+    addresses,
+    collateralReserves,
+    collateralReservesInBase
+  });
+
+  let protocols = [];
+
+  if (excludeSources.length > 0) {
```
scott-silver, line 153: - [ ] delete

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -20,54 +22,264 @@ export interface Asset {
...
+  // console.log(tx);
+  // const trace = await world.deploymentManager.hre.network.provider.send("debug_traceTransaction", [
+  //   tx.hash
+  // ]);
+  // console.log(`trace:`);
+  // console.log(trace);
+
+
+  // try {
+  //   googleCloudLog(LogSeverity.INFO, `Attempting to liquidate ${targetAddresses} via ${liquidator.address}`);
```
scott-silver, line 253: - [ ] restore

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,284 @@
...
+        // expect that there is only dust (< 1 unit) left of the asset
+        assertLt(CometInterface(comet).getCollateralReserves(asset), 10 ** ERC20(asset).decimals());
+
+        // expect the balance of the recipient to have increased
+        assertGt(ERC20(usdc).balanceOf(liquidator_eoa), initialRecipientBalance);
+
+        // expect the protocol reserves to have increased
+        assertGt(CometInterface(comet).getReserves(), initialReserves);
+
+        // XXX make sure that you're making > 1% of the value of the swap
```
scott-silver, line 161: unsure how to retrieve this information in Forge. is there a way to read values from an emitted event and then use them in an assertion? as far as I can tell, you can only confirm the values of an expected event, not read values from an emitted event

forge/test/LiquidationBotV2.t.sol
kevincheng96: For now, we could keep them separate for easier debugging. But we can always combine them into one test if maintenance ever becomes a pain in the future.

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,284 @@
...
+        // assertLt(CometInterface(comet).getCollateralReserves(asset), 10 ** ERC20(asset).decimals());
+
+        // expect the balance of the recipient to have increased
+        assertGt(ERC20(usdc).balanceOf(liquidator_eoa), initialRecipientBalance);
+
+        // expect the protocol reserves to have increased
+        assertGt(CometInterface(comet).getReserves(), initialReserves);
+    }
+
+    // XXX test actually liquidating an underwater account
```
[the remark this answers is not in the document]
kevincheng96: Not an answer to your original question, but I Googled that error and it seems to be coming from Synthetix. Is Curve routing through Synthetix (or maybe the other way around)?

contracts/liquidator/LiquidatorV2.sol
kevincheng96: I do like the idea of having `Comet` be part of the calldata. Shouldn't be much gassier but offers more flexibility.

forge/test/LiquidationBotV2.t.sol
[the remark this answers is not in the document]
scott-silver: My bad, I meant Synthetix not Curve. It comes from this contract:

https://github.com/Synthetixio/synthetix/blob/v2.80.0-alpha/contracts/Exchanger.sol#L527

contracts/liquidator/LiquidatorV2.sol
```diff
@@ -0,0 +1,235 @@
...
+        recipient = _recipient;
+    }
+
+    function balanceOfAsset(address asset) internal returns (uint256, uint256) {
+        uint256 collateralBalance = comet.getCollateralReserves(asset);
+
+        uint256 baseScale = comet.baseScale();
+
+        uint256 quotePrice = comet.quoteCollateral(asset, QUOTE_PRICE_SCALE * baseScale);
+        uint256 collateralBalanceInBase = baseScale * QUOTE_PRICE_SCALE * collateralBalance / quotePrice;
```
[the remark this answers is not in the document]
kevincheng96: I'll need to spend more time looking at this, but what's the issue with not buying the exact amount of collateral if it's strictly <= the protocol's collateral reserves? The txn should still succeed and leave dust amounts in Comet.

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,280 @@
...
+    address public constant weth9 = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+
+    // whales
+    address public constant comp_whale = 0x2775b1c75658Be0F640272CCb8c72ac986009e38;
+    address public constant link_whale = 0xfB682b0dE4e0093835EA21cfABb5449cA9ac9e5e;
+    address public constant uni_whale = 0x1a9C8182C09F50C8318d769245beA52c32BE35BC;
+    address public constant weth_whale = 0x2F0b23f53734252Bda2277357e97e1517d6B042A;
+
+    // wallets
+    address public constant liquidator_eoa = 0x5a13D329A193ca3B1fE2d7B459097EdDba14C28F;
```
kevincheng96, line 38: nit: all caps for constants?

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,280 @@
...
+    function setUp() public {
+        liquidator = new LiquidatorV2(
+            CometInterface(comet),
+            address(uniswap_v3_factory),
+            address(weth9),
+            address(liquidator_eoa)
+        );
+
+        // contracts
+        vm.label(aggregation_router_v5, "AggregationRouterV5");
```
kevincheng96, line 49: TIL, this is cool

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,280 @@
...
+        inputs[5] = vm.toString(fromTokenAddress);
+        inputs[6] = vm.toString(toTokenAddress);
+        inputs[7] = vm.toString(swapAmount);
+
+        string memory responseJson = string(vm.ffi(inputs));
+
+        return (
+            abi.decode(vm.parseJson(responseJson, ".target"), (address)),
+            abi.decode(vm.parseJson(responseJson, ".tx"), (bytes))
+        );
```
kevincheng96, line 92: Also very cool

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    function swap(address asset) public {
+        address[] memory liquidatableAccounts;
+
+        // XXX not a static call; will actually absorb the liquidatableAccounts
```
kevincheng96, line 98: If you want to do a static call, you can wrap this around a `snapshot()` and `revert()`. https://github.com/foundry-rs/foundry/issues/2463#issuecomment-1194281965

forge/test/LiquidationBotV2.t.sol
kevincheng96: You should be able to use `recordLogs` and `getRecordedLogs` https://book.getfoundry.sh/cheatcodes/record-logs

forge/test/LiquidationBotV2.t.sol
```diff
@@ -0,0 +1,280 @@
...
+        // expect that there is only dust (< 1 unit) left of the asset
+        assertLt(CometInterface(comet).getCollateralReserves(asset), 10 ** ERC20(asset).decimals());
+
+        // expect the balance of the recipient to have increased
+        assertGt(ERC20(usdc).balanceOf(liquidator_eoa), initialRecipientBalance);
+
+        // expect the protocol reserves to have increased
+        assertGt(CometInterface(comet).getReserves(), initialReserves);
+
+        // XXX make sure that you're making > 1% of the value of the swap
```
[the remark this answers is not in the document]
scott-silver: no, it's just an approximation of how much you should be making. if the protocol is selling at a 2.5% discount, and you have to pay some fee for the flash loan and there is some slippage in the swap, you should end up receiving ~1-2% of the value of the collateral

contracts/liquidator/LiquidatorV2.sol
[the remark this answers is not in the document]
scott-silver: it's fine to leave a little dust in the protocol. the issue is for the size of the swap that you request from 1inch.

if you request a swap for the entirety of the collateral (say 120e18 WBTC), you'll end up with a transfer error since you will have bought an amount that is every so slightly less than that amount (e.g. 119.999999999e18 WTBC).

basically, you need to make sure that you're not requesting a swap through 1inch for an amount that you don't actually have.

.github/workflows/run-forge-tests.yaml
```diff
@@ -20,8 +25,11 @@ jobs:
       - name: Install dependencies
         run: forge install
 
+      - name: Install packages
+        run: yarn install --non-interactive --frozen-lockfile && yarn build
+
       - name: Run tests
-        run: forge test -vvv --via-ir
+        run: forge test -vvv --via-ir --ffi --fork-url https://mainnet.infura.io/v3/$INFURA_KEY
```
scott-silver, line 32: I'm still learning the Forge APIs, but I think it might be possible to specify this forking at the test level, instead of having this be a global setting

forge/test/LiquidationBotV2.t.sol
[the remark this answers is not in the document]
cwang25: How about prank as chainlink and decrease the price of the assets in it's price feed contract?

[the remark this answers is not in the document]
scott-silver: oh that's an interesting idea

.github/workflows/run-forge-tests.yaml
scott-silver: updated; we now fork during test setup instead

forge/test/LiquidationBotV2.t.sol
scott-silver: updated

contracts/liquidator/LiquidatorV2.sol
scott-silver: updated to take Comet as an argument

contracts/liquidator/LiquidatorV2.sol
```diff
@@ -0,0 +1,274 @@
...
+
+    function uniswapV3FlashCallback(
+        uint256 fee0,
+        uint256 fee1,
+        bytes calldata data
+    ) external override {
+        FlashCallbackData memory flashCallbackData = abi.decode(data, (FlashCallbackData));
+        CallbackValidation.verifyCallback(factory, flashCallbackData.poolKey);
+
+        TransferHelper.safeApprove(CometInterface(flashCallbackData.comet).baseToken(), address(flashCallbackData.comet), flashCallbackData.flashLoanAmount);
```
scott-silver, line 218: use `baseToken` from line 220

PR #636, opened 2022-12-13
Deploy cWETHv3 to Goerli

kevincheng96:
**Status:**

The deploy script is now working fully and uses the existing `Configurator` and `Rewards` contracts. **But**, it skips over the admin-only actions that require a governance proposal on purpose, so we still need to create a migration script to create a proposal for the 3 functions: 

1. Setting the factory address in Configurator
2. Setting the configuration in Configurator and upgrading Comet to the new implementation
3. Setting the reward token config in CometRewards

In terms of scenarios, surprisingly a majority of the tests pass. The two types of failing scenarios are:

1. Divide by 0 errors because all the storage values in the newly-deployed Comet are 0
2. Any scenarios using the `upgrade` constraint will fail because the `ModernConstraint` will re-deploy `Comet` using the existing configuration stored in `Configurator`, which are currently 0 until the proposal to set configuration is made and executed


src/deploy/Network.ts
```diff
@@ -146,10 +146,12 @@ export async function deployNetworkComet(
     maybeForce(deploySpec.cometMain)
   );
 
+  const existingCometImpl = (await deploymentManager.contracts()).get('comet:implementation');
+  const tempCometImpl = existingCometImpl ? existingCometImpl.address : cometFactory.address;
   const cometProxy = await deploymentManager.deploy(
     'comet',
     'vendor/proxy/transparent/TransparentUpgradeableProxy.sol',
-    [cometFactory.address, cometAdmin.address, []], // NB: temporary implementation contract
+    [tempCometImpl, cometAdmin.address, []], // NB: temporary implementation contract
```
jflatow, line 154: I think this breaks initialization though

src/deploy/Network.ts
```diff
@@ -215,25 +217,29 @@ export async function deployNetworkComet(
...
+    async () => isInitialMarket && !sameAddress(await configurator.factory(comet.address), cometFactory.address),
     async () => {
       trace(`Setting factory in Configurator to ${cometFactory.address}`);
       trace(await wait(configurator.connect(admin).setFactory(comet.address, cometFactory.address)));
     }
   );
 
   await deploymentManager.idempotent(
-    async () => isFirstDeploy || deploySpec.all || deploySpec.cometMain || deploySpec.cometExt,
+    async () => isInitialMarket && isFirstDeploy || deploySpec.all || deploySpec.cometMain || deploySpec.cometExt,
```
jflatow, line 242: I think we'd need to `isInitialMarket && (....)`

kevincheng96:
Closing in favor of #646, which is based on top of the latest changes in #587. These latest changes handle some of the issues discovered in this PR, mainly the ability to re-use an existing deployment's `Configurator` and `Rewards` contracts.

PR #638, opened 2022-12-22
Do not source from actors

jflatow:
This can lead to multiple constraints violating each other, which is causing scen failures. Not sure when/why we needed this but we should find another way if sourcing without it is not enough.

### DIFF
--- a/scenario/LiquidationBotScenario.ts
+++ b/scenario/LiquidationBotScenario.ts
@@ -315,7 +315,7 @@ scenario(
     },
     cometBalances: {
       albert: {
-        $asset0: ' == 5',
+        $asset0: ' == 10',
       },
       betty: { $base: 1000 },
     },
@@ -443,7 +443,7 @@ scenario(
     },
     cometBalances: {
       albert: {
-        $asset0: ' == 5',
+        $asset0: ' == 10',
       },
       betty: { $base: 1000 },
     },
--- a/scenario/TransferScenario.ts
+++ b/scenario/TransferScenario.ts
@@ -1,6 +1,6 @@
 import { CometContext, scenario } from './context/CometContext';
 import { expect } from 'chai';
-import { expectApproximately, expectRevertCustom, getExpectedBaseBalance, getInterest, isTriviallySourceable, isValidAssetIndex, MAX_ASSETS } from './utils';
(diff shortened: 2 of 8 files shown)


scenario/context/CometContext.ts
```diff
@@ -210,7 +210,7 @@ export class CometContext {
     let amountRemaining = BigInt(amount);
 
     // Try to steal from a known whale or actor
-    for (const [name, actor] of whales.concat(Object.entries(this.actors))) {
+    for (const [name, actor] of whales) {
```
scott-silver, line 213: nit: update comment: `// Try to steal from a known whale`

scott-silver (approved):

scenario/context/CometContext.ts
jflatow: Yeah, actually cleaning this up even further now

scenario/LiquidationBotScenario.ts
```diff
@@ -443,7 +443,7 @@ scenario(
     },
     cometBalances: {
       albert: {
-        $asset0: ' == 5',
+        $asset0: ' == 10',
```
jflatow, line 446: Due to prices falling, we were apparently not borrowing enough anymore (below min)

scenario/TransferScenario.ts
```diff
@@ -92,8 +92,8 @@ scenario(
     const baseTransferred = getExpectedBaseBalance(50n * scale, baseIndexScale, baseSupplyIndex);
     const baseOfTransferrer = getExpectedBaseBalance(baseSupplied - toTransfer, baseIndexScale, baseSupplyIndex);
 
-    expect(await comet.balanceOf(albert.address)).to.be.equal(baseOfTransferrer);
-    expect(await comet.balanceOf(betty.address)).to.be.equal(baseTransferred);
+    expectBase(await albert.getCometBaseBalance(), baseOfTransferrer);
+    expectBase(await betty.getCometBaseBalance(), baseTransferred);
```
jflatow, line 96: Unrelated but intermittently failing while testing this

scott-silver: `comet.balanceOf(actor)` and `actor.getCometBaseBalance()` aren't identical. looks like `getCometBaseBalance` subtracts the borrow balance.

was the original test incorrect?

scott-silver: oh, I see `expectBase` does an approximate match, not an exact one

scott-silver:
re-approving. LGTM ✅

PR #643, opened 2022-12-30
Convert from scientific notation to bigint in configuration

kevincheng96:
This PR adds a new `ScientificNotation` string type to `NetworkConfiguration` that is required for certain fields (e.g. `supplyCap`, `baseMinForRewards`). The configuration parser will safely convert these values into bigints without losing precision; with one caveat, scientific notations that use decimals in the coefficient still run the risk of precision loss because those are converted directly to javascript numbers.

### DIFF
--- a/deployments/fuji/usdc/configuration.json
+++ b/deployments/fuji/usdc/configuration.json
@@ -29,7 +29,7 @@
       "borrowCF": 0.7,
       "liquidateCF": 0.75,
       "liquidationFactor": 0.93,
-      "supplyCap": "35000"
+      "supplyCap": "35000e8"
     },
     "WAVAX": {
       "priceFeed": "0x5498BB86BC934c8D34FDA08E81D444153d0D06aD",
@@ -37,7 +37,7 @@
       "borrowCF": 0.82,
       "liquidateCF": 0.85,
       "liquidationFactor": 0.93,
-      "supplyCap": "1000000"
+      "supplyCap": "1000000e18"
     }
   }
 }
--- a/src/deploy/NetworkConfiguration.ts
+++ b/src/deploy/NetworkConfiguration.ts
@@ -36,17 +36,18 @@ function stringToBigInt(x: ScientificNotation) {
   if (typeof x !== 'string') {
     throw new Error(`expected argument to be string, got ${x}`);
   }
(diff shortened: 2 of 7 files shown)


deployments/fuji/usdc/configuration.json
```diff
@@ -29,15 +29,15 @@
...
+      "supplyCap": "35000e8"
     },
     "WAVAX": {
       "priceFeed": "0x5498BB86BC934c8D34FDA08E81D444153d0D06aD",
       "decimals": "18",
       "borrowCF": 0.82,
       "liquidateCF": 0.85,
       "liquidationFactor": 0.93,
-      "supplyCap": "1000000"
+      "supplyCap": "1000000e18"
```
jflatow, line 40: We're sure we prefer this being redundant? In that case, maybe worth going so far as removing `_` too so can write `1_000_000e18`

kevincheng96: Good point, added support for underscores.

kevincheng96: Honestly, I can be swayed both ways. I'm starting to appreciate the simplicity of the other approach. We just have to make sure to apply the right scales to some other values like `baseMinForRewards`

src/deploy/NetworkConfiguration.ts
```diff
@@ -36,17 +36,18 @@ function stringToBigInt(x: ScientificNotation) {
   if (typeof x !== 'string') {
     throw new Error(`expected argument to be string, got ${x}`);
   }
-  if (!x.match(/^[0-9]+([.][0-9]+)?e[0-9]+$/)) {
+  let sanitizedInput = x.replace(/_/g, '');
```
jflatow, line 39: `const`?

kevincheng96: Good catch. Done.

jflatow (approved):
🔥

deployments/fuji/usdc/configuration.json
jflatow: Yeah, up to you if you want to merge, I often think less is more though

kevincheng96: I'll merge this since it reduces the amount of configuration diffs we have against `main`. Functionally, both approaches are the same so we could revisit in a separate PR that focuses on refactoring the configuration.

PR #644, opened 2022-12-30
Spider USDC deployment before calling `fromDep`

kevincheng96:
We need to spider the `USDC` deployment before calling `fromDep` on it so all the aliases are populated in the cache. Otherwise, `fromDep` may try to import an `undefined` contract. 

There are many approaches to this and this PR takes the simplest approach of spidering in the deploy script. Another approach could be to handle the spidering directly in `fromDep`.

### DIFF
--- a/deployments/mainnet/weth/deploy.ts
+++ b/deployments/mainnet/weth/deploy.ts
@@ -37,6 +37,16 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
   );
 
   // Import shared contracts from cUSDCv3
+  // Note: We need to first spider the contracts on the cUSDC base. Consider moving this spider logic into `fromDep`.
+  const usdcDm = new DeploymentManager(
+    'mainnet',
+    'usdc',
+    deploymentManager.hre,
+    {
+      writeCacheToDisk: true,
+    }
+  );
+  await usdcDm.spider();
   const cometAdmin = await deploymentManager.fromDep('cometAdmin', 'mainnet', 'usdc');
   const cometFactory = await deploymentManager.fromDep('cometFactory', 'mainnet', 'usdc');
   const $configuratorImpl = await deploymentManager.fromDep('configurator:implementation', 'mainnet', 'usdc');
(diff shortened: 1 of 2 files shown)


plugins/deployment_manager/DeploymentManager.ts
```diff
@@ -215,7 +215,19 @@ export class DeploymentManager {
...
+          network,
+          deployment,
+          this.hre,
+          {
+            writeCacheToDisk: true,
+          }
+        );
+        await dm.spider();
+        address = await this.readAlias(network, deployment, otherAlias);
+      }
```
jflatow, line 230: As is we call this multiple times, so I think we'd at least want caching with this approach, but kinda like calling it from the script before

kevincheng96: So you prefer spidering from the deploy script? I'm on board with that too.

This approach does cache already right? `writeCacheToDisk` is `true`, so any future spiders won't make network calls to etherscan.

jflatow: Yea we cache network requests, but still does a bunch of work / noise on each call. Perhaps not so bad if limited to a handful of calls in deploy, but also imagine it may pop up in migration scripts

kevincheng96: Yeah definitely could be optimized further. I got rid of this change for now and moved the spidering into the deploy script.

deployments/mainnet/weth/deploy.ts
```diff
@@ -37,6 +37,16 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
...
+  // Note: We need to first spider the contracts on the cUSDC base. Consider moving this spider logic into `fromDep`.
+  const usdcDm = new DeploymentManager(
+    'mainnet',
+    'usdc',
+    deploymentManager.hre,
+    {
+      writeCacheToDisk: true,
+    }
+  );
+  await usdcDm.spider();
```
jflatow, line 49: 💪

jflatow (approved):

PR #645, opened 2023-01-03
Add unit tests to sanity check tracking index scale bounds for WETH

kevincheng96:
Looks like `1e15` is a reasonable `trackingIndexScale` for the `WETH` deploy. We could consider `1e12` to increase the upper bound (less chance of overflow), but it's nice to keep the index scale the same as that of the `USDC` deploy.

A `baseMinForRewards` of `100e18` seems to be good enough, but I recommend `1000e18` just to be safe.


jflatow (approved):
Beautifully done! 👏

PR #650, opened 2023-01-05
mainnet-eth-live patch

scott-silver:


scenario/LiquidationBotScenario.ts
```diff
@@ -64,7 +64,7 @@ for (let i = 0; i < MAX_ASSETS; i++) {
     {
       filter: async (ctx) => ctx.world.base.network === 'mainnet' && await isValidAssetIndex(ctx, i),
       tokenBalances: {
-        $comet: { $base: 10000 },
+        $comet: { $base: 3000 },
```
scott-silver, line 67: we're getting `NotForSale()` errors in these scenarios, since the amount of `$base` specified previously (10K) is greater than the new `targetReserves` value (5k).

deployments/mainnet/weth/migrations/1666906000_configurate_seed_reserves_and_rewards.ts
```diff
@@ -15,7 +15,10 @@ export default migration('1666906000_configurate_seed_reserves_and_rewards', {
     const trace = deploymentManager.tracer();
     const ethers = deploymentManager.hre.ethers;
 
-    const comptrollerV2 = await deploymentManager.fromDep('comptrollerV2', 'mainnet', 'usdc');
+    const comptrollerV2Proxy = await deploymentManager.fromDep('comptrollerV2', 'mainnet', 'usdc');
+    const comptrollerV2Impl = await deploymentManager.fromDep('comptrollerV2:implementation', 'mainnet', 'usdc');
```
scott-silver, line 19: necessary for us to be able to call `populateTransaction` with `'_setCompSpeeds(address[],uint256[],uint256[])'` in the `proposal` function

deployments/mainnet/weth/migrations/1666906000_configurate_seed_reserves_and_rewards.ts
```diff
@@ -91,8 +94,11 @@ export default migration('1666906000_configurate_seed_reserves_and_rewards', {
 
 
   async verify(deploymentManager: DeploymentManager) {
+    const comptrollerV2Proxy = await deploymentManager.fromDep('comptrollerV2', 'mainnet', 'usdc');
+    const comptrollerV2Impl = await deploymentManager.fromDep('comptrollerV2:implementation', 'mainnet', 'usdc');
```
scott-silver, line 98: necessary for us to be able to call `comptrollerV2.compSupplySpeeds()` below

deployments/mainnet/weth/migrations/1666906000_configurate_seed_reserves_and_rewards.ts
```diff
@@ -15,7 +15,10 @@ export default migration('1666906000_configurate_seed_reserves_and_rewards', {
     const trace = deploymentManager.tracer();
     const ethers = deploymentManager.hre.ethers;
 
-    const comptrollerV2 = await deploymentManager.fromDep('comptrollerV2', 'mainnet', 'usdc');
+    const comptrollerV2Proxy = await deploymentManager.fromDep('comptrollerV2', 'mainnet', 'usdc');
+    const comptrollerV2Impl = await deploymentManager.fromDep('comptrollerV2:implementation', 'mainnet', 'usdc');
+    const comptrollerV2 = comptrollerV2Impl.attach(comptrollerV2Proxy.address);
```
jflatow, line 20: This doesn't really make sense though, we should have it and not just sometimes, right? Seems like there's some other bug 🤔

scott-silver: Which part doesn't make sense?

If you don't attach the impl contract to the proxy contract, then the contract instance won't have the impl contract functions (which we need when we call `.populateTransaction`)

jflatow: Then why doesn't `mainnet/usdc` need to do this? And also why is it working for me and 🪖 locally? https://github.com/compound-finance/comet/blob/main/deployments/relations.ts#L4-L8 spider should be crawling the impl

scott-silver:
closing for now; will investigate further

PR #651, opened 2023-01-06
rework fromDep to include proxy functions

scott-silver:


jflatow (commented):
This looks pretty good, will give it a run shortly and I think pull this and the other fromDep changes here into a PR to merge into the deploy branch

PR #653, opened 2023-01-06
bump target reserves

scott-silver:
need to ensure that the protocol is selling collateral in the LiquidationBotScenarios


jflatow (approved):
👍

PR #657, opened 2023-01-11
XXX DAI DSR RESTORE EXAMPLE WIP

jflatow:
This is an example script showing how to mechanically change (and test) a cDAI change to restore the DSR logic. This proposal needs to be further checked and analyzed for correctness.


jflatow:
Closing since #663 picks this up

PR #660, opened 2023-01-12
Some basic v2 scenarios

jflatow:
This change isn't specific to the DSR branch, but running against that migration looks like there might be some issues redeeming after minting to look into...

### DIFF
--- a/scenario/V2Scenario.ts
+++ b/scenario/V2Scenario.ts
@@ -1,0 +1,78 @@
+import { scenario } from './context/CometContext';
+import { exp } from '../test/helpers';
+import { expect } from 'chai';
+
+// note: meant for sanity checking v2 proposals, can normally be set to skip
+//  enable to check specific v2 actions with addresses hard-coded by network
+//  addresses to check and markets will need to be updated between runs
+
+scenario.skip(
+  'Compound v2 > allows a user to repay, borrow, repay cETH',
+  {},
+  async (_, context, world) => {
+    const dm = context.world.deploymentManager;
+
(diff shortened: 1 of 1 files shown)


kevincheng96:
Neat! Is the plan to merge this into main or keep it as a branch?

jflatow:
> Neat! Is the plan to merge this into main or keep it as a branch?

I think we can merge and store w/ tests skipped on `main`, but probably don't need to merge into this branch (thats why its a draft)?

scenario/V2Scenario.ts
```diff
@@ -0,0 +1,82 @@
...
+
+// note: meant for sanity checking v2 proposals, can normally be set to skip
+//  enable to check specific v2 actions with addresses hard-coded by network
+//  addresses to check and markets will need to be updated between runs
+
+scenario.skip(
+  'Compound v2 > allows a user to repay, borrow, repay cETH',
+  {},
+  async ({ actors }, context, world) => {
+    const { albert, betty } = actors; albert; betty;
```
scott-silver, line 13: just curious, what does appending `albert; better;` to this line do?

isn't this the same as just:

```
const { albert, betty } = actors;
```

jflatow: Whoops good catch!

scott-silver (approved):
🔧

kevincheng96 (approved):
A couple lint issues, but otherwise LGTM!

scenario/V2Scenario.ts
jflatow: Actually these were to quiet linter; but deleted the actors altogether now

PR #665, opened 2023-01-26
Address incorrectly setting `localTimelock`

jflatow:
Addresses issue M-01 on the Polygon cross-chain governance OZ audit:

`M-01 BaseBridgeReceiver can be rendered inoperable by incorrectly setting the localTimelock` - Ideally, we would like to keep the Timelock’s interface the same as the currently-deployed Timelock that is in use on mainnet. Extending the Timelock (by including the ERC165Checker, for example) would result in mainnet and L2 Timelocks with different interfaces.

The issue raised here is valid: setting the `localTimelock` value to an invalid address would lock the entire deployment.

Updating the `localTimelock` address is fortunately an extremely uncommon action; we may never need to do it. And we can accomplish the same net result by deploying a new BridgeReceiver and a new Timelock, and then updating Comet’s governor and timelock value (and the ProxyAdmin’s admin) to the newly-deployed Timelock.

We will delete the following functions from the BaseBridgeReceiver contract:

    acceptLocalTimelockAdmin()
    setLocalTimelock(address newTimelock)
    setGovTimelock(address newTimelock)

This means that the only place we can make this mistake is during initialization of the BaseBridgeReceiver contract. If we make an error at this point, we can scrap the entire deployment (since there will be no borrow/supply activity at this point).

(During initialization, we can make the check that the localTimelock’s admin is set to the BridgeReceiver).


kevincheng96:
Contract changes are ready. I'm working on a scenario to test the "deploy a new bridge receiver + local timelock to replace the old one" flow.

jflatow:
I can't approve bc I created the dummy but LGTM 💪

scott-silver (approved):
🙌

scenario/GovernanceScenario.ts
```diff
@@ -216,49 +217,176 @@ scenario('add new asset (mainnet-weth)',
...
+      [
+        [timelock.address, comet.address],
+        [0, 0],
+        ['setDelay(uint256)', 'pause(bool,bool,bool,bool,bool)'],
+        [setDelayCalldata, pauseCalldata]
+      ]
+    );
+
+    expect(await timelock.delay()).to.eq(currentTimelockDelay);
+    expect(currentTimelockDelay).to.not.eq(newTimelockDelay);
```
jflatow, line 240: 🔥

jflatow (commented):
Awesome work 💪 🏆 Reassuring to see the timelock replacement flow demonstrated

PR #668, opened 2023-01-27
Polygon collateral assets

scott-silver:
Updating the Polygon config to reflect our current proposed set of assets: USDC as base, with WMATIC, WETH and WBTC as collateral assets.

All configuration variables (collateral factor, liquidation factor, etc) are placeholder, pending guidance from Gauntlet

### DIFF
--- a/deployments/polygon/usdc/configuration.json
+++ b/deployments/polygon/usdc/configuration.json
@@ -24,6 +24,15 @@
     "baseMinForRewards": "1000000e6"
   },
   "assets": {
+    "WMATIC": {
+      "address": "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270",
+      "priceFeed": "0xAB594600376Ec9fD91F8e885dADF0CE036862dE0",
+      "decimals": "18",
+      "borrowCF": 0.825,
+      "liquidateCF": 0.895,
+      "liquidationFactor": 0.95,
+      "supplyCap": "500000e18"
+    },
     "WETH": {
       "address": "0x7ceb23fd6bc0add59e62ac25578270cff1b9f619",
       "priceFeed": "0xF9680D99D6C9589e2a93a78A04A279e509205945",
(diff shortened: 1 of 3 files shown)


deployments/polygon/usdc/configuration.json
```diff
@@ -24,6 +24,15 @@
     "baseMinForRewards": "1000000e6"
   },
   "assets": {
+    "WMATIC": {
+      "address": "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270",
+      "priceFeed": "0xAB594600376Ec9fD91F8e885dADF0CE036862dE0",
```
scott-silver, line 29: This is the `MATIC / USD` price feed:

https://polygonscan.com/address/0xAB594600376Ec9fD91F8e885dADF0CE036862dE0#readContract

deployments/polygon/usdc/configuration.json
```diff
@@ -24,6 +24,15 @@
...
+      "decimals": "18",
+      "borrowCF": 0.825,
+      "liquidateCF": 0.895,
+      "liquidationFactor": 0.95,
+      "supplyCap": "500000e18"
+    },
     "WETH": {
       "address": "0x7ceb23fd6bc0add59e62ac25578270cff1b9f619",
       "priceFeed": "0xF9680D99D6C9589e2a93a78A04A279e509205945",
```
scott-silver, line 38: WETH is using the `ETH / USD` price feed

deployments/polygon/usdc/configuration.json
```diff
@@ -34,31 +43,13 @@
       "supplyCap": "500000e18"
     },
     "WBTC": {
-      "address": "0xc907E116054Ad103354f2D350FD2514433D57F6f",
-      "priceFeed": "0xDE31F8bFBD8c84b5360CFACCa3539B938dd78ae6",
+      "address": "0x1bfd67037b42cf73acf2047067bd4f2c47d9bfd6",
+      "priceFeed": "0xc907E116054Ad103354f2D350FD2514433D57F6f",
```
scott-silver, line 47: This is the `BTC / USD` price feed, although Chainlink also provides a `WBTC / USD` price feed:

https://polygonscan.com/address/0xc907E116054Ad103354f2D350FD2514433D57F6f#readContract

kevincheng96: I think we should use the `WBTC` price feed since it's available.

kevincheng96 (approved):
Awesome!

jflatow (approved):
💪

deployments/polygon/usdc/configuration.json
scott-silver: sounds good. update to:

https://polygonscan.com/address/0xDE31F8bFBD8c84b5360CFACCa3539B938dd78ae6#readContract

scenario/BulkerScenario.ts
```diff
@@ -10,15 +10,16 @@ scenario(
     filter: async (ctx) => await isBulkerSupported(ctx) && !matchesDeployment(ctx, [{deployment: 'weth'}, {network: 'mumbai'}]),
     supplyCaps: {
       $asset0: 3000,
+      $asset2: 1,
```
scott-silver, line 13: This scenario assumes that the wrapped native token collateral asset has a supply cap of at least `.01` units.

In the Polygon deploy, that asset is `$asset2`; somewhat awkward to assume which asset is the wrapper

jflatow: Yeah hmm, should we try to add a marker or something (e.g. in NativeTokenConstraint)?

scott-silver: like something that would let you specify a `supplyCaps` value for `$nativeTokenWrapper`, and the constraint would find which asset is the wrapper?

jflatow: Yeah, maybe just `$wnative` is an alias?

PR #669, opened 2023-01-30
reduce LINK amounts in LiquidationBot scenarios

scott-silver:
Liquidity for LINK has become a bit tighter; swapping 250,000 LINK currently results in >4% slippage (which makes liquidating unprofitable):

image

So we're starting to see LiquidationBot scenario failures:

Screen Shot 2023-01-30 at 2 41 28 PM

Reducing the amount we attempt to liquidate in LiquidationBot scenarios.


jflatow (approved):
💪 🤔

PR #674, opened 2023-02-07
Add official markets ENS record to the Polygon proposal

jflatow:


deployments/polygon/usdc/migrations/1675200105_supply_caps_and_ens.ts
```diff
@@ -70,11 +74,34 @@ export default migration('1675200105_raise_polygon_supply_caps', {
...
+    const ENSResolver = await deploymentManager.existing('ENSResolver', ENSResolverAddress);
+    const subdomainHash = ethers.utils.namehash(ENSSubdomain);
+    const officialMarketsJSON = await ENSResolver.text(subdomainHash, ENSTextRecordKey);
+    const officialMarkets = JSON.parse(officialMarketsJSON);
+    officialMarkets[137] = [
+      {
+        baseSymbol: 'USDC',
+        cometAddress: comet.address,
+      }
+    ];
```
kevincheng96, line 87: 🆒

kevincheng96 (approved):
Looks good!

scenario/context/CometContext.ts
```diff
@@ -407,6 +407,7 @@ async function forkContext(c: CometContext, w: World): Promise<CometContext> {
 export const constraints = [
   new FilterConstraint(),
   new NativeTokenConstraint(),
+  new ProposalConstraint(), // XXX
```
jflatow, line 410: Note: remove this after ENS proposal passes, want to assume that proposal has run in the migration script

scott-silver (approved):
Looks good!

Does verify never get called anywhere in `tasks/deployment_manager/task.ts`? Is a migration's `verify` function only ever run in scenarios?

jflatow:
> Looks good!
> 
> Does verify never get called anywhere in `tasks/deployment_manager/task.ts`? Is a migration's `verify` function only ever run in scenarios?

Yeah it only runs in scenarios because it has to run after the proposal actually executes

jflatow:
Moved the full ENS setup into this proposal since the original proposal with just the initial setup failed to reach quorum.

kevincheng96 (approved):

PR #676, opened 2023-02-08
Fix mumbai deploy script and re-deploy

kevincheng96:
We were incorrectly cloning WETH from Mainnet rather than Polygon and Bulker was being initialized with the wrong arguments.


scott-silver (approved):
🙌

PR #677, opened 2023-02-09
Polygon liquidator updates

scott-silver:
This PR does the following:

- adds Forge tests for Polygon liquidation
- updates scenarios to run for Polygon
- updates the Liquidation Bot deploy script.

The existing Liquidation contract works for Polygon. Using Uniswap, we are currently able to swap

- 500 WETH (~$750K)
- 30 WBTC (~$600K)
- 500,000 WMATIC (~$600K)

Swap liquidity is the limiting factor for the Liquidation Bot (how much of a given asset can you swap before the price slippage exceeds the discount that the protocol has sold the collateral for, making the swap unprofitable).

But there is an additional challenge for Polygon: flash loan liquidity.

When taking out a flash loan, we have to find a Uniswap pool where 1) one of the paired tokens is the base token of the protocol, and 2) the other token is not one of the collateral assets.

(The second token cannot be a collateral asset because if you take out a flash loan from a pool and then later attempt to swap via that same pool, you will exit with an error from the Uniswap's contracts reentrancy guard.)

Finding a pool that fits this criteria is easy on mainnet (we currently use USDC/DAI), but harder on Polygon since there are fewer pools and they are smaller.

On Polygon, the largest `[TOKEN]/USDC` pools that do not involve WETH, WBTC or WMATIC are `METADEX/USDC` and `USDC/USDT`. The METADEX pool currently has ~$3M worth of USDC to take out as a flash loan; the USDT pool has ~$1M USDC.

Those amounts are probably enough for now, but the risk of running out of sufficient flash loan liquidity is another issue to be aware of on L2s.

### DIFF
--- a/scenario/LiquidationBotScenario.ts
+++ b/scenario/LiquidationBotScenario.ts
@@ -1,19 +1,44 @@
 import { scenario } from './context/CometContext';
 import { expect } from 'chai';
-import { isValidAssetIndex, MAX_ASSETS, timeUntilUnderwater } from './utils';
+import { isValidAssetIndex, matchesDeployment, MAX_ASSETS, timeUntilUnderwater } from './utils';
 import { ethers, event, exp, wait } from '../test/helpers';
 import CometActor from './context/CometActor';
 import { CometInterface, OnChainLiquidator } from '../build/types';
 import { getPoolConfig, flashLoanPools } from '../scripts/liquidation_bot/liquidateUnderwaterBorrowers';
 
-const BALANCER_VAULT = '0xBA12222222228d8Ba445958a75a0704d566BF2C8';
-const SUSHISWAP_ROUTER = '0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9F';
-const UNISWAP_ROUTER = '0xE592427A0AEce92De3Edee1F18E0157C05861564';
-const UNISWAP_V3_FACTORY = '0x1F98431c8aD98523631AE4a59f267346ea31F984';
+interface LiquidationAddresses {
+  balancerVault: string;
+  uniswapRouter: string;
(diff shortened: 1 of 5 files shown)


scenario/LiquidationBotScenario.ts
```diff
@@ -505,7 +604,7 @@ scenario(
     upgrade: {
       targetReserves: exp(20_000, 18)
     },
-    filter: async (ctx) => ctx.world.base.network === 'mainnet',
+    filter: async (ctx) => matchesDeployment(ctx, [{deployment: 'mainnet'}]), // XXX enable for Polygon
```
scott-silver, line 607: This scenario is challenging on Polygon. You need to take out a flash loan large enough that attempting to swap that amount will result in huge price slippage. Taking out a loan of that size if challenging on Polygon, however, since the pools available to provide flash loan are relatively small.

It may be possible to thread that needle at current prices/liquidity, but it seems very likely this test would start failing in the near future when prices or liquidity change; not sure what would be best here.

jflatow: Shouldn't the small market size mean you just have to take out a smaller loan for the huge slippage to occur?

scott-silver: sure, but it's still a fairly tight window between an amount that is big enough to cause sufficient slippage and the max amount of the flash loan you can borrow.

right now, swapping 35 WBTC (~$700K USDC) will cause sufficient slippage; but the pool we're using for flash loans only has $1M of USDC to borrow.

we can set a value right now that is within that window, but it wouldn't take a huge price increase in WBTC or a huge decrease in the pool liquidity before the scenario starts failing

jflatow (approved):

PR #683, opened 2023-02-14
Set rewards token address in Polygon configuration

kevincheng96:
This is so the rewards config can be set on deploy. The alternative is to set the reward config in the cross-chain proposal to initialize the Polygon market.


jflatow (approved):

jflatow:
Also maybe worth noting, since we are doing setConfiguration, whatever we set on deploy that we don't change in configuration.json will be set again to the same value

PR #672, opened 2023-02-01
Polygon enable cUSDCv3 proposal

scott-silver:
- [x] add description
- [x] add actual supply cap values
- [x] update reward speed
- [x] update ENS text record

PR #686, opened 2023-02-16
Fix Polygon scenarios

kevincheng96:
Issues fixed:

- The proposer not having enough gas to propose
- Some state synced events not being recognized while bridging
- Getting the reward token when it is not a collateral asset
- Liquidation bot scenarios failing due to high WBTC slippage

### DIFF
--- a/scenario/GovernanceScenario.ts
+++ b/scenario/GovernanceScenario.ts
@@ -159,7 +159,7 @@ scenario('add new asset',
 scenario(
   'execute Polygon governance proposal',
   {
-    filter: async ctx => ctx.world.base.network === 'mumbai' || ctx.world.base.network === 'polygon'
+    filter: async ctx => isBridgedDeployment(ctx)
   },
   async ({ comet, timelock, bridgeReceiver }, _context, world) => {
     const currentTimelockDelay = await timelock.delay();
@@ -195,7 +195,7 @@ scenario(
 scenario(
   'upgrade Polygon governance contracts and ensure they work properly',
   {
-    filter: async ctx => ctx.world.base.network === 'mumbai' || ctx.world.base.network === 'polygon'
+    filter: async ctx => isBridgedDeployment(ctx)
   },
   async ({ comet, configurator, proxyAdmin, timelock: oldLocalTimelock, bridgeReceiver: oldBridgeReceiver }, _context, world) => {
     const dm = world.deploymentManager;
--- a/scenario/GovernanceScenario.ts
+++ b/scenario/GovernanceScenario.ts
@@ -6,7 +6,7 @@ import { BaseBridgeReceiver, FaucetToken } from '../build/types';
 import { calldata } from '../src/deploy';
 import { COMP_WHALES } from '../src/deploy';
 import { impersonateAddress } from '../plugins/scenario/utils';
-import { isBridgedDeployment, fastL2GovernanceExecute } from './utils';
+import { isBridgedDeployment, fastL2GovernanceExecute, matchesDeployment } from './utils';
 import { World } from '../plugins/scenario';
 
 scenario('upgrade Comet implementation and initialize', {filter: async (ctx) => !isBridgedDeployment(ctx)}, async ({ comet, configurator, proxyAdmin }, context) => {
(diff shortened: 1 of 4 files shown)


scenario/GovernanceScenario.ts
```diff
@@ -157,9 +157,9 @@ scenario('add new asset',
   });
 
 scenario(
-  'execute Mumbai governance proposal',
+  'execute Polygon governance proposal',
   {
-    filter: async ctx => ctx.world.base.network === 'mumbai'
+    filter: async ctx => ctx.world.base.network === 'mumbai' || ctx.world.base.network === 'polygon'
```
scott-silver, line 162: we could use the `matchesDeployment` helper fn here

scenario/GovernanceScenario.ts
```diff
@@ -313,7 +313,8 @@ async function fastL1ToPolygonGovernanceExecute(
     throw new Error('cannot execute governance without governance deployment manager');
   }
 
-  const proposer = await impersonateAddress(governanceDeploymentManager, COMP_WHALES.testnet[0]);
+  const compWhale = world.base.network === 'polygon' ? COMP_WHALES.mainnet[0] : COMP_WHALES.testnet[0];
```
scott-silver, line 316: I think we should be able to use `ctx.getProposer()` here.

We basically mimic this logic in `getProposer`:

https://github.com/compound-finance/comet/pull/672/files#diff-a1090d669b0fd5457f94875e225762c4326775adfa733c3bfa9f8689ffa2c24dR70

scenario/LiquidationBotScenario.ts
```diff
@@ -101,7 +101,7 @@ for (let i = 0; i < MAX_ASSETS; i++) {
       upgrade: {
         targetReserves: exp(20_000, 18)
       },
-      filter: async (ctx) => await isValidAssetIndex(ctx, i) && matchesDeployment(ctx, [{deployment: 'mainnet'}, {network: 'polygon'}]),
+      filter: async (ctx) => await isValidAssetIndex(ctx, i) && matchesDeployment(ctx, [{network: 'mainnet'}, {network: 'polygon'}]),
```
scott-silver, line 104: 🙌

scenario/GovernanceScenario.ts
kevincheng96: Good rec. I changed it to use `isBridgedDeployment` instead since I think that's easier to maintain.

scenario/GovernanceScenario.ts
kevincheng96: Actually, I don't think we can because we need to use the `governanceDeploymentManager` here while `ctx.getProposer()` uses the base `DeploymentManager`.

scenario/GovernanceScenario.ts
scott-silver: `isBridgedDeployment` will be true for other L2 deployments (like Optimism in the near future), right? but these tests only apply to Mumbai and Polygon

scenario/GovernanceScenario.ts
scott-silver: aha, because of `impersonateAddress`

scenario/GovernanceScenario.ts
kevincheng96: ah you're absolutely right

kevincheng96: Done

scenario/GovernanceScenario.ts
kevincheng96: yep

scenario/LiquidationBotScenario.ts
```diff
@@ -95,13 +95,14 @@ for (let i = 0; i < MAX_ASSETS; i++) {
...
   };
+  // XXX enable for mainnet WETH
   scenario(
     `LiquidationBot > liquidates an underwater position of $asset${i} with no maxAmountToPurchase`,
     {
       upgrade: {
         targetReserves: exp(20_000, 18)
       },
-      filter: async (ctx) => await isValidAssetIndex(ctx, i) && matchesDeployment(ctx, [{network: 'mainnet'}, {network: 'polygon'}]),
+      filter: async (ctx) => await isValidAssetIndex(ctx, i) && matchesDeployment(ctx, [{network: 'mainnet', deployment: 'usdc'}, {network: 'polygon'}]),
```
kevincheng96, line 105: Mainnet WETH stalls for these liquidation bot scenarios because it cannot source that much base asset. I'm skipping these for now because they are not related to this PR.

jflatow (approved):
🧼

PR #688, opened 2023-02-17
Fix Liquidation Bot scenarios

scott-silver:
- add the fix for issue that ⛑️ found (`deployment` keys that should be `network` keys)
- reduce `cbETH` swap amounts from 1000 to 750 (due to shrinking liquidity)
- add deployment-specific base token amounts, so that we don't try to source 3M WETH for the `mainnet-weth` deployment


jflatow (approved):
Nice! LGTM🥼

kevincheng96 (approved):
Very nice

PR #692, opened 2023-02-21
Support multiple connections to Seacrest in Enact

kevincheng96:
Currently, running `Enact` in CI using Seacrest will fail when multiple networks are used (e.g. for cross-chain governance proposals that read from Polygon and write to mainnet). This is because Seacrest's connection with WalletConnect is for a specific chain, so trying to access a different chain will result in a `could not detect network` error.

This change supports making multiple connections to Seacrest/WalletConnect, all from a single phone. We introduce optional `GOV_NETWORK` and `GOV_NETWORK_PROVIDER` env variables to override the provider for a specific chain when creating the HREs. The `Enact` CI job determines the `GOV_NETWORK` based on what the original selected network is.

### DIFF
--- a/.github/workflows/enact-migration.yaml
+++ b/.github/workflows/enact-migration.yaml
@@ -12,9 +12,6 @@ on:
           - goerli
           - mumbai
           - polygon
-      different_enact_network:
-        type: boolean
-        description: Use a different network to enact the migration (e.g. used for cross-chain proposals that need to be made on mainnet)
       deployment:
         description: Deployment Name (e.g. "usdc")
         required: true
@@ -42,7 +39,6 @@ jobs:
       POLYGONSCAN_KEY: ${{ secrets.POLYGONSCAN_KEY }}
     steps:
       - name: Get governance network
-        if: ${{ github.event.inputs.different_enact_network == 'true' }}
         run: |
           case ${{ github.event.inputs.network }} in
             polygon)
@@ -50,7 +46,7 @@ jobs:
             mumbai)
                 echo "GOV_NETWORK=goerli" >> $GITHUB_ENV ;;
             *)
-              echo "No governance network for selected network" && exit 1 ;;
+                echo "No governance network for selected network" ;;
           esac
 
       - name: Seacrest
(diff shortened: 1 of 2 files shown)


hardhat.config.ts
```diff
@@ -39,7 +39,9 @@ const {
   POLYGONSCAN_KEY,
   REPORT_GAS = 'false',
   NETWORK_PROVIDER = '',
-  REMOTE_ACCOUNTS = '',
+  GOV_NETWORK_PROVIDER = '',
+  GOV_NETWORK = '',
```
kevincheng96, line 43: Approach recommended by🥈. This allows us to override a specific network's provider with the least amount of env variables. Otherwise, we might need a separate env variable per network.

.github/workflows/enact-migration.yaml
```diff
@@ -38,13 +41,32 @@ jobs:
       INFURA_KEY: ${{ secrets.INFURA_KEY }}
       POLYGONSCAN_KEY: ${{ secrets.POLYGONSCAN_KEY }}
     steps:
+      - name: Get governance network
+        if: ${{ github.event.inputs.different_enact_network == 'true' }}
```
jflatow, line 45: Should we even check/have an input, or just use the mapping?

jflatow (approved):
👏

.github/workflows/enact-migration.yaml
kevincheng96: Good point. You actually reminded me that we need this input to condition on the second Seacrest action. Basically, we should only show a second QR code if there is a second network to connect to. Will update.

kevincheng96: This change should be backwards-compatible in that if `inputs.different_enact_network == false`, this GH action will behave exactly like it did without this change.

kevincheng96: Done

jflatow: So 2 QR codes works then?

kevincheng96: Actually yeah, we could still get rid of the input here and simplify things.

kevincheng96: Ok got rid of the input. It will show the 2 QR codes only when it's on a network that has an auxiliary gov network (e.g. Polygon/Mumbai). Will test it manually again before merging.

kevincheng96: > So 2 QR codes works then?

Yeah 2 QR codes work. It can read from two different chains and write to whichever one you have up on your mobile wallet. I assume if you connect with two different mobile devices, you could write to both chains.

jflatow: Oh, yeah if you could do that I bet it would

kevincheng96:
Just tested this again and think it's ready to merge.

PR #699, opened 2023-02-27
Optimism deployment OZ feedback

scott-silver:


contracts/bridges/optimism/OvmL2CrossDomainMessengerInterface.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+interface OvmL2CrossDomainMessengerInterface {
```
kevincheng96, line 4: Should we prefix the interface with `I`? That's what we did with the Polygon `IFxMessageProcessor` interface.

deployments/optimism/usdc/deploy.ts
```diff
@@ -63,7 +63,7 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
 
   // Deploy Bulker
   const bulker = await deploymentManager.deploy(
-    'bulker',
+    'bulkers/BaseBulker.sol',
```
scott-silver, line 66: reversed these arguments, one minute

contracts/bridges/optimism/OvmL2CrossDomainMessengerInterface.sol
scott-silver: sure, that works too

scott-silver: updated

kevincheng96 (approved):
Nice

jflatow (approved):
🧼

PR #700, opened 2023-02-28
Static constraints and other scen improvements

jflatow:
This PR mainly adds the ability to have static constraints, which can be run after deploy on a base, and shared across all scenarios which are run. This saves a ton of redundant processing for things like migrations and proposal execution, and speeds up scenarios considerably (~300-600%)


scenario/context/CometContext.ts
```diff
@@ -394,28 +392,27 @@ async function getContextProperties(context: CometContext): Promise<CometPropert
...
-}
-
-export const constraints = [
-  new FilterConstraint(),
+export const staticConstraints = [
   new NativeTokenConstraint(),
   new MigrationConstraint(),
   new ProposalConstraint(),
   new VerifyMigrationConstraint(),
+];
```
kevincheng96, line 400: 🆒

plugins/scenario/Runner.ts
```diff
@@ -74,7 +91,7 @@ export class Runner<T, U, R> {
     let cumulativeGas = 0;
     for (const combo of combos(baseSolutions.concat(solutionChoices))) {
       // create a fresh copy of context that solutions can modify
-      let ctx: T = await scenario.forker(context, world);
+      let ctx: T = await env.initializer(world);
```
kevincheng96, line 94: Seems like forking context wasn't actually needed then?

tasks/scenario/task.ts
```diff
@@ -26,14 +25,12 @@ function getBasesFromTaskArgs(givenBases: string | undefined, env: HardhatRuntim
 task('scenario', 'Runs scenario tests')
   .addOptionalParam('bases', 'Bases to run on [defaults to all]')
   .addFlag('spider', 'run spider persistently before scenarios')
-  .addOptionalParam('stall', 'milliseconds to wait until we fail for stalling', 240_000, types.int)
-  .addOptionalParam('workers', 'count of workers', 1, types.int) // TODO: optimize parallelized workers better (1 per base?)
```
kevincheng96, line 30: Curious why we are getting rid of workers? Is it mainly for simplicity?

kevincheng96 (approved):
Nice changes! You might want to test against the Polygon branch first for better benchmarking since `main` doesn't have all the bridging scenarios yet.

jflatow:
> Nice changes! You might want to test against the Polygon branch first for better benchmarking since `main` doesn't have all the bridging scenarios yet.

I actually rebased at the end today, benchmarking was on Polygon proposal branch

plugins/scenario/Runner.ts
jflatow: Oh I was wondering where this went, I actually had some old changes that I think actually slipped in removing forking first, I should maybe add it back, its possibly a little weird conceptually right now. I'll take another look at this tmr

tasks/scenario/task.ts
jflatow: Yeah the first step was to remove the workers to simplify to be able to do the refactor (which I don't think anyone is using now as it defaults without them since a few months ago)

plugins/scenario/Runner.ts
jflatow: Added a note about it, its not really necessary at all right now, but could be added back if we need to communicate something from static constraints to dynamic ones

PR #702, opened 2023-02-28
CometRewards with multiplier

kevincheng96:
This allows us to scale the Comet accrued tracking balances by varying amounts for different reward tokens.

### DIFF
--- a/contracts/CometRewards.sol
+++ b/contracts/CometRewards.sol
@@ -14,6 +14,8 @@ contract CometRewards {
         address token;
         uint64 rescaleFactor;
         bool shouldUpscale;
+        // Note: We define new variables after existing variables to keep interface backwards-compatible
+        uint256 multiplier;
     }
 
     struct RewardOwed {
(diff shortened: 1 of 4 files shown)


contracts/CometRewards.sol
```diff
@@ -55,29 +59,64 @@ contract CometRewards {
...
+                    shouldUpscale: false
+                });
+            } else {
+                rescaleFactor = safe64(uint256(rescaleFactor) * multiplier / FACTOR_SCALE);
+                rewardConfig[comet] = RewardConfig({
+                    token: token,
+                    rescaleFactor: rescaleFactor,
+                    shouldUpscale: true
+                });
+            }
```
jflatow, line 107: Yeah definitely more complex, but maybe not *so* bad?

kevincheng96: Probably acceptable given the fact that it maintains backward compatibility.

contracts/CometRewards.sol
```diff
@@ -69,43 +69,34 @@ contract CometRewards {
...
-                });
+                shouldUpscale = true;
             }
         }
+
+        rewardConfig[comet] = RewardConfig({
+            token: token,
+            rescaleFactor: rescaleFactor,
+            shouldUpscale: shouldUpscale
+        });
```
jflatow, line 99: 👍

contracts/CometRewards.sol
kevincheng96: Refactored the code slightly to try to improve readability

kevincheng96:
While writing scenarios, I noticed an edge case where applying the `multiplier` directly to the `rescaleFactor` could lose a lot of precision if the `rescaleFactor` is already small. e.g. if `rescaleFactor` is `100` and `multiplier` is `0.00355`, we can't precisely apply the `multiplier` to the `rescaleFactor` because `100 * 0.00355 < 1`.

The best solution (which also happens to be the simplest implementation) is to just store the `multiplier` directly in the `RewardConfig`. Adding the `multiplier` to the end of the `RewardConfig` should make the interface backwards-compatible for tests/UIs that call the getter.

contracts/CometRewards.sol
kevincheng96: Due to an edge case with this implementation that I caught while writing scenarios, I decided to re-implement with a simpler solution. See: https://github.com/compound-finance/comet/pull/702#issuecomment-1468736922

test/rewards-test.ts
```diff
@@ -665,6 +666,60 @@ describe('CometRewards', () => {
...
+        tokens: { COMP },
+        users: [alice],
+      } = await makeProtocol();
+      const { rewards } = await makeRewards({ governor, configs: [[comet, COMP]] });
+
+      await expect(
+        rewards
+          .connect(alice)
+          .setRewardsClaimed(comet.address, [alice.address], [exp(100, 18)])
+      //).to.be.revertedWith(`custom error 'NotPermitted("${alice.address}")'`);
```
scott-silver, line 718: left in intentionally?

scott-silver (approved):
very thorough tests!

test/rewards-test.ts
kevincheng96: Yep, left in on purpose since we can't directly check that the `address` arg in the event is `alice.address`.

contracts/CometRewards.sol
```diff
@@ -67,17 +75,44 @@ contract CometRewards {
...
             });
         }
     }
 
+    /**
+     * @notice Set the reward token for a Comet instance
+     * @param comet The protocol instance
+     * @param token The reward token address
+     */
+    function setRewardConfig(address comet, address token) external {
```
jflatow, line 96: I suppose this was the main compatibility question anyway, glad we have this

test/helpers.ts
```diff
@@ -476,8 +476,9 @@ export async function makeRewards(opts: RewardsOpts = {}): Promise<Rewards> {
   const rewards = await RewardsFactory.deploy(governor.address);
   await rewards.deployed();
 
-  for (const [comet, token] of configs) {
-    await wait(rewards.setRewardConfig(comet.address, token.address));
+  for (const [comet, token, multiplier] of configs) {
+    if (multiplier === undefined) await wait(rewards.setRewardConfig(comet.address, token.address));
+    else await wait(rewards.setRewardConfigWithMultiplier(comet.address, token.address, multiplier));
   }
```
jflatow, line 482: Should we read it back out here?

jflatow (approved):
💪

test/helpers.ts
kevincheng96: Discussed offline. This is just a helper function to instantiate the rewards contract. The unit tests would be reading and asserting the values are set correctly.

PR #703, opened 2023-03-01
Add ability to pre-populate rewards claimed for users in CometRewards

kevincheng96:
This PR introduces a `setRewardsClaimed` function that gives the admin the ability to set rewards claimed for any user. This gives us the ability to initialize a new reward distribution that doesn't necessarily have to retroactively reward all past usage of Comet. 

We would have to compute off-chain the amount that each user would be able to claim at a specific block number `x` and write it on-chain via the `setRewardsClaimed` function. This allows us to essentially set the start time of the rewards accrual to be after that specific block number `x`.

### DIFF
--- a/contracts/CometRewards.sol
+++ b/contracts/CometRewards.sol
@@ -42,6 +42,7 @@ contract CometRewards {
     /** Custom errors **/
 
     error AlreadyConfigured(address);
+    error BadData();
     error InvalidUInt64(uint);
     error NotPermitted(address);
     error NotSupported(address);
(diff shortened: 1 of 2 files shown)


jflatow (approved):
Very nice 🧼 I don't know if it matters but we *could* also require the existing value to be zero, I suppose

kevincheng96:
> Very nice 🧼 I don't know if it matters but we _could_ also require the existing value to be zero, I suppose

Yeah we could. Though the tradeoff is we would be able to set less accounts per txn because of the higher gas cost for reading. The check also wouldn't protect against malicious governors since they could just withdraw the rewards. So, I could add it if we want the extra on-chain sanity check to make sure we aren't accidentally overwriting an existing field.

scott-silver (approved):
Nice!

PR #704, opened 2023-03-02
Add relation to spider the rewards token

kevincheng96:
This allows us to spider the rewards token directly from the rewards contract. This is useful for networks where the rewards token (e.g. COMP) is not a collateral asset.


jflatow (approved):
Great 💪

scott-silver (approved):
👍

PR #705, opened 2023-03-03
Example script for seeding state on a fork

kevincheng96:
This is an example script that can be used to seed a fork with some initial state relevant for testing Comet.


jflatow (approved):
💪

PR #706, opened 2023-03-06
`hasPurchasableCollateral` fixes for liquidation bot

scott-silver:
After checking all addresses, the Liquidation Bot does a check for any purchasable collateral left in the protocol.

It checks the value of `baseReserves` against the value of `targetReserves` before checking for each asset.

`baseReserves` and `targetReserves` are both BigNumbers however, and they have their own comparison functions:

```
> BigNumber.from("10") > BigNumber.from("5")
false
> BigNumber.from("10").gt(BigNumber.from("5"))
true
```

Converting those values to BigInts first should get us the comparisons we expect.

### DIFF
--- a/scripts/liquidation_bot/index.ts
+++ b/scripts/liquidation_bot/index.ts
@@ -99,7 +99,7 @@ async function main() {
   let lastBlockNumber: number;
   let loops = 0;
   while (true) {
-    if (loops >= loopsUntilUpdateAssets) {
+    if (assets.length == 0 || loops >= loopsUntilUpdateAssets) {
       googleCloudLog(LogSeverity.INFO, 'Updating assets');
       assets = await getAssets(comet);
       loops = 0;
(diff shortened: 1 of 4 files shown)


jflatow (approved):
💪

kevincheng96 (approved):
🕵️

scott-silver:
Dug in a bit further.

BigNumbers behave very strangely when used with normal comparison operators:

```
> BigNumber.from("10") >= BigNumber.from("1000000000000000000")
false
> BigNumber.from("11") >= BigNumber.from("1000000000000000000")
true
```

We had some false positives in our unit tests. Several tests intended to check that there was no more than $1 worth of an asset in the protocol, but they were actually checking that there was no more than 1 wei USDC of that asset in the protocol.

Those tests should have failed, but were succeeding because `hasPurchaseableCollateral` was returning `false` because the protocol had exceeded targetReserves.

PR #708, opened 2023-03-08
Fix typescript errors

kevincheng96:
Fixes some typescript errors.

### DIFF
--- a/src/deploy/NetworkConfiguration.ts
+++ b/src/deploy/NetworkConfiguration.ts
@@ -187,5 +187,5 @@ export async function getConfigurationStruct(
   const contracts = await deploymentManager.contracts();
   const configuration = await getConfiguration(deploymentManager, configOverrides);
   const extensionDelegate = configOverrides.extensionDelegate ?? getContractAddress('comet:implementation:implementation', contracts);
-  return { ...configuration, extensionDelegate };
+  return { ...(configuration as ConfigurationStruct), extensionDelegate };
 }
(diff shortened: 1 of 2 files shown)


src/deploy/NetworkConfiguration.ts
```diff
@@ -187,5 +187,5 @@ export async function getConfigurationStruct(
   const contracts = await deploymentManager.contracts();
   const configuration = await getConfiguration(deploymentManager, configOverrides);
   const extensionDelegate = configOverrides.extensionDelegate ?? getContractAddress('comet:implementation:implementation', contracts);
-  return { ...configuration, extensionDelegate };
+  return { ...(configuration as ConfigurationStruct), extensionDelegate };
```
kevincheng96, line 190: I don't think this would change any runtime behavior, but to be doubly safe I could also implement a type guard here. Would be slightly tedious because it would have to check for the presence for each field, but maybe worth it for the extra safety.

jflatow: Should we just make sure `getConfiguration` return is typed and possibly do the casting there if needed?

kevincheng96: `getConfiguration` does currently return a `ProtocolConfiguration` type. It's basically a superset of `ConfigurationStruct` except all of its fields are optional whereas `ConfigurationStruct`'s fields are all required. `ProtocolConfiguration` also has a few extra fields that are used by the deploy scripts, so we'll have to do some refactoring if we wanted to change the return type of `getConfiguration`.

I can take a look and see if these types can be simplified.

jflatow: Oh I see, in that case should we just cast it on the assignment line? I think it might have done that at one point already

kevincheng96: Sure just did that.

jflatow (approved):

PR #711, opened 2023-03-08
Fix more typescript type inconsistencies

kevincheng96:
This fixes some more type inconsistencies that Typescript was complaining about throughout the scenario codebase.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
     "moduleResolution": "node",
     "declaration": true
   },
-  "include": ["./scripts", "./test", "./tasks", "./plugins"],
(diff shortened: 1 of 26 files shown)


tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "moduleResolution": "node",
     "declaration": true
   },
-  "include": ["./scripts", "./test", "./tasks"],
+  "include": ["./scripts", "./test", "./tasks", "./plugins", "./scenario"],
```
scott-silver, line 10: 🥳

jflatow (approved):
💪

kevincheng96:
> should we include other directories (like `./plugins` and `./scenarios`), too?

Good catch! Just added these and resolved a few more typescript errors.

PR #715, opened 2023-03-12
Fix typo in Proxy.sol

eltociear:
overriden -> overridden


jflatow:
This is a vendored contract that was written externally, we use them here as is. But this could be a PR to the OZ repo

PR #716, opened 2023-03-13
Fix coverage

kevincheng96:
This fixes our coverage CI job by turning on the optimizer when compiling contracts.


jflatow (approved):

jflatow (commented):
Makes sense if we can't compile without viaIR anymore

PR #717, opened 2023-03-14
Arbitrum mainnet

scott-silver:


deployments/arbitrum/usdc/configuration.json
```diff
@@ -0,0 +1,55 @@
...
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0e15",
+    "baseBorrowSpeed": "0e15",
+    "baseMinForRewards": "10000e6"
+  },
+  "assets": {
+    "LINK": {
```
scott-silver, line 27: Not final assets or configuration parameters; just needed some values to get scenarios passing.

Added LINK because WBTC is difficult to source in large quantities, and using WETH as `$asset0` is incompatible with the Bulker scenarios.

PR #718, opened 2023-03-15
Testing Base

jflatow:
Currently using a random USDC and a symlink to the base optimism deployment, using the optimism deploy scripts for testing, and not using assets there are no price feeds for yet. Finally succeeds on deploying Comet, but fails on importing USDC as the contract is not verified.


kevincheng96:
Closing in favor of #744 744

PR #719, opened 2023-03-17
Arbitrum proposals

scott-silver:
- [x] add `ARBISCAN_KEY` to Github secrets
- [x] update configuration.json with actual supply caps
- [x] add proposal description
- [x] determine correct L2 refund address
- [x] uncomment the `diffState` test (once contracts are deployed)`
- [x] add GMX
- [x] ~~decide on baseMinForRewards~~ set to 10K
- [x] add setRewardConfig
- [x] set supply caps
- [x] update references to `l1GatewayRouter` to `arbitrumL1GatewayRouter`
- [x] updates references to `inbox` to `arbitrumInbox`
- [x] move COMP distribution from v2 USDT borrow to Arbitrum supply side 
- [x] finalize the amount of USDC and COMP to bridge


deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,321 @@
...
+
+    const l1GatewayRouterInterace = l1GatewayRouter.interface;
+
+    const USDCAmountToBridge = exp(10_000, 6);
+    const COMPAmountToBridge = exp(2_500, 18);
+    const usdcGatewayAddress = await l1GatewayRouter.getGateway(USDC.address);
+    const compGatewayAddress = await l1GatewayRouter.getGateway(COMP.address);
+    const refundAddress = comet.address; // XXX timelock address?
+
+    const usdcDummyCallData = l1GatewayRouterInterace.encodeFunctionData(
```
scott-silver, line 55: This is the slightly unfortunate part of this migration.

For each bridged action, we have to pass in several gas-related fields (maxGas/gasLimit, gasPriceBid/maxFeePerGas, maxSubmissionCost/maxSubmissionFee, as well as the value).

One way to get those estimations is to encode the function data (with dummy values for the gas-related fields), and then pass those to function from the Arbitrum SDK (`L1ToL2MessageGasEstimator#estimateAll`). That function requires a specific type of Provider, however (a JsonRpcProvider) because it calls a few special provider functions (like `getGasPrice`). The default provider that Hardhat gives you is an `EthereumProvider`.

This solution works, but it's pretty verbose. We might be able to reuse the same dummy call data for both USDC and COMP bridging (the values are very close, and the SDK function includes some buffer).

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,321 @@
...
+    const l2Provider = new providers.JsonRpcProvider(`https://arbitrum-mainnet.infura.io/v3/${INFURA_KEY}`);
+    const gasEstimator = new L1ToL2MessageGasEstimator(l2Provider);
+
+    const l1GatewayRouterInterace = l1GatewayRouter.interface;
+
+    const USDCAmountToBridge = exp(10_000, 6);
+    const COMPAmountToBridge = exp(2_500, 18);
+    const usdcGatewayAddress = await l1GatewayRouter.getGateway(USDC.address);
+    const compGatewayAddress = await l1GatewayRouter.getGateway(COMP.address);
+    const refundAddress = comet.address; // XXX timelock address?
```
scott-silver, line 53: if there is unused gas for bridged actions, Arbitrum has you specify an L2 address that will receive the refund. Not sure if that should be Comet. maybe the timelock?

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
jflatow: Hmm yeah this seems like it could be challenging to make Arbitrum proposals

jflatow (commented):
Looking great 💪

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
scott-silver: The Arbitrum docs lay out the formula for estimating gas: https://developer.arbitrum.io/devs-how-tos/how-to-estimate-gas

It might be possible to recreate the calculations from `estimateAll` on our own (and potentially without the JsonRpcProvider), but we'd still need the dummy call data

jflatow: Right we could populate with https://github.com/OffchainLabs/arbitrum-sdk/blob/main/src/lib/message/L1ToL2MessageGasEstimator.ts#L215

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,355 @@
...
+    'function finalizeInboundTransfer(address _token, address _from, address _to, uint256 _amount, bytes calldata _data) external payable'
+  ]
+);
+
+async function estimateL2Transaction(
+  { from, to, data }: { to: string, from: string, data: string },
+  l2DeploymentManager: DeploymentManager
+) {
+  // guess what the l1 gas price will be when the proposal is executed
+  const l1GasPrice = (utils.parseUnits('33', 'gwei')).toNumber();
```
scott-silver, line 31: we have a few different places to add buffer to this calculation.

- we can increase the `l1GasPrice` and the `l2GasPrice
- we can add a non-zero `gasBuffer`
- we can increase the buffer on the `l2GasLimit`

scott-silver: we should almost certainly bump the l1GasPrice higher. maybe 100 gwei

kevincheng96: `min(100, currentGas * 2)`?

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,355 @@
...
+  // XXX add buffer?
```
kevincheng96, line 32: Yeah let's add a buffer for sure.

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,355 @@
...
+import { calldata, exp, getConfigurationStruct, proposal } from '../../../../src/deploy';
+
+const ENSName = 'compound-community-licenses.eth';
+const ENSResolverAddress = '0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41';
+const ENSSubdomainLabel = 'v3-additional-grants';
+const ENSSubdomain = `${ENSSubdomainLabel}.${ENSName}`;
+const ENSTextRecordKey = 'v3-official-markets';
+
+// https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-eth/contracts/libraries/AddressAliasHelper.sol
+function applyL1ToL2Alias(l1Address: string) {
```
kevincheng96, line 14: I think we should move all these helpers into a separate util file to be shared with future migrations.

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,355 @@
...
+
+  const bytesLength = utils.hexDataLength(data);
+  // https://etherscan.io/address/0x5aed5f8a1e3607476f1f81c3d8fe126deb0afe94#code
+  // calculateRetryableSubmissionFee
+  const submissionCost = (1400 + 6 * bytesLength) * l1GasPrice;
+  const submissionCostWithMargin = utils.parseUnits('10', 'gwei').add(submissionCost);
+
+  const deposit = submissionCostWithMargin.add(l2GasLimit.mul(l2GasPrice));
+
+  // XXX add l2CallValue?
```
kevincheng96, line 53: We expect this to always be 0 right? Only reason it would be populated is if we want to send ETH over the bridge?

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
kevincheng96: +1 on setting this to the L2 Timelock

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    // 2. & 3.
+    expect(await comet.getReserves()).to.be.equal(exp(10_000, 6));
+
+    // 4. & 5.
+    const arbitrumCOMP = new Contract(
+      '0x354a6da3fcde098f8389cad84b0182725c6c91de',
+      ['function balanceOf(address account) external view returns (uint256)'],
+      deploymentManager.hre.ethers.provider
+    );
```
kevincheng96, line 320: We could also call spider in `verify()` and pull COMP from the deployment manager.

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,355 @@
...
+    const configuration = await getConfigurationStruct(deploymentManager);
+
+    const setConfigurationCalldata = await calldata(
+      configurator.populateTransaction.setConfiguration(comet.address, configuration)
+    );
+    const deployAndUpgradeToCalldata = utils.defaultAbiCoder.encode(
+      ['address', 'address'],
+      [configurator.address, comet.address]
+    );
+    const l2ProposalData = utils.defaultAbiCoder.encode(
```
kevincheng96, line 168: We also need to `setRewardConfig` in the rewards contract. See this example https://github.com/compound-finance/comet/blob/c91bffce984ec6f8276b6f7e9ef362d9293416c1/deployments/optimism-goerli/usdc/migrations/1679592519_configurate_and_ens.ts#L77

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    // 1.
+    const wbtcInfo = await comet.getAssetInfoByAddress(WBTC.address);
+    const wethInfo = await comet.getAssetInfoByAddress(WETH.address);
+    const arbInfo = await comet.getAssetInfoByAddress(ARB.address);
+    expect(wbtcInfo.supplyCap).to.be.eq(exp(400, 8));
+    expect(wethInfo.supplyCap).to.be.eq(exp(11_000, 18));
+    expect(arbInfo.supplyCap).to.be.eq(exp(10_000_000, 18));
+    // XXX
+    // expect(await comet.pauseGuardian()).to.be.eq('');
```
kevincheng96, line 310: Let's use the state differ here to have an exact check of all the params that have changed. See example: https://github.com/compound-finance/comet/blob/c91bffce984ec6f8276b6f7e9ef362d9293416c1/deployments/optimism-goerli/usdc/migrations/1679592519_configurate_and_ens.ts#L158-L174

You might need to carry over a few lines of changes to `DiffState.ts` from the Optimism branch.

kevincheng96 (commented):
Nice, looking good 👍

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
scott-silver: reworked the gas estimation

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
scott-silver: either way, we're making a guess at what the gas price is going to be when the proposal is executed in a week.

we can definitely do a min, but it might be easiest to just be explicit about what the value was.

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
scott-silver: moved into `scenario/utils/arbitrumUtils.ts`

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
scott-silver: updated `DiffState.ts` and added a diffState-based test in place of these checks.

i've commented it out for the time-being; will uncomment it once we've deployed the contracts

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
scott-silver: updated to the L2 timelock

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
scott-silver: good call. updated

jflatow (approved):

scenario/BulkerScenario.ts
```diff
@@ -168,6 +168,10 @@ scenario(
       albert: { $base: '== 1000000', $asset0: 100 },
       $comet: { $base: 5000 },
     },
+    prices: {
```
scott-silver, line 171: This bulker scenario makes an assumption about the relative price of the base asset and asset0; added an explicit price to make sure it's true

deployments/arbitrum/usdc/configuration.json
```diff
@@ -31,7 +31,7 @@
       "borrowCF": 0.79,
       "liquidateCF": 0.85,
       "liquidationFactor": 0.93,
-      "supplyCap": "0e18"
+      "supplyCap": "10000000e18"
```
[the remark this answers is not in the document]
scott-silver: added GMX

kevincheng96:
Proposal for goerli successfully executed.

Goerli txn: https://goerli.etherscan.io/tx/0x327a2907df01f8bafc51efd6a76102a26a651bed999123d43775228c27db5b21

Arbitrum Goerli txn: https://goerli.arbiscan.io/tx/0xf814555086f2f34adb705e6f7924f3c4ca723d1652476c52ce2f16ead4ea5862

deployments/arbitrum/usdc/migrations/1679020486_configurate_and_ens.ts
kevincheng96: We set this as 100 to be safe.

deployments/arbitrum-goerli/usdc/migrations/1679518383_configurate_and_ens.ts
```diff
@@ -0,0 +1,296 @@
...
+    trace(`Created proposal ${proposalId}.`);
+  },
+
+  async enacted(deploymentManager: DeploymentManager): Promise<boolean> {
+    return true;
+  },
+
+  async verify(deploymentManager: DeploymentManager, govDeploymentManager: DeploymentManager, preMigrationBlockNumber: number) {
+    const ethers = deploymentManager.hre.ethers;
+    await deploymentManager.spider(); // await deploymentManager.spider(); // Pull in Arbitrum COMP now that reward config has been set
```
scott-silver, line 217: delete extra comment

scenario/context/CometContext.ts
```diff
@@ -43,6 +43,7 @@ export type MigrationData = {
   migration: Migration<any>;
   lastProposal?: number;
   preMigrationBlockNumber?: number;
+  skip?: boolean;
```
scott-silver, line 46: maybe `skipVerify` to be more specific about what we're skipping?

kevincheng96: Done

kevincheng96:
[Proposal 160](https://compound.finance/governance/proposals/160) has been created on mainnet: https://github.com/compound-finance/comet/actions/runs/4911368836/jobs/8769343015

kevincheng96 (approved):
Should be ready to merge now that the proposal has been created on both testnet and mainnet.

PR #720, opened 2023-03-17
Use getProposer

jflatow:


plugins/scenario/World.ts
```diff
@@ -84,7 +84,8 @@ export class World {
   }
 
   async impersonateAddress(address: string, value?: bigint): Promise<SignerWithAddress> {
-    return impersonateAddress(this.deploymentManager, address, value);
+    const dm = this.auxiliaryDeploymentManager ?? this.deploymentManager;
```
kevincheng96, line 87: I think this might break our other scenarios that are impersonating on the main (non-auxiliary) network. WDYT about splitting this into a separate function called `impersonateAddressOnAuxiliary`?

kevincheng96 (approved):
Otherwise, LGTM. Feel free to merge as is and I can address issues on my branch.

plugins/scenario/World.ts
scott-silver, line 87: this change makes me nervous; things start behaving strangely when you impersonate an address with the wrong deployment manager.

I got stuck for a long time yesterday because I was impersonating an address with the wrong deployment manager. transactions that I knew should be reverting were executing successfully.

swapping out this value seems likely to cause difficult debugging issues in the future

scott-silver: agreed; we should be really explicit about which deployment manager we're impersonating from

jflatow: Fair, but then should we just pass a flag to make it explicit?

jflatow:
> Otherwise, LGTM. Feel free to merge as is and I can address issues on my branch.

Yeah I think we should merge even if we want to add an explicit flag to make sure we get the inlining of whales out of deploy.ts

plugins/scenario/World.ts
scott-silver: our just use the underyling `impersonateAddress` function directly

kevincheng96:
Closing since this commit has been included in #722.

PR #721, opened 2023-03-20
Fix Optimism scenarios

kevincheng96:
~~This PR refactors the cross-chain governance scenarios so that a single scenario can be used for all L2s/sidechains. We also move cross-chain governance scenarios into their own file.~~

This PR fixes a lot of different errors that are cropping up for Optimism scenarios.

### DIFF
--- a/scenario/utils/index.ts
+++ b/scenario/utils/index.ts
@@ -58,9 +58,10 @@ export function expectRevertCustom(tx: Promise<ContractReceipt | ContractTransac
       const patterns = [
         new RegExp(`custom error '${custom.replace(/[()]/g, '\\$&')}'`),
         new RegExp(`unrecognized custom error with selector ${selector}`),
+        new RegExp(`unrecognized custom error \\(return data: 0x${selector}\\)`)
       ];
       for (const pattern of patterns)
-        if (pattern.test(e.message))
+        if (pattern.test(e.message) || pattern.test(e.reason))
           return;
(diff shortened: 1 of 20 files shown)


kevincheng96:
Fixed up all the immediately fixable scenarios for Optimism. 

The one remaining failure for Optimism is due to our scenario code using a Bedrock-only interface when Optimism mainnet hasn't been upgraded yet. The one remaining failure for Optimism-goerli is due to the fact that we only have a single collateral asset in the deploy script right now, which would fail for any scenarios that try to bump the supply cap for a collateral asset that is not asset 0.

jflatow (approved):
Great work 💯

scenario/utils/index.ts
```diff
@@ -415,6 +416,65 @@ export async function fastL2GovernanceExecute(
...
+  const signatures: string[] = [];
+  const calldata: string[] = [];
+
+  // Create the chain-specific wrapper around the L2 proposal data
+  switch (bridgeNetwork) {
+    case 'optimism':
+    case 'optimism-goerli': {
+      const sendMessageCalldata = utils.defaultAbiCoder.encode(
+        ['address', 'bytes', 'uint32'],
+        [bridgeReceiver.address, l2ProposalData, 1_000_000] // XXX find a reliable way to estimate the gasLimit
```
jflatow, line 435: 🤔

scott-silver (approved):
👍

PR #722, opened 2023-03-20
Simplify cross-chain governance scenarios

kevincheng96:
This PR refactors the cross-chain governance scenarios so that a single scenario can be shared across all L2s/sidechains, rather than having to implement a separate scenario per network. We also move cross-chain governance scenarios into their own file to differentiate them from normal governance scenarios.


jflatow (approved):
Very nice 💪

PR #725, opened 2023-03-23
Porting in OZ clone

jflatow:
https://github.com/OpenZeppelin/compound-cross-chain-pause-demo/tree/main/defender-multisig-setup


scripts/clone-multisig.ts
```diff
@@ -0,0 +1,44 @@
...
+  const guardian = await comet.pauseGuardian();
+
+  // Get owners and threshold from existing multisig
+  const GnosisABI = [
+    {"constant":true,"inputs":[],"name":"getOwners","outputs":[{"internalType":"address[]","name":"","type":"address[]"}],"payable":false,"stateMutability":"view","type":"function"},
+    {"constant":true,"inputs":[],"name":"getThreshold","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"payable":false,"stateMutability":"view","type":"function"},
+  ];
+  const GnosisSafeContract = new hre.ethers.Contract(guardian, GnosisABI, hre.ethers.provider);
+  const owners = await GnosisSafeContract.getOwners();
+  const threshold = await GnosisSafeContract.getThreshold();
```
kevincheng96, line 25: Should we check that these are non-zero values, just to check that the `pauseGuardian` is indeed a gnosis multi-sig?

kevincheng96 (approved):
LGTM

scripts/clone-multisig.ts
jflatow: These calls will revert if its not a multisig. It could be empty but it also prints out what config so you can see 🤷

PR #729, opened 2023-03-30
Diff checker for state changes after running a migration

kevincheng96:
This adds a diff checker to check for state changes after a migration is run. This allows us to more comprehensively check for state changes because, right now, our migration verification logic will not catch any unintentional state changes. For example, in a proposal that only touches collateral factors, we'll check that `collateralFactor == some new amount`, but not that `liquidationFactor == old amount`. We could also potentially generate a report of all the state changes for easier viewing.

In order to isolate the state changes of each migration, we get rid of the `VerifyMigrationConstraint` and move the verify checks directly into the `ProposalConstraint` so that verification is called right after a proposal is executed. This way, we can verify the state changes of a specific migration without having the diffs be affected by other migrations or pending proposals.

Example usage ([branch](https://github.com/compound-finance/comet/pull/730/files)):
```
const stateChanges = await diffState(comet, getCometConfig, preMigrationBlockNumber, postMigrationBlockNumber);
expect(stateChanges).to.deep.equal({
  COMP: {
    supplyCap: exp(600_000, 18)
  },
  WBTC: {
    supplyCap: exp(6_000, 8)
  },
  WETH: {
    supplyCap: exp(75_000, 18)
  }
});
```

Example output from logs (could also be dumped as a report):
```
State changes after migration
+ New config
- Old config

  Object {
    "COMP": Object {
      "asset": "0xc00e94Cb662C3520282E6f5717214004A7f26888",
      "borrowCollateralFactor": 650000000000000000n,
      "liquidateCollateralFactor": 700000000000000000n,
      "liquidationFactor": 880000000000000000n,
      "offset": 0,
      "priceFeed": "0xdbd020CAeF83eFd542f4De03e3cF0C28A4428bd5",
      "scale": 1000000000000000000n,
+     "supplyCap": 600000000000000000000000n,
-     "supplyCap": 900000000000000000000000n,
    },
    "LINK": Object {
      "asset": "0x514910771AF9Ca656af840dff83E8264EcF986CA",
      "borrowCollateralFactor": 790000000000000000n,
      "liquidateCollateralFactor": 850000000000000000n,
      "liquidationFactor": 930000000000000000n,
      "offset": 4,
      "priceFeed": "0x2c1d072e956AFFC0D435Cb7AC38EF18d24d9127c",
      "scale": 1000000000000000000n,
      "supplyCap": 1250000000000000000000000n,
    },
    "UNI": Object {
      "asset": "0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984",
      "borrowCollateralFactor": 750000000000000000n,
      "liquidateCollateralFactor": 810000000000000000n,
      "liquidationFactor": 930000000000000000n,
      "offset": 3,
      "priceFeed": "0x553303d460EE0afB37EdFf9bE42922D8FF63220e",
      "scale": 1000000000000000000n,
      "supplyCap": 2300000000000000000000000n,
    },
    "WBTC": Object {
      "asset": "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599",
      "borrowCollateralFactor": 700000000000000000n,
      "liquidateCollateralFactor": 770000000000000000n,
      "liquidationFactor": 950000000000000000n,
      "offset": 1,
      "priceFeed": "0xF4030086522a5bEEa4988F8cA5B36dbC97BeE88c",
      "scale": 100000000n,
+     "supplyCap": 600000000000n,
-     "supplyCap": 1200000000000n,
    },
    "WETH": Object {
      "asset": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
      "borrowCollateralFactor": 825000000000000000n,
      "liquidateCollateralFactor": 895000000000000000n,
      "liquidationFactor": 950000000000000000n,
      "offset": 2,
      "priceFeed": "0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419",
      "scale": 1000000000000000000n,
+     "supplyCap": 75000000000000000000000n,
-     "supplyCap": 350000000000000000000000n,
    },
    "baseBorrowMin": 100000000n,
    "baseMinForRewards": 1000000000000n,
    "baseToken": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
    "baseTokenPriceFeed": "0x8fFfFfd4AfB6115b954Bd326cbe7B4BA576818f6",
    "baseTrackingBorrowSpeed": 3257060185185n,
    "baseTrackingSupplySpeed": 0n,
    "borrowKink": 800000000000000000n,
    "borrowPerSecondInterestRateBase": 475646879n,
    "borrowPerSecondInterestRateSlopeHigh": 7927447995n,
    "borrowPerSecondInterestRateSlopeLow": 1109842719n,
    "extensionDelegate": "0x285617313887d43256F852cAE0Ee4de4b68D45B0",
    "governor": "0x6d903f6003cca6255D85CcA4D3B5E5146dC33925",
    "numAssets": 5,
    "pauseGuardian": "0xbbf3f1421D886E9b2c5D716B5192aC998af2012c",
    "storeFrontPriceFactor": 600000000000000000n,
    "supplyKink": 800000000000000000n,
    "supplyPerSecondInterestRateBase": 0n,
    "supplyPerSecondInterestRateSlopeHigh": 12683916793n,
    "supplyPerSecondRateSlopeLow": 1030568239n,
    "targetReserves": 5000000000000n,
  }
```

### DIFF
--- a/plugins/deployment_manager/Migration.ts
+++ b/plugins/deployment_manager/Migration.ts
@@ -5,7 +5,7 @@ export interface Actions<T> {
   prepare: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<T>;
   enact: (dm: DeploymentManager, govDm: DeploymentManager, t: T) => Promise<void>;
   enacted?: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<boolean>;
-  verify?: (dm: DeploymentManager, govDm: DeploymentManager, preMigrationBlockNumber: number, postMigrationBlockNumber: number) => Promise<void>;
+  verify?: (dm: DeploymentManager, govDm: DeploymentManager, preMigrationBlockNumber: number) => Promise<void>;
 }
 
 export class Migration<T> {
--- a/plugins/deployment_manager/Migration.ts
+++ b/plugins/deployment_manager/Migration.ts
@@ -5,7 +5,7 @@ export interface Actions<T> {
   prepare: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<T>;
   enact: (dm: DeploymentManager, govDm: DeploymentManager, t: T) => Promise<void>;
   enacted?: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<boolean>;
-  verify?: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<void>;
+  verify?: (dm: DeploymentManager, govDm: DeploymentManager, preMigrationBlockNumber: number, postMigrationBlockNumber: number) => Promise<void>;
 }
 
 export class Migration<T> {
--- a/scenario/constraints/MigrationConstraint.ts
+++ b/scenario/constraints/MigrationConstraint.ts
@@ -84,7 +84,6 @@ export class VerifyMigrationConstraint<T extends CometContext> implements Static
             if (migration.actions.verify && !(await isEnacted(migration.actions, ctx.world.deploymentManager, govDeploymentManager))) {
               // XXX we really need to be capturing the block at which the proposal was executed instead of the migration being run
               migrationData.postMigrationBlockNumber = await ctx.world.deploymentManager.hre.ethers.provider.getBlockNumber();
-              console.log(ctx.migrations)
               await migration.actions.verify(ctx.world.deploymentManager, govDeploymentManager, migrationData.preMigrationBlockNumber, migrationData.postMigrationBlockNumber);
               debug(`${label} Verified migration "${migration.name}"`);
             }
--- a/scenario/constraints/MigrationConstraint.ts
+++ b/scenario/constraints/MigrationConstraint.ts
@@ -33,6 +33,7 @@ export class MigrationConstraint<T extends CometContext> implements StaticConstr
       }
       solutions.push(async function (ctx: T): Promise<T> {
         const govDeploymentManager = ctx.world.auxiliaryDeploymentManager || ctx.world.deploymentManager;
+        const governor = await govDeploymentManager.getContractOrThrow('governor');
         const compWhale = (await ctx.getCompWhales())[0];
         const proposer = await impersonateAddress(govDeploymentManager, compWhale, exp(1, 18)); // give them enough ETH to make the proposal
 
@@ -52,7 +53,13 @@ export class MigrationConstraint<T extends CometContext> implements StaticConstr
             debug(`${label} Migration ${migration.name} has already been enacted`);
           } else {
             migrationData.preMigrationBlockNumber = await ctx.world.deploymentManager.hre.ethers.provider.getBlockNumber();
+            const lastProposalBefore = await governor.proposalCount();
             await migration.actions.enact(ctx.world.deploymentManager, govDeploymentManager, artifact);
+            const lastProposalAfter = await governor.proposalCount();
+            // Store the latest proposal id if one was created by this migration
+            if (lastProposalAfter > lastProposalBefore) {
+              migrationData.lastProposal = lastProposalAfter.toNumber();
+            }
             debug(`${label} Enacted migration ${migration.name}`);
           }
         }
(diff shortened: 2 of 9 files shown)


scenario/context/CometContext.ts
```diff
@@ -40,6 +40,11 @@ import { Requirements } from '../constraints/Requirements';
 
 export type ActorMap = { [name: string]: CometActor };
 export type AssetMap = { [name: string]: CometAsset };
+export type MigrationData = {
+  migration: Migration<any>;
+  preMigrationBlockNumber?: number;
+  postMigrationBlockNumber?: number;
+}
```
kevincheng96, line 47: The current approach is to cache the old and new block numbers at which to grab the states.

scenario/constraints/MigrationConstraint.ts
```diff
@@ -77,9 +79,13 @@ export class VerifyMigrationConstraint<T extends CometContext> implements Static
       async function (ctx: T): Promise<T> {
         const govDeploymentManager = ctx.world.auxiliaryDeploymentManager || ctx.world.deploymentManager;
         if (ctx.migrations) {
-          for (const migration of ctx.migrations) {
+          for (const migrationData of ctx.migrations) {
+            const migration = migrationData.migration;
             if (migration.actions.verify && !(await isEnacted(migration.actions, ctx.world.deploymentManager, govDeploymentManager))) {
-              await migration.actions.verify(ctx.world.deploymentManager, govDeploymentManager);
+              // XXX we really need to be capturing the block at which the proposal was executed instead of the migration being run
```
kevincheng96, line 85: This is the tricky part. We actually want to be capturing the block number after the proposal has been executed. Otherwise, the diff might also include state changes from other proposals, not related to this migration, being executed.

plugins/deployment_manager/Migration.ts
```diff
@@ -5,7 +5,7 @@ export interface Actions<T> {
   prepare: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<T>;
   enact: (dm: DeploymentManager, govDm: DeploymentManager, t: T) => Promise<void>;
   enacted?: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<boolean>;
-  verify?: (dm: DeploymentManager, govDm: DeploymentManager) => Promise<void>;
+  verify?: (dm: DeploymentManager, govDm: DeploymentManager, preMigrationBlockNumber: number, postMigrationBlockNumber: number) => Promise<void>;
```
kevincheng96, line 8: Open to other suggestions as well. We could also have `enact` output a new type and add as an argument in `verify`.

jflatow: Oh interesting 🤔

scenario/constraints/MigrationConstraint.ts
jflatow: Can we just pass it from ProposalConstraint?

jflatow: Although actually we should be verifying as of whatever block it is. Should we only be passing the pre block number?

kevincheng96: > Can we just pass it from ProposalConstraint?

Yeah I thought about passing it in from the ProposalConstraint, but we actually don't know what the proposal is for a specific migration unless we output that information from the migration itself.

> Although actually we should be verifying as of whatever block it is. Should we only be passing the pre block number?

Originally that's what I did. But I realized that we should probably be diffing the states of each migration in isolation. Reason being that there could be other proposals that get executed by the ProposalConstraint that would also alter the state, which would cause the `verify()` step for this migration to fail.

jflatow: But in that case, I think you want to run the whole verify check on the post block? You could make the same argument for all the checks it does

kevincheng96: Agreed, currently exploring this.

kevincheng96: I've implemented what we discussed offline. Basically, got rid of the `VerifyMigrationConstraint` and moved the verify step into the `ProposalConstraint`. This allows us to verify each migration right after its proposal is executed. We also keep track of which migrations have been verified already to make sure all migrations are verified, even those that are not tied to a specific proposal.

scenario/constraints/ProposalConstraint.ts
```diff
@@ -65,9 +75,33 @@ export class ProposalConstraint<T extends CometContext> implements StaticConstra
...
+      // Verify all unverified migrations (e.g. ones that are not tied to proposals)
+      if (ctx.migrations) {
+        for (const migrationData of ctx.migrations) {
+          if (migrationData.verified === true) continue;
+          await migrationData.migration.actions.verify(
+            ctx.world.deploymentManager,
+            govDeploymentManager,
+            migrationData.preMigrationBlockNumber
+          );
+          migrationData.verified = true;
```
jflatow, line 104: More complicated, but not too bad all in all

jflatow (approved):
💪

PR #733, opened 2023-04-07
Add token sweeping functions to Arbitrum Bridge Receiver

scott-silver:
This PR adds native token and ERC20 sweeping functions to the Arbitrum bridge receiver per OpenZeppelin's recommendations.

Changes include:
- a `SweepableBridgeReceiver` contract, which exposes `sweepToken` and `sweepNativeToken` functions that can be called by the localTimelock (adapted from the Bulker contract
- ~~a`BridgeReceiverCore` contract that is shared by `BaseBridgeReceiver` and `SweepableBridgeReceiver` (to allow both to use the `localTimelock` variable and the `Unauthorized` custom error)~~
- unit tests, including a SweepableBridgeReceiver harness contract

### DIFF
--- a/contracts/bridges/SweepableBridgeReceiver.sol
+++ b/contracts/bridges/SweepableBridgeReceiver.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+import "../IERC20NonStandard.sol";
+import "./BaseBridgeReceiver.sol";
+
+contract SweepableBridgeReceiver is BaseBridgeReceiver {
+    error FailedToSendNativeToken();
+    error TransferOutFailed();
+
+    /**
+     * @notice A public function to sweep accidental ERC-20 transfers to this contract
+     * @dev Note: Make sure to check that the asset being swept out is not malicious
+     * @param recipient The address that will receive the swept funds
+     * @param asset The address of the ERC-20 token to sweep
+     */
+    function sweepToken(address recipient, address asset) external {
+        if (msg.sender != localTimelock) revert Unauthorized();
+
+        uint256 balance = IERC20NonStandard(asset).balanceOf(address(this));
+        doTransferOut(asset, recipient, balance);
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


contracts/bridges/BridgeReceiverCore.sol
```diff
@@ -0,0 +1,42 @@
...
+    error InvalidProposalId();
+    error InvalidTimelockAdmin();
+    error ProposalNotExecutable();
+    error TransactionAlreadyQueued();
+    error Unauthorized();
+
+    /** Public variables **/
+
+    /// @notice Address of the governing contract that this bridge receiver expects to
+    ///  receive messages from; likely an address from another chain (e.g. mainnet)
```
kevincheng96, line 17: nit: extra space in front of "receive"

contracts/bridges/SweepableBridgeReceiver.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+import "../IERC20NonStandard.sol";
+import "./BridgeReceiverCore.sol";
+
+contract SweepableBridgeReceiver is BridgeReceiverCore {
```
kevincheng96, line 7: What about having the `SweepableBridgeReceiver` extend `BaseBridgeReceiver` directly? That way, the `ArbitrumBridgeReceiver` can just extend `SweepableBridgeReceiver` instead of both `BaseBridgerReceiver, SweepableBridgeReceiver`, which is a bit confusing IMO.

scott-silver: was just thinking something similar, it's an extra contract for no particular reason

contracts/bridges/BridgeReceiverCore.sol
scott-silver: deleted this file

kevincheng96 (approved):
Looks great!

jflatow (approved):

kevincheng96:
This should be ready to merge.

PR #735, opened 2023-04-12
Proposal to transfer more COMP to Polygon

jflatow:
TODO:
 - [x] Fix transfer
 - [x] Finalize amount
 - [x] Proposal text

### DIFF
--- a/deployments/polygon/usdc/migrations/1681313581_more_comp.ts
+++ b/deployments/polygon/usdc/migrations/1681313581_more_comp.ts
@@ -7,7 +7,7 @@ import { expect } from 'chai';
 const ERC20PredicateAddress = '0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf';
 const RootChainManagerAddress = '0xA0c68C638235ee32657e8f720a23ceC1bFc77C77';
 
-export default migration('1661899622_rewards', {
+export default migration('1681313581_more_comp', {
   async prepare(deploymentManager: DeploymentManager) {
     return {};
   },
@@ -51,7 +51,7 @@ export default migration('1661899622_rewards', {
         calldata: depositForCOMPCalldata
       },
     ];
-    const description = "# Refresh Polygon COMP\n\n## Explanation\n\nSince the launch of cUSDCv3 on Polygon mainnet a month ago, the market has grown steadily and proven to be in good order. The initial seeding of rewards only provisioned enough for a few months, as a conservative starting point. Now is a good time to renew the COMP rewards going to the market.\n\nThis proposal bridges an additional 12,500 COMP to sustain the current rewards speeds of the market for an additional year (approximately).\n\n## Proposal\n\nThe proposal itself is to be made from [this pull request](https://github.com/compound-finance/comet/pull/735). \n\nThe first action approves the transfer of COMP by the Polygon bridge contract.\n\nThe second action triggers the transfer across the bridge to the rewards contract.";
+    const description = "# Refresh Polygon COMP\n\nThis is a repost of [Proposal 158](https://compound.finance/governance/proposals/158), given that the proposal failed to reach quorum. More discussions can be found on this [forum thread](https://www.comp.xyz/t/refresh-comp-rewards-on-polygon/4253).\n\n## Explanation\n\nSince the launch of cUSDCv3 on Polygon mainnet a month ago, the market has grown steadily and proven to be in good order. The initial seeding of rewards only provisioned enough for a few months, as a conservative starting point. Now is a good time to renew the COMP rewards going to the market.\n\nThis proposal bridges an additional 12,500 COMP to sustain the current rewards speeds of the market for an additional year (approximately).\n\n## Proposal\n\nThe proposal itself is to be made from [this pull request](https://github.com/compound-finance/comet/pull/735). \n\nThe first action approves the transfer of COMP by the Polygon bridge contract.\n\nThe second action triggers the transfer across the bridge to the rewards contract.";
     const txn = await deploymentManager.retry(
       async () => governor.propose(...await proposal(actions, description))
     );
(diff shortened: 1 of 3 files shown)


deployments/polygon/usdc/migrations/1681313581_more_comp.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+    const COMPAmountToBridge = exp(10_000, 18); // XXX
+    const depositCOMPData = ethers.utils.defaultAbiCoder.encode(['uint256'], [COMPAmountToBridge]);
+    const depositForCOMPCCalldata = ethers.utils.defaultAbiCoder.encode(
+      ['address', 'address', 'bytes'],
+      [rewards.address, COMP.address, depositCOMPData]
+    );
+
+    const actions = [
+      // 4. Approve Polygon's ERC20Predicate to take Timelock's COMP (for bridging)
```
scott-silver, line 41: Should this be `1.` instead of `4.`?

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
```diff
@@ -0,0 +1,78 @@
...
+      rewards,
+    } = await deploymentManager.getContracts();
+
+    const RootChainManager = await deploymentManager.existing(
+      'RootChainManager',
+      RootChainManagerAddress
+    );
+    const COMPAmountToBridge = exp(10_000, 18); // XXX
+    const depositCOMPData = ethers.utils.defaultAbiCoder.encode(['uint256'], [COMPAmountToBridge]);
+    const depositForCOMPCCalldata = ethers.utils.defaultAbiCoder.encode(
```
kevincheng96, line 35: nit: extra `C` in `Calldata`

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
```diff
@@ -0,0 +1,78 @@
...
+    const {
+      rewards,
+    } = await deploymentManager.getContracts();
+
+    // 4. & 5.
+    const polygonCOMP = new Contract(
+      '0x8505b9d2254A7Ae468c0E9dd10Ccea3A837aef5c',
+      ['function balanceOf(address account) external view returns (uint256)'],
+      deploymentManager.hre.ethers.provider
+    );
```
kevincheng96, line 75: You should be able to get COMP directly from `deploymentManager.getContracts()` now that it is part of roots.

kevincheng96 (commented):
Seems like the scenario is not recognizing this as a cross-chain proposal. I think you'll have to add the `RootChainManager` address to the check here: https://github.com/compound-finance/comet/blob/9ce9aa93222ec05d5414df23427f6bebf60622a1/scenario/utils/isBridgeProposal.ts#L16

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
jflatow: Whoops, copied from other proposal

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
jflatow: Yeah I didn't love this but saw we had done this way in the other one, will try

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
jflatow: > Seems like the scenario is not recognizing this as a cross-chain proposal. I think you'll have to add the `RootChainManager` address to the check here:
> 
> https://github.com/compound-finance/comet/blob/9ce9aa93222ec05d5414df23427f6bebf60622a1/scenario/utils/isBridgeProposal.ts#L16

Hm are we even storing this address to check?

jflatow: I clicked reply on Kevin's comment, not sure how it ended up here!

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
kevincheng96: Yeah the reason was because rewards were not turned on yet before that proposal so COMP was not spidered. Another way we could've done it was to spider in the `verify` block (that's what we currently do for the Optimism, which I think is cleaner).

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
kevincheng96: > Hm are we even storing this address to check?

I don't think it's currently stored, but we should probably add it to roots like we did for the other cross-chain entry point contracts. Though that makes me want to re-think where we should be storing these types of addresses, since roots can become bloated over time by contracts not related directly to Comet.

kevincheng96 (approved):
Nice 👍

deployments/polygon/usdc/migrations/1681313581_more_comp.ts
```diff
@@ -0,0 +1,74 @@
+import { Contract } from 'ethers';
+import { DeploymentManager, migration } from '../../../../plugins/deployment_manager';
+import { calldata, exp, proposal } from '../../../../src/deploy';
+
+import { expect } from 'chai';
+
+const ERC20PredicateAddress = '0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf';
+const RootChainManagerAddress = '0xA0c68C638235ee32657e8f720a23ceC1bFc77C77';
+
+export default migration('1681313581_more_comp', {
```
kevincheng96, line 10: For posterity, the reason why the CI job failed to write `enacted` on the first proposal submission was because the migration name did not match the migration file name, so no file was find when trying to write `enacted`.

kevincheng96:
Created [proposal 159](https://compound.finance/governance/proposals/159).

kevincheng96:
Merging this now since the proposal was executed on 5/7.

PR #736, opened 2023-04-13
Emit event when rewards claimed is set by the admin

kevincheng96:
This adds an event for a function that modifies storage.

### DIFF
--- a/contracts/CometRewards.sol
+++ b/contracts/CometRewards.sol
@@ -32,13 +32,13 @@ contract CometRewards {
     /// @notice Rewards claimed per Comet instance and user account
     mapping(address => mapping(address => uint)) public rewardsClaimed;
 
-
     /// @dev The scale for factors
     uint256 internal constant FACTOR_SCALE = 1e18;
 
     /** Custom events **/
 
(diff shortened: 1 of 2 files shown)


contracts/CometRewards.sol
```diff
@@ -32,13 +32,13 @@ contract CometRewards {
...
     mapping(address => mapping(address => uint)) public rewardsClaimed;
 
-
     /// @dev The scale for factors
     uint256 internal constant FACTOR_SCALE = 1e18;
 
     /** Custom events **/
 
     event GovernorTransferred(address indexed oldGovernor, address indexed newGovernor);
+    event RewardsClaimedSet(address indexed user, address indexed comet, uint256 amount);
```
jflatow, line 41: Seems a little inconsistent to emit `comet` for this one but not `RewardsClaimed`

jflatow (approved):

contracts/CometRewards.sol
kevincheng96: Agreed. We could add `comet` to `RewardClaimed`, but unfortunately it won't be indexed because we already have three other indexed fields.

scott-silver (approved):

PR #742, opened 2023-05-02
Adding the Semgrep worflow by Decurity

beched:
This Semgrep workflow runs 2 rule sets over all the pull requests.
The rule sets include both generic Solidity security detections and the comet-specific ones.
We've reviewed the comet repository issues, pull requests and commits and created a few rules according to the fixes that were proposed by OZ/ChainSecurity/Certora and have been accepted by Compound.
Most of the issues have been closed without fixes, and some of others are unfeasible to detect with Semgrep.

You can see it in action here: https://github.com/Decurity/comet/pull/2

Right now if you run these rules over the whole comet codebase, the stats is as follows:
Ran 9 rules on 36 files: 32 findings.

23 of these are missing sanity checks on constructor arguments (each argument = 1 finding). Of course many of them might not require the sanity checks so this rule can spam the alerts if run over the whole codebase but can be useful if run over the new contracts (there's only one constructor, so there won't be overwhelmingly many findings)

Also there's a couple of "A constant name is not in UPPER_CASE like other constant variables." and "Event parameters with type 'address' should be indexed" findings.

### DIFF
--- a/.github/workflows/run-semgrep.yaml
+++ b/.github/workflows/run-semgrep.yaml
@@ -24,17 +24,23 @@ jobs:
     steps:
       # Fetch project source with GitHub Actions Checkout.
       - uses: actions/checkout@v3
-      # Fetch rules
-      - name: Fetch semgrep rules
-        uses: actions/checkout@v2
+      # Fetch generic rules
+      - name: Fetch generic semgrep rules
+        uses: actions/checkout@v3
         with:
           repository: decurity/semgrep-smart-contracts
           path: rules
-      # Run security & performance rules
+      # Fetch Compound rules
+      - name: Fetch Compound semgrep rules
+        uses: actions/checkout@v3
+        with:
+          repository: decurity/compound-semgrep-rules
+          path: compound
+      # Run security, performance & Compound specific rules
       - run: semgrep ci --sarif --output=semgrep.sarif || true
         env:
-           SEMGREP_RULES: rules/solidity/security rules/solidity/performance
-
+           SEMGREP_RULES: rules/solidity/security rules/solidity/performance compound/solidity
+      # Upload findings to GitHub Advanced Security Dashboard
       - name: Upload findings to GitHub Advanced Security Dashboard
         uses: github/codeql-action/upload-sarif@v2
         with:
--- a/.github/workflows/run-semgrep.yaml
+++ b/.github/workflows/run-semgrep.yaml
@@ -1,0 +1,42 @@
+# Name of this GitHub Actions workflow.
+name: Run Semgrep
+
+on:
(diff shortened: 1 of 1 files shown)


kevincheng96:
Looks like this is causing CI to fail. I think you'll need to add env vars to the workflow ([example](https://github.com/compound-finance/comet/blob/485d9ee129d7a6d190e1c071205ec5d929795a80/.github/workflows/run-eslint.yaml#L9-L13)).

Are these semgrep rulesets being constantly maintained? How does one silence a specific rule if it becomes too noisy?

Raz0r:
Hey @kevincheng96, I have added the env vars. As for the maintenance of the rules, yes we constantly enhance the rules both to cover new vulnerabilities and to reduce the false positives of the existing rules. We do not expect the rules to be noisy, but you can dismiss individual alerts from the PR interface as well as mute a particular rule by using an in-line comment like this:`// nosemgrep: compound.solidity.constant-not-in-uppercase` or just `// nosemgrep` (should be placed above the line that triggers the alert). Alternatively, you can disable semgrep scanning for entire folders using `.semgrepignore`: https://semgrep.dev/docs/ignoring-files-folders-code/

kevincheng96:
Looks like CI is still failing to run, but looks like it might be because your branch is on a fork of the Comet repo and the env variables are not set there. Since this PR only adds a new workflow, I think it is safe to merge.

kevincheng96 (approved):

cylon56:
@kevincheng96 given that the changes are approved, is there anything else blocking this from being merged?

kevincheng96:
> @kevincheng96 given that the changes are approved, is there anything else blocking this from being merged?

Nope, I think this is good to merge.

kevincheng96:
> you can dismiss individual alerts from the PR interface

@Raz0r I've dismissed some of the alerts in this [PR](https://github.com/compound-finance/comet/pull/757) but that does not hide the alerts. Is there a way to collapse the alert similar to how we can do so for a Github comment?

Raz0r:
Hey @kevincheng96 according to the [GitHub docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#dismissing--alerts), if you dismiss an alert apparently it will not hide automatically in the current PR for some reason, but the same code won't generate the dismissed alert again in future PRs. Anyway you can click on the alert header to collapse the block or alternatively add `// nosemgrep` to the necessary lines of code.

kevincheng96:
Got it, thanks! Unfortunately I can only temporarily collapse the block. When the page is reloaded, the alerts are expanded again.

PR #748, opened 2023-05-25
misc liquidation bot fixes

scott-silver:
Three fixes

- include a chainId in the transaction (Flashbots has begun validating this field and will reject transactions that fail to include it)
- increase the gas multiplier to be more competitive with our transactions
- connect the Comet instance to the signer so that direct calls on Comet (e.g. `await comet.absorb(signerAddress, targetAddresses)`) will succeed

### DIFF
--- a/scripts/liquidation_bot/index.ts
+++ b/scripts/liquidation_bot/index.ts
@@ -49,7 +49,7 @@ async function main() {
   await dm.spider();
 
   const contracts = await dm.contracts();
-  const comet = contracts.get('comet') as CometInterface;
+  let comet = contracts.get('comet') as CometInterface;
 
   // Flashbots provider requires passing in a standard provider
   let flashbotsProvider: FlashbotsBundleProvider;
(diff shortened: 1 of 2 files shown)


kevincheng96 (approved):
🔥

scott-silver:
One more update.

We've been using the USDC/USDT/.01% Uniswap pool as the flash loan pool for Polygon.

But recently, the liquidity in that pool has started to dip, and our scenarios have started failing when trying to take out a flash loan big enough to swap 400 WETH:

```
[polygon] ❌ LiquidationBot > liquidates an underwater position of $asset0 with no maxAmountToPurchase: Error Error: VM Exception while processing transaction: reverted with reason string 'TF'
```

The biggest Uniswap pool that swaps USDC for an asset that we don't list as collateral is the USDC/BOB/.01% pool. That pool has ~$2.6M USDC available to borrow; the USDC/USDT pool has less than $700K.

I know virtually nothing about the BOB coin, but the Uniswap pool is definitely better.

scott-silver:
Also, updated to read the chainId from the network config instead of the provider; unit tests fail with the following error when reading from provider:

```
'get' on proxy: property '_network' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected '#<Object>' but got '#<Object>')
```

kevincheng96:
lol never heard of BOB token either 🤷

kevincheng96 (approved):

PR #752, opened 2023-05-31
Do not attach signer to comet instance

scott-silver:
In #748, we updated the Comet instance used in the liquidation bot to attach the signer before being passed around as an argument.

https://github.com/compound-finance/comet/pull/748/files#diff-985edbf80fe4c689f86ad3dcd779b4effe7ead300ec1f6dd5068f7b8d47e41eeR90

This approach doesn't work when `USE_FLASHBOTS` is set to `true`. In that case, we set the signer to be an Ethers Wallet, which has the side-effect of removing the Provider from the Comet instance. Subsequent attempts to read values via the provider fail.

### DIFF
--- a/scripts/liquidation_bot/index.ts
+++ b/scripts/liquidation_bot/index.ts
@@ -86,9 +86,6 @@ async function main() {
 
   const signerWithFlashbots = { signer, flashbotsProvider };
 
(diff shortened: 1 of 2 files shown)


kevincheng96 (approved):
Interesting, nice fix 🔨

PR #753, opened 2023-06-01
Use Sleuth to Bulk Liquidity Checks

hayesgm:
This patch uses Sleuth to bulk the liquidity checks on a given network. This should drop us from an O(n) query to an O(1) query. There are a number of other improvements we can make to drop this further, e.g. by using block_number as returned from Sleuth instead of constantly querying the block number; plus we should convert to using StaticJsonRpcProvider as another way to drop our query count. But that all aside, this is the first and biggest improvement we could make.

Note: I'm a bit weary that we might ask for so many addresses' liquidities in one query that it runs out of gas (e.g. max block gas) and then doesn't work. I would propose that we change this to using a partial scan each time (e.g. 100-200 addresses at random) and we would pretty quickly converge on all addresses (that is, allowing us to make multiple smaller queries per block, but allowing this to be somewhat random). If we sliced the pool for each query and reshuffled on each new block, we would be close to guaranteeing that we will check all addresses within one or two blocks.

That is, you basically run:

```
given allAddresses;

let blockNumber = currentBlockNumber;

while(true) {
  let random = allAddresses.shuffle();
  while (random.length > 0) {
    let [blockNumber, checks] = check(random.pop(0, 100)); // Pop 100 addresses and check them
    // do something with checks
    if (blockNumber > currentBlockNumber) {
      currentBlockNumber = blockNumber;
      break; // We're on a new block, reshuffle and start over
    }
    sleep(5000); // Wait a few moments
    // ... and then check more addresses from this shuffle
  }

  // Reshuffle and start over, given we're on a new block
}
```

That's a little different, but basically it should converge on checking all addresses with high probability.

### DIFF
--- a/scripts/liquidation_bot/index.ts
+++ b/scripts/liquidation_bot/index.ts
@@ -13,6 +13,8 @@ import {
 import { FlashbotsBundleProvider } from '@flashbots/ethers-provider-bundle';
 import { Signer, Wallet } from 'ethers';
 import googleCloudLog, { LogSeverity } from './googleCloudLog';
+import { Sleuth } from '@compound-finance/sleuth';
+import * as liquidatableQuerySol from '../../artifacts/contracts/LiquidatableQuery.sol/LiquidatableQuery.json';
 
 const loopDelay = 5000;
 const loopsUntilUpdateAssets = 1000;
@@ -54,6 +56,7 @@ async function main() {
   // Flashbots provider requires passing in a standard provider
   let flashbotsProvider: FlashbotsBundleProvider;
   let signer: Signer;
+
   if (useFlashbots && useFlashbots.toLowerCase() === 'true') {
     // XXX use a designated auth signer
     // `authSigner` is an Ethereum private key that does NOT store funds and is NOT your bot's primary key.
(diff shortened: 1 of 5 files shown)


scott-silver:
This is awesome! 🙌 

Much more efficient, this should greatly cut down on the number of requests we're making.

Running this locally, it seems to be able to handle querying 1600 addresses without running out of gas.

Couldn't you take all the returned addresses, break them up into chunks of no more than 1K addresses, do the query and then return the results? No need to grab a random grouping of addresses, just keep popping off addresses until you've check them all, then move on to the next block?

scott-silver (approved):

scripts/liquidation_bot/index.ts
```diff
@@ -108,19 +120,23 @@ async function main() {
       loops = 0;
     }
 
+    // TODO: Remove this line entirely
     const currentBlockNumber = await hre.ethers.provider.getBlockNumber();
 
-    googleCloudLog(LogSeverity.INFO, `currentBlockNumber: ${currentBlockNumber}`);
+    googleCloudLog(LogSeverity.INFO, `currentBlockNumber: ${currentBlockNumber} [checked=${lastCheckedBlockNumber}]`);
 
     if (currentBlockNumber !== lastBlockNumber) {
```
kevincheng96, line 128: This check should now be checking that `currentBlockNumber != lastCheckedBlockNumber` right?

kevincheng96 (approved):
Nice improvement! Should save a TON of Infura calls. 

Before you merge though, `yarn tsc` seems to be failing.

scott-silver:
looks great! you've got some ESLint errors and some type errors though

PR #737, opened 2023-04-13
Create custom price feed for WBTC

kevincheng96:
This PR creates a custom price feed for WBTC / USD by reading from both the WBTC / BTC and BTC / USD price feeds. This will allow the cUSDCv3 market to get a more accurate price for WBTC.

We also organize the contracts repo by moving all the custom price feed contracts into a separate `pricefeeds` subdirectory.

PR #757, opened 2023-06-08
Multiplicative Price Feed

kevincheng96:
This generalizes the logic used in the recently audited [`WBTCPriceFeed`](https://github.com/compound-finance/comet/pull/737) to create a `MultiplicativePriceFeed` that derives its price from multiplying the prices from two other price feeds together. 

This is a flexible wrapper price feed that can be used to generate prices for any asset that does not have a price feed that fits Comet's expected price denomination. e.g. if we wanted to add `cbETH` to `cUSDCv3`, there is no Chainlink price feed for `cbETH / USD`, so we would need to multiply the `cbETH / ETH` and `ETH / USD` price feeds together to get `cbETH / USD`

### DIFF
--- a/contracts/pricefeeds/MultiplicativePriceFeed.sol
+++ b/contracts/pricefeeds/MultiplicativePriceFeed.sol
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.15;
+
+import "../vendor/@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
+import "../IPriceFeed.sol";
+
+/**
+ * @title Multiplicative price feed
+ * @notice A custom price feed that multiplies the prices from two price feeds and returns the result
+ * @author Compound
+ */
+contract MultiplicativePriceFeed is IPriceFeed {
+    /** Custom errors **/
+    error BadDecimals();
+    error InvalidInt256();
+
+    /// @notice Version of the price feed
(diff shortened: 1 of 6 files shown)


contracts/pricefeeds/MultiplicativePriceFeed.sol
```diff
@@ -0,0 +1,81 @@
...
+        decimals = decimals_;
+        description = description_;
+        priceFeedScale = int256(10 ** decimals);
+    }
+
+    /**
+     * @notice Price for the latest round
+     * @return roundId Round id from price feed B
+     * @return answer Latest price
+     * @return startedAt Timestamp when the round was started; passed on from price feed B
```
scott-silver, line 62: The only weirdness to this approach is that you could call this function twice, get the same "round id" and timestamps, but a different price (if the feed for price A had updated since your last call, but the feed for price B had not)

scott-silver (approved):
nice! 🙌

contracts/pricefeeds/MultiplicativePriceFeed.sol
kevincheng96: Yeah I think it's impossible to avoid any weirdness here. The only thing Comet cares about is the price, so we could eventually just use `latestAnswer()` instead of `latestRoundData()`.

If we eventually want to improve the oracle and check timestamps for staleness, that work would be done in the price feed so Comet really never needs to know about timestamp, round id, etc.

PR #758, opened 2023-06-13
Linea Goerli cUSDCv3 deploy

Julink-eth:
Add changes to support deployment and scenarios for linea goerli network.


deployments/linea-goerli/usdc/configuration.json
```diff
@@ -0,0 +1,37 @@
...
+    "supplyBase": 0,
+    "borrowKink": 0.8,
+    "borrowSlopeLow": 0.035,
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000011574074074074073e15",
+    "baseBorrowSpeed": "0.0011458333333333333e15",
```
kevincheng96, line 23: Ideally, we set these as 0 on deploy and increase them with a cross-chain governance proposal.

deployments/linea-goerli/usdc/configuration.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "assets": {
+    "WETH": {
+      "address": "0x2C1b868d6596a18e32E61B901E4060C872647b6C",
+      "priceFeed": "0x45A61d74eaa5F78a6e1919Ca74Db2615743190Ad",
+      "decimals": "18",
+      "borrowCF": 0.775,
+      "liquidateCF": 0.825,
+      "liquidationFactor": 0.95,
+      "supplyCap": "50_000e18"
```
kevincheng96, line 34: Same here. We set them as 0 on deploy and increase them afterwards.

deployments/linea-goerli/usdc/relations.ts
```diff
@@ -0,0 +1,34 @@
+import baseRelationConfig from '../../relations';
+
+export default {
+  ...baseRelationConfig,
+  governor: {
+    artifact: 'contracts/bridges/linea/LineaBridgeReceiver.sol:LineaBridgeReceiver'
+  },
+  ClonableBeaconProxy: {
+    artifact: 'contracts/ERC20.sol:ERC20'
+  },
```
kevincheng96, line 10: Is this needed? Seems specific to Arbitrum

plugins/import/etherscan.ts
```diff
@@ -75,7 +78,26 @@ export function getEtherscanApiKey(network: string): string {
...
+    if (network === 'linea-goerli') {
+      res = (
+        await axios.get(url, {
+          params: data,
+          auth: {
+            username: 'eslineastage',
+            password: 'QW$eKsy3J%~v'
+          }
+        })
+      )['data'];
```
kevincheng96, line 94: Note: Remember to remove.

tasks/deployment_manager/task.ts
kevincheng96, line 1: Are the diffs in this file all just formatting changes? If so, let's revert the changes.

kevincheng96 (commented):
Looks good overall! Left some minor comments here and there. A lot of the diff is formatting related which I think can be reverted.

deployments/linea-goerli/usdc/relations.ts
Julink-eth: You're right, I remove it.

plugins/import/etherscan.ts
Julink-eth: Good catch thank you, completely forgot about that TODO

Julink-eth:
I just added the changes to fix all the comments you guys had except the one for the migration file that I will in another commit.
If I may add just one comment, all the formatting issues seemed to have occured because of the file .prettierrc.json that specifies  
```
{
  "singleQuote": true,
  "printWidth": 100
}
```

And a lot of files don’t respect those rules, I know it’s not the PR to do this kind of clean up, I just wanted to bring that up :)

Julink-eth:
> One thing that is missing is the migration script to create a cross-chain proposal to initialize Comet (increase supply caps, set COMP speeds, etc.). Here's an [example](https://github.com/compound-finance/comet/blob/main/deployments/base-goerli/usdc/migrations/1679592519_configurate_and_ens.ts).

What values should I put in this file, can I just use the exact same values as what's been done for base-goerli ?

kevincheng96:
> If I may add just one comment, all the formatting issues seemed to have occured because of the file .prettierrc.json that specifies

Yeah, that is our bad. Through this PR, we realized that we should either enforce the Prettier or just remove it. We removed it in https://github.com/compound-finance/comet/pull/760.

> What values should I put in this file, can I just use the exact same values as what's been done for base-goerli ?

Yes, values such as the amount of COMP to bridge could be the same. However, a lot of the values used in the proposal are actually pulled from  the `configuration.json` file when calling [`getConfigurationStruct`](https://github.com/compound-finance/comet/blob/main/deployments/base-goerli/usdc/migrations/1679592519_configurate_and_ens.ts#L53). So it will actually be a 2-step process:

1. Set supply caps and reward speeds to 0 in `configuration.json`. Deploy the contracts with this initial configuration so `Comet` is not live yet.
2. Update the supply caps and reward speeds in `configuration.json`. Run the migration script (which pulls in these updated configuration) to create the on-chain proposal. 

We usually keep the migration script in a separate PR for this reason.

Julink-eth:
> Yes, values such as the amount of COMP to bridge could be the same. However, a lot of the values used in the proposal are actually pulled from the `configuration.json` file when calling [`getConfigurationStruct`](https://github.com/compound-finance/comet/blob/main/deployments/base-goerli/usdc/migrations/1679592519_configurate_and_ens.ts#L53). So it will actually be a 2-step process:
> 
> 1. Set supply caps and reward speeds to 0 in `configuration.json`. Deploy the contracts with this initial configuration so `Comet` is not live yet.
> 2. Update the supply caps and reward speeds in `configuration.json`. Run the migration script (which pulls in these updated configuration) to create the on-chain proposal.
> 
> We usually keep the migration script in a separate PR for this reason.

Thank you very clear, in that case if you can review my last changes taking care of all the comments except the migration that'd be great.
I'll make a new PR for the migration once this one is accepted.

kevincheng96:
> Thank you very clear, in that case if you can review my last changes taking care of all the comments except the migration that'd be great.

Thanks for making the changes. Overall looks good! Left a few comments about some formatting changes, but not urgent and blocking.

Looks like the CI jobs aren't running because your forked repo doesn't have the env vars set up. What we could do is have you merge the PR into a feature branch so we can run the tests using the env vars in our Comet repo. I created the `linea/usdc-deploy` branch and will change the target of this PR to point there instead.

> I'll make a new PR for the migration once this one is accepted.

I don't think we should let the merging of this PR block the migration work. The code to make a proposal can still be executed without being merged into main. In fact, we run the migration and make sure the proposal passes first before merging.

deployments/linea-goerli/usdc/roots.json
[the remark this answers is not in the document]
Julink-eth: Oh my bad I meant to remove the ones in the localhost folder.
We still needed to redo a deployment anyway for the LineaBridgerReceiver change

Julink-eth:
I pushed the latest changes to account for the latest comments.
I have redeployed the contracts, so the roots.json has been updated, also I added one more contract to the relations.ts:
`lineaL1TokenBridge` that will be used in the migration script.
Let me know if there is anything that needs to be changed.

I'm working on a different branch for the migration script.

scenario/BulkerScenario.ts
```diff
@@ -7,7 +7,7 @@ import { exp } from '../test/helpers';
 scenario(
   'Comet#bulker > (non-WETH base) all non-reward actions in one txn',
   {
-    filter: async (ctx) => await isBulkerSupported(ctx) && !matchesDeployment(ctx, [{deployment: 'weth'}, {network: 'mumbai'}]),
+    filter: async (ctx) => await isBulkerSupported(ctx) && !matchesDeployment(ctx, [{deployment: 'weth'}, {network: 'mumbai'}, { network: 'linea-goerli' }]),
```
kevincheng96, line 10: Curious why we need to exclude Linea from this test?

plugins/import/etherscan.ts
kevincheng96: Following up on this again, did you want to remove this before merging?

kevincheng96 (approved):
Thanks for making the changes. Left a final comment to follow up on removing the password. Everything else looks good!

scenario/BulkerScenario.ts
Julink-eth: Otherwise the scenario was failing, I can't remember should I remove it ?

kevincheng96:
Another question: Are the supported assets minted by Linea or bridged from Goerli? If the former, I recommend sending a good amount to the `Fauceteer` so users can claim small amounts of it to play around with the protocol.

scenario/BulkerScenario.ts
kevincheng96: Yeah I think it's better to let the scenario fail so we can see if it's an actual problem or not.

Julink-eth:
https://github.com/compound-finance/comet/pull/758#discussion_r1232495625
It should be removed, this is outdated.

Julink-eth:
> Another question: Are the supported assets minted by Linea or bridged from Goerli? If the former, I recommend sending a good amount to the `Fauceteer` so users can claim small amounts of it to play around with the protocol.

COMP, USDC, ETH should be bridged from goerli to Linea.
In our quests we already five the users faucets to get those assets on Linea :)

PR #759, opened 2023-06-13
Make OptimismBridgeReceiver sweepable

kevincheng96:
This change should extend the `OptimismBridgeReceiver` with some functionality to sweep ERC20 and native tokens.


scott-silver (approved):
👍

PR #766, opened 2023-06-21
Migration script to deploy a custom WBTC price feed and upgrade Comet to use it

kevincheng96:
I also change the `verify` step to take in as an argument the artifact from the `prepare` step

### DIFF
--- a/deployments/mainnet/usdc/migrations/1686950699_update_wbtc_price_feed.ts
+++ b/deployments/mainnet/usdc/migrations/1686950699_update_wbtc_price_feed.ts
@@ -1,0 +1,77 @@
+import { expect } from 'chai';
+import { diffState } from '../../../../plugins/deployment_manager';
+import { DeploymentManager } from '../../../../plugins/deployment_manager/DeploymentManager';
+import { getCometConfig } from '../../../../plugins/deployment_manager/DiffState';
+import { migration } from '../../../../plugins/deployment_manager/Migration';
+import { proposal } from '../../../../src/deploy';
+
+interface Vars {
+  newWBTCPriceFeed: string
+};
+
+export default migration('1686950699_update_wbtc_price_feed', {
+  prepare: async (deploymentManager: DeploymentManager) => {
+    // Deploy custom WBTC price feed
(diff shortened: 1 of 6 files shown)


deployments/mainnet/usdc/migrations/1686950699_update_wbtc_price_feed.ts
```diff
@@ -0,0 +1,77 @@
...
+      },
+
+      // 2. Deploy and upgrade to a new version of Comet
+      {
+        contract: cometAdmin,
+        signature: 'deployAndUpgradeTo(address,address)',
+        args: [configurator.address, comet.address],
+      },
+    ];
+    const description = 'TODO'
```
scott-silver, line 53: don't forget to do this.

also, missing semi-colon

plugins/deployment_manager/MigrationTemplate.ts
```diff
@@ -12,12 +12,16 @@ import { migration } from '../../../../plugins/deployment_manager/Migration';
 interface Vars {};
 
 export default migration('${timestamp}_${name}', {
-  prepare: async (deploymentManager: DeploymentManager) => {
+  async prepare(deploymentManager: DeploymentManager) {
```
scott-silver, line 15: just curious: why change this function from `fnName: async() =>` to `async fnName() {`?

kevincheng96: I noticed a lot of our migration scripts were using `async fnName()` so I changed this to be more consistent. But no preference on my end, so I'm down to revert this change if you want.

kevincheng96:
Contract deployed via txn: https://etherscan.io/tx/0xd516d4e3cc9f7540984829f4f365f417d12a1da974b3741e356365c526c6d731

Github action with artifact: https://github.com/compound-finance/comet/actions/runs/5360714422

deployments/mainnet/usdc/migrations/1686950699_update_wbtc_price_feed.ts
kevincheng96: Updated the description.

PR #769, opened 2023-06-22
Deploy and initialize native USDC market on Arbitrum 

cwang25:
This PR introduces the deployment for USDC Native on Arbitrum as well as the proposal to initialize the market there.

This PR also rename existing `usdc` to `usdc.e` to distinguish from bridged usdc and native usdc

Note: when this PR only has deploy scripts (when migration PR hasn't merged into this PR yet), the scenario will fail as the existing market has already transfer ownership to the timelock, thus the scenario tests will fail for not being able to upgrade the comet. But it will be resolved after migration PR is in place, which will upgrade the contracts via DAO's proposal.) 

Reference: https://github.com/compound-finance/comet/pull/769/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

TODO:
- [x] deploy arbitrum native USDC instance
- [x] run market initialization proposal
- [x] run attestation script to mint native USDC on Arbitrum


deployments/arbitrum/usdc.n/configuration.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "Compound USDC(native)",
+  "symbol": "cUSDCNv3",
+  "baseToken": "USDC",
+  "baseTokenAddress": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831",
+  "baseTokenPriceFeed": "0x50834F3163758fcC1Df9973b6e91f0F0F0434aD3",
```
scott-silver, line 6: This is the same price feed that we're using for the current, bridged version of USDC. Do you know if this price feed will reflect the price of the bridged version or the native version?

deployments/arbitrum/usdc.n/configuration.json
```diff
@@ -0,0 +1,65 @@
...
+  "borrowMin": "100e6",
```
scott-silver, line 7: We can probably crank down the `borrowMin`. Maybe `10e6`?

deployments/arbitrum/usdc.n/configuration.json
```diff
@@ -0,0 +1,65 @@
...
+    "ARB": {
+      "address": "0x912ce59144191c1204e64559fe8253a0e49e6548",
+      "priceFeed": "0xb2A824043730FE05F3DA2efaFa1CBbe83fa548D6",
+      "decimals": "18",
+      "borrowCF": 0.55,
+      "liquidateCF": 0.60,
+      "liquidationFactor": 0.93,
+      "supplyCap": "0e18"
+    },
+    "GMX": {
```
scott-silver, line 37: I wonder if it's worth listing `GMX` on this deployment as well. There's only $72K of GMX supplied in the first market.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,278 @@
...
+      // 4. Update the list of official markets
+      {
+        target: ENSResolverAddress,
+        signature: 'setText(bytes32,string,string)',
+        calldata: ethers.utils.defaultAbiCoder.encode(
+          ['bytes32', 'string', 'string'],
+          [subdomainHash, ENSTextRecordKey, JSON.stringify(officialMarketsJSON)]
+        )
+      },
+      // 5. Displace v2 USDT COMP rewards
```
scott-silver, line 160: USDT v2 COMP speeds are already set to 0, yeah? We shouldn't need to set it to 0 again

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,278 @@
...
+    const arbitrumChainId = (await deploymentManager.hre.ethers.provider.getNetwork()).chainId.toString();
+    const newMarketObject = { baseSymbol: 'USDC.n', cometAddress: comet.address };
+    const officialMarketsJSON = JSON.parse(await ENSResolver.text(subdomainHash, ENSTextRecordKey));
+    if (officialMarketsJSON[arbitrumChainId]) {
+      officialMarketsJSON[arbitrumChainId].push(newMarketObject);
+    } else {
+      officialMarketsJSON[arbitrumChainId] = [newMarketObject];
+    }
+
+    const mainnetActions = [
```
scott-silver, line 110: Are we going to seed this deployment with USDC reserves?

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,278 @@
...
+        signature: '_setCompSpeeds(address[],uint256[],uint256[])',
+        args: [
+          [cUSDTAddress],
+          [0],
+          [0],
+        ],
+      },
+    ];
+
+    const description = "# Configurate Arbitrum cUSDCNv3 market, set reward config, USDC native and COMP, and update ENS text record.";
```
scott-silver, line 172: We should flesh out this description

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,278 @@
...
+        ],
+        value: createRetryableTicketGasParams.deposit
+      },
+      // 2. Approve the COMP gateway to take Timelock's COMP for bridging
+      {
+        contract: COMP,
+        signature: 'approve(address,uint256)',
+        args: [compGatewayAddress, COMPAmountToBridge]
+      },
+      // 3. Bridge COMP from mainnet to Arbitrum rewards
```
scott-silver, line 133: Do we need to bridge more COMP to Arbitrum?

Seems like we should just turn the reward speed on the first market to 0.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,278 @@
...
+      trace(await governor.propose(...(await proposal(mainnetActions, description))))
+    );
+
+    const event = txn.events.find(event => event.event === 'ProposalCreated');
+    const [proposalId] = event.args;
+
+    trace(`Created proposal ${proposalId}.`);
+  },
+
+  // async enacted(deploymentManager: DeploymentManager): Promise<boolean> {
```
scott-silver, line 183: You can just delete this, not comment it out

scenario/LiquidationBotScenario.ts
```diff
@@ -122,6 +123,16 @@ for (let i = 0; i < MAX_ASSETS; i++) {
         ' == 500',
         // WBTC
         ' == 50'
+      ], 
+      'usdc.n': [
```
scott-silver, line 127: Why are these values lower than the values for `arbitrum/usdc`? The liquidity for Arbitrum GMX, WETH and WBTC should be identical

scenario/LiquidationBotScenario.ts
```diff
@@ -337,6 +359,16 @@ for (let i = 0; i < MAX_ASSETS; i++) {
         exp(500, 18),
         // WBTC
         exp(50, 8),
+      ], 
+      'usdc.n': [
+        // ARB
+        exp(10000, 18),
```
scott-silver, line 365: Same question here as above. Why lower?

src/deploy/index.ts
```diff
@@ -133,4 +134,4 @@ export async function proposal(actions: ProposalAction[], description: string):
     }
   }
   return [targets, values, signatures, calldatas, description];
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
cwang25: Ah yeah, this was brought up from  /c3-api indentation setting.  😅 
🧇 brought up adding new line at the end of the files are the correct style via: https://thoughtbot.com/blog/no-newline-at-end-of-file

Too easy to mixed up the styles for VSC lol, since both are typescript, which format setting will always apply one setting that's active on typescript.

Let me change the setting back to default (no new line at the end), and re-push the files to remove those new line.

deployments/arbitrum/usdc.n/configuration.json
cwang25: Yeah unfortunately, I didn't see a new pricefeed from USDC native on Chainlink yet :( 
https://docs.chain.link/data-feeds/price-feeds/addresses?network=arbitrum
I can update to the latest pricefeed once Chainlink has added the USDC native pricefeed. 


Part of me is a bit unsure that if it's ok to use the same pricefeed since those USDC.e and USDC Native are in a sense kind of interchangeable?  (USDC.e bridge back to source chain and burn, then can always mint USDC Native at Arbitrum 1:1).

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: I see.  Yeah will remove it.  

I was a bit confused about why there is v2 rewards in arbitrum migration script. 😅

deployments/arbitrum/usdc.n/configuration.json
cwang25: 👍  sounds good.

deployments/arbitrum/usdc.n/configuration.json
cwang25: Ok let me remove it.  (This is mainly borrowed from original USDC, was thinking the market would be identical between two type of USDC

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: I see now.  

That make sense, we should just need to do it once and don't necessary do ti again for USDC native as long as reward contract has enough rewards token.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Ok 👍

scenario/LiquidationBotScenario.ts
cwang25: Let me tweak those number again. 

Yeah main reason to set those lower was due to the not having enough liquidity in the pool for USDC native now, and when liquidator bot try to borrow too much, it will fail. 

There is one scenario seems let liquidator bot to borrow as much as possible (with no max purchase setting), that one will fail if test user can borrow too big, and when liquidator bot attempts to borrow enough USDC to cover that debt.

deployments/arbitrum/usdc.n/configuration.json
scott-silver: I meant that more as an open question, whether it's worth adding it again

scenario/LiquidationBotScenario.ts
scott-silver: You should be able to borrow more if you switch to a `flashLoanPool` with greater liquidity. See: https://github.com/compound-finance/comet/pull/769#discussion_r1239026099

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Hmm not sure what happened.  
But after removing those unneeded and seems innocent steps, scenario liquidator bot started failing again :(

scenario/LiquidationBotScenario.ts
cwang25: Thanks after using the new pool suggested in the comment, I'm able to bump the volume back up without getting 'TF' error.

scripts/liquidation_bot/liquidateUnderwaterBorrowers.ts
```diff
@@ -102,7 +103,11 @@ export const flashLoanPools = {
     usdc: {
       tokenAddress: addresses.arbitrum.USDT,
       poolFee: 100
-    }
+    }, 
+    'usdc.n': {
```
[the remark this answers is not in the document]
cwang25: Thanks 👍  for finding the another pool with higher liquidity.

scenario/LiquidationBotScenario.ts
cwang25: Bumped back to the same as original usdc, the new liquidity pool and handle such volume flash loan 👍

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: I think yes, but we need to figure out how to work with CCTP without relying on their attestation api.

deployments/arbitrum/usdc.n/configuration.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "Compound USDC(native)",
+  "symbol": "cUSDCNv3",
```
kevincheng96, line 3: I think we should just use `cUSDCv3`. Could be slighlty confusing to have two `cUSDCv3`'s, but we can potentially upgrade the old one to be `cUSDC.ev3`. Though I don't think that's high priority.

deployments/arbitrum/usdc.n/configuration.json
kevincheng96: Interesting to think about because they are different in some sense. But I think the existing `USDC / USD` price feed makes more sense for native USDC than USDC.e since it is the canonical USDC.

deployments/arbitrum/usdc.n/configuration.json
```diff
@@ -0,0 +1,65 @@
...
+  "baseToken": "USDC",
+  "baseTokenAddress": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831",
+  "baseTokenPriceFeed": "0x50834F3163758fcC1Df9973b6e91f0F0F0434aD3",
+  "borrowMin": "10e6",
```
kevincheng96, line 7: Thoughts on dropping this even lower to something like 1e6?

deployments/arbitrum/usdc.n/configuration.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "Compound USDC(native)",
```
kevincheng96, line 2: Not sure if we should put `native` in the name.

deployments/arbitrum/usdc.n/configuration.json
kevincheng96: IMO this is supposed to be a replacement of the USDC.e market so we should keep the configuration mostly the same.

deployments/arbitrum/usdc.n/relations.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export default {
+  ...baseRelationConfig,
+  governor: {
+    artifact: 'contracts/bridges/arbitrum/ArbitrumBridgeReceiver.sol:ArbitrumBridgeReceiver'
+  },
+  ClonableBeaconProxy: {
+    artifact: 'contracts/ERC20.sol:ERC20'
+  },
+  // USDC native
```
kevincheng96, line 11: nit: Native USDC

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,216 @@
...
+
+const ENSName = 'compound-community-licenses.eth';
+const ENSResolverAddress = '0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41';
+const ENSSubdomainLabel = 'v3-additional-grants';
+const ENSSubdomain = `${ENSSubdomainLabel}.${ENSName}`;
+const ENSTextRecordKey = 'v3-official-markets';
+
+const arbitrumCOMPAddress = '0x354A6dA3fcde098F8389cad84b0182725c6C91dE';
+
+const cUSDTAddress = '0xf650c3d88d12db855b8bf7d11be6c55a4e07dcc9';
```
kevincheng96, line 17: Can remove both of these.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
kevincheng96: Yes we should. Discussed with Hans offline on the approach

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,216 @@
...
+        to: bridgeReceiver.address,
+        data: l2ProposalData
+      },
+      deploymentManager
+    );
+
+    const ENSResolver = await govDeploymentManager.existing('ENSResolver', ENSResolverAddress);
+    const subdomainHash = ethers.utils.namehash(ENSSubdomain);
+    const arbitrumChainId = (await deploymentManager.hre.ethers.provider.getNetwork()).chainId.toString();
+    const newMarketObject = { baseSymbol: 'USDC.n', cometAddress: comet.address };
```
kevincheng96, line 88: I think this market should be `USDC` and the old market changed to `USDC.e`.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,216 @@
...
+        target: ENSResolverAddress,
+        signature: 'setText(bytes32,string,string)',
+        calldata: ethers.utils.defaultAbiCoder.encode(
+          ['bytes32', 'string', 'string'],
+          [subdomainHash, ENSTextRecordKey, JSON.stringify(officialMarketsJSON)]
+        )
+      },
+    ];
+
+    const description = "# Configurate Arbitrum cUSDCv3 market for USDC native, and set ENS record for official markets";
```
kevincheng96, line 125: Please add a TODO to update this.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,216 @@
...
+    );
+    const deployAndUpgradeToCalldata = utils.defaultAbiCoder.encode(
+      ['address', 'address'],
+      [configurator.address, comet.address]
+    );
+    const setRewardConfigCalldata = utils.defaultAbiCoder.encode(
+      ['address', 'address'],
+      [comet.address, arbitrumCOMPAddress]
+    );
+    const l2ProposalData = utils.defaultAbiCoder.encode(
```
kevincheng96, line 62: We'll also want to turn off the speed on the USDC.e market.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -0,0 +1,216 @@
...
+          cometAddress: comet.address,
+        }
+      ],
+    });
+
+    // 3.
+    expect(await comptrollerV2.compBorrowSpeeds(cUSDTAddress)).to.be.equal(0);
+    expect(await comptrollerV2.compSupplySpeeds(cUSDTAddress)).to.be.equal(0);
+    expect(await comet.baseTrackingSupplySpeed()).to.be.equal(exp(34.74 / 86400, 15, 18));
+    expect(await comet.baseTrackingBorrowSpeed()).to.be.equal(0);
```
kevincheng96, line 214: Looks outdated

kevincheng96:
Great work so far 👏

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Solved it,  by adding existing rewards contract to deploy script. :)

deployments/arbitrum/usdc.n/configuration.json
cwang25: That make sense, I can go ahead to to rename `usdc` -> `usdc.e` and rename this native from `usdc.n` -> `usdc`

deployments/arbitrum/usdc.n/configuration.json
cwang25: I have removed it :) 
(Thinking in follow-up PR will add (bridged) to the original market)

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Removed `cUSDT` one, but I think 
```
const setRewardConfigCalldata = utils.defaultAbiCoder.encode(
      ['address', 'address'],
      [comet.address, arbitrumCOMPAddress]
    );
```
still uses arbitrumCOMPAddress for reward config.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Have added porposal actions to turn off the bridged USDC.e market in migration as well.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: added 
```
 expect(await bridgedComet.baseTrackingSupplySpeed()).to.be.equal(0);
 expect(await bridgedComet.baseTrackingBorrowSpeed()).to.be.equal(0);
```
👍

deployments/arbitrum/usdc.n/deploy.ts
```diff
@@ -17,20 +17,28 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
   const GMX = await deploymentManager.existing('GMX', '0xfc5a1a6eb076a2c7ad06ed22c90d7e710e35ad0a', 'arbitrum');
   const WETH = await deploymentManager.existing('WETH', '0x82af49447d8a07e3bd95bd0d56f35241523fbab1', 'arbitrum');
   const WBTC = await deploymentManager.existing('WBTC', '0x2f2a2543b76a4166549f7aab2e75bef0aefc5b0f', 'arbitrum');
+  const arbitrumCCTPTokenMinter = await deploymentManager.existing('arbitrumCCTPTokenMinter', '0xE7Ed1fa7f45D05C508232aa32649D89b73b8bA48', 'arbitrum');
```
kevincheng96, line 20: I don't think this belongs in the deploy script and roots. It's only used by scenarios, so can you import it in `relayArbitrumMessage.ts` instead?

deployments/arbitrum/usdc.n/deploy.ts
```diff
@@ -17,20 +17,28 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
...
+  const cometAdmin = await deploymentManager.fromDep('cometAdmin', 'arbitrum', 'usdc');
+  const cometFactory = await deploymentManager.fromDep('cometFactory', 'arbitrum', 'usdc');
+  const $configuratorImpl = await deploymentManager.fromDep('configurator:implementation', 'arbitrum', 'usdc');
+  const configurator = await deploymentManager.fromDep('configurator', 'arbitrum', 'usdc');
+  const rewards = await deploymentManager.fromDep('rewards', 'arbitrum', 'usdc');
   const bulker = await deploymentManager.fromDep('bulker', 'arbitrum', 'usdc');
   const localTimelock = await deploymentManager.fromDep('timelock', 'arbitrum', 'usdc');
   const bridgeReceiver = await deploymentManager.fromDep('bridgeReceiver', 'arbitrum', 'usdc');
-  const rewards = await deploymentManager.fromDep('rewards', 'arbitrum', 'usdc');
+  const bridgedComet = await deploymentManager.fromDep('bridgedComet', 'arbitrum', 'usdc', 'comet');
```
kevincheng96, line 31: Likewise, don't think this should be in the deploy script and be a root. It can be explicitly imported in the migration script that needs it.

scenario/LiquidationBotScenario.ts
kevincheng96: Marking this as unresolved again since the liquidity in these tests is marked down again.

deployments/mainnet/usdc/roots.json
```diff
@@ -6,5 +6,7 @@
     "bulker": "0xa397a8C2086C554B531c02E29f3291c9704B00c7",
     "fxRoot": "0xfe5e5D361b2ad62c541bAb87C45a0B9B018389a2",
     "arbitrumInbox": "0x4Dbd4fc535Ac27206064B68FfCf827b0A60BAB3f",
-    "arbitrumL1GatewayRouter": "0x72Ce9c846789fdB6fC1f34aC4AD25Dd9ef7031ef"
+    "arbitrumL1GatewayRouter": "0x72Ce9c846789fdB6fC1f34aC4AD25Dd9ef7031ef", 
+    "mainnetCCTPTokenMessenger": "0xbd3fa81b58ba92a82136038b25adec7066af3155", 
+    "mainnetCCTPMessageTransmitter": "0x0a992d191deec32afe36203ad87d7d289a738f81"
```
kevincheng96, line 11: nit: don't think we need to prepend with `mainnet` to the names here since the this is the roots for mainnet.

deployments/arbitrum/usdc/roots.json
```diff
@@ -3,5 +3,7 @@
     "configurator": "0xb21b06D71c75973babdE35b49fFDAc3F82Ad3775",
     "rewards": "0x88730d254A2f7e6AC8388c3198aFd694bA9f7fae",
     "bridgeReceiver": "0x42480C37B249e33aABaf4c22B20235656bd38068",
-    "bulker": "0xbdE8F31D2DdDA895264e27DD990faB3DC87b372d"
+    "bulker": "0xbdE8F31D2DdDA895264e27DD990faB3DC87b372d",
+    "arbitrumCCTPTokenMinter": "0xE7Ed1fa7f45D05C508232aa32649D89b73b8bA48",
+    "arbitrumCCTPMessageTransmitter": "0xC30362313FBBA5cf9163F0bb16a0e01f01A896ca"
```
kevincheng96, line 8: nit: don't think we need to prepend with `arbitrum` to the names here since the this is the roots for arbitrum.

kevincheng96: Also I don't think these need to be roots on the Arbitrum deploy.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -201,16 +242,17 @@ export default migration('1686953660_configurate_and_ens', {
           cometAddress: '0xA5EDBDD9646f8dFF606d7448e414884C7d905dCA',
         }, 
         {
-          baseSymbol: 'USDC.n',
+          baseSymbol: 'USDC',
```
kevincheng96, line 245: The `baseSymbol` for the old Comet on line 241 should be updated to `USDC.e.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
```diff
@@ -26,50 +24,77 @@ export default migration('1686953660_configurate_and_ens', {
...
       ['address', 'address'],
       [configurator.address, comet.address]
     );
+    
     const setRewardConfigCalldata = utils.defaultAbiCoder.encode(
       ['address', 'address'],
       [comet.address, arbitrumCOMPAddress]
     );
+
+    const turnOffBridgeCometSupplySpeedCalldata = utils.defaultAbiCoder.encode(
```
kevincheng96, line 68: Could we call it `BridgedUSDCComet` or `USDCEComet`?

scenario/utils/relayArbitrumMessage.ts
```diff
@@ -110,4 +110,114 @@ export default async function relayArbitrumMessage(
...
+    end = start + length.bytes32;
+    // rawMsgBody messageSender, skip won't use
+
+    return {
+      recipient: mintRecipient,
+      amount: amount,
+      sourceDomain: msgSourceDomain,
+      burnToken: burnToken
+    }
+  });
```
kevincheng96, line 197: Nice! Maybe extract this into a separate function?

kevincheng96:
Good progress on the proposal and CCTP message relaying!

scenario/utils/relayArbitrumMessage.ts
cwang25: Ok 👍

deployments/arbitrum/usdc/roots.json
cwang25: Removed 👍

deployments/mainnet/usdc/roots.json
cwang25: Removed 👍

deployments/arbitrum/usdc.n/deploy.ts
cwang25: Moved to migration script 👍

deployments/arbitrum/usdc.n/deploy.ts
cwang25: True,  moved to relay script.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Ok I will now on call it as usdceComet :) 
Will go ahead and rename all instances to be usdcecomet 👍

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Renamed to USDC.e. 
Not sure if I did this right by just renaming it in Native USDC migration script: 
```
    // Rename old USDC market into USDC.e
    officialMarketsJSON[arbitrumChainId][0].baseSymbol = 'USDC.e';
```

scenario/LiquidationBotScenario.ts
cwang25: Will leave it as unresolved. But as the time of testing today, the liquidity is enough again :P 
So reverted those number back up as the same as bridged version

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
cwang25: Curious what would be the process to draft out the description?
Should we go though some external review before putting down a official description?

Was thinking to put this as description 
```
# Configurate Arbitrum cUSDCv3 market for Native USDC native, and set ENS record for official markets
```
But not sure if this is good or not 😛

kevincheng96 (commented):
I think this looks ready! 💪 

Lets port over the changes to the testnet branch and run the proposal on testnet. Let me know when that branch is ready.

deployments/arbitrum/usdc.n/migrations/1686953660_configurate_and_ens.ts
kevincheng96: We can use this for testnet, but for mainnet, we'll start a doc to draft out a proper description.

deployments/arbitrum/usdc/deploy.ts
```diff
@@ -11,62 +11,30 @@ export default async function deploy(deploymentManager: DeploymentManager, deplo
...
-    async () => {
-      trace(`Initializing BridgeReceiver`);
-      await bridgeReceiver.initialize(
-        MAINNET_TIMELOCK,     // govTimelock
-        localTimelock.address // localTimelock
-      );
-      trace(`BridgeReceiver initialized`);
-    }
-  );
+  // Import shared contracts from cUSDCv3
```
kevincheng96, line 21:
```suggestion
  // Import shared contracts from the USDC.e market
```

kevincheng96 (approved):
Looks good! This should be ready to be deployed once we see there's no big issues with scenarios.

cwang25:
⛑️  and I deployed comet to arbitrum via Deploy Market (https://github.com/compound-finance/comet/actions/runs/5884462049/job/15959206776) 

The commit 427bbf5 is commited by the workflow.

PR #801, opened 2023-08-03
Deploy and proposal for USDT market on Goerli

cwang25:
This PR introduces the deployment for USDT Goerli as well as the proposal to initialize the market there.

TODO:

- [ ]  deploy USDT instance to Goerli

- [ ] run market initialization proposal

### DIFF
--- a/deployments/goerli/usdt/configuration.json
+++ b/deployments/goerli/usdt/configuration.json
@@ -1,0 +1,55 @@
+{
+  "name": "Compound USDT",
+  "symbol": "cUSDTv3",
+  "baseToken": "USDT",
+  "baseTokenAddress": "0xfad6367E97217cC51b4cd838Cc086831f81d38C2",
+  "baseTokenPriceFeed": "0xAb5c49580294Aff77670F839ea425f5b78ab3Ae7",
+  "borrowMin": "10e6",
+  "governor": "0x8Fa336EB4bF58Cfc508dEA1B0aeC7336f55B1399",
+  "pauseGuardian": "0x8Fa336EB4bF58Cfc508dEA1B0aeC7336f55B1399",
+  "storeFrontPriceFactor": 0.5,
+  "targetReserves": "5000e18",
+  "rates": {
+    "supplyKink": 0.9,
+    "supplySlopeLow": 0.01690681444,
+    "supplySlopeHigh": 0.6066567706,
+    "supplyBase": 0,
+    "borrowKink": 0.9,
+    "borrowSlopeLow": 0.05171500002,
+    "borrowSlopeHigh": 0.5171500339,
+    "borrowBase": 0.009945209674
+  },
+  "rewardToken": "COMP",
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "447916666666e0",
+    "baseBorrowSpeed": "0e15",
+    "baseMinForRewards": "1000e18"
+  },
+  "assets": {
+    "COMP": {
+      "priceFeed": "0x54a06047087927D9B0fb21c1cf0ebd792764dDB8",
+      "decimals": "18",
+      "borrowCF": 0.65,
+      "liquidateCF": 0.7,
+      "liquidationFactor": 0.92,
+      "supplyCap": "500000e18"
+    },
+    "WBTC": {
+      "priceFeed": "0xA39434A63A52E749F02807ae27335515BA4b07F7",
+      "decimals": "8",
+      "borrowCF": 0.7,
+      "liquidateCF": 0.75,
+      "liquidationFactor": 0.93,
+      "supplyCap": "35000e8"
+    },
+    "WETH": {
+      "priceFeed": "0xD4a33860578De61DBAbDc8BFdb98FD742fA7028e",
+      "decimals": "18",
+      "borrowCF": 0.82,
+      "liquidateCF": 0.85,
+      "liquidationFactor": 0.93,
+      "supplyCap": "1000000e18"
+    }
+  }
+}
--- a/deployments/goerli/usdt/deploy.ts
+++ b/deployments/goerli/usdt/deploy.ts
@@ -1,0 +1,40 @@
+import { Deployed, DeploymentManager } from '../../../plugins/deployment_manager';
+import { debug, DeploySpec, deployComet, exp, sameAddress, wait } from '../../../src/deploy';
+
+const clone = {
+  cbETHImpl: '0x31724cA0C982A31fbb5C57f4217AB585271fc9a5',
+  cbETHProxy: '0xBe9895146f7AF43049ca1c1AE358B0541Ea49704',
+};
+
+export default async function deploy(deploymentManager: DeploymentManager, deploySpec: DeploySpec): Promise<Deployed> {
+  const deployed = await deployContracts(deploymentManager, deploySpec);
+  return deployed;
+}
+
+async function deployContracts(deploymentManager: DeploymentManager, deploySpec: DeploySpec): Promise<Deployed> {
+  const ethers = deploymentManager.hre.ethers;
+  const signer = await deploymentManager.getSigner();
+
+  // Declare existing assets as aliases
+  const USDT = await deploymentManager.existing('USDT', '0xfad6367E97217cC51b4cd838Cc086831f81d38C2', 'goerli');
+  const COMP = await deploymentManager.existing('COMP', '0x3587b2F7E0E2D6166d6C14230e7Fe160252B0ba4', 'goerli');
+  const WBTC = await deploymentManager.existing('WBTC', '0xAAD4992D949f9214458594dF92B44165Fb84dC19', 'goerli');
+  const WETH = await deploymentManager.existing('WETH', '0x42a71137C09AE83D8d05974960fd607d40033499', 'goerli');
+
+  // Import shared contracts from cUSDCv3
+  const cometAdmin = await deploymentManager.fromDep('cometAdmin', 'goerli', 'usdc');
+  const cometFactory = await deploymentManager.fromDep('cometFactory', 'goerli', 'usdc');
+  const $configuratorImpl = await deploymentManager.fromDep('configurator:implementation', 'goerli', 'usdc');
+  const configurator = await deploymentManager.fromDep('configurator', 'goerli', 'usdc');
+  const rewards = await deploymentManager.fromDep('rewards', 'goerli', 'usdc');
+  const fauceteer = await deploymentManager.fromDep('fauceteer', 'goerli', 'usdc');
+  const fxRoot = await deploymentManager.fromDep('fxRoot', 'goerli', 'usdc');
+  const bulker = await deploymentManager.fromDep('bulker', 'goerli', 'usdc');
+  const timelock = await deploymentManager.fromDep('timelock', 'goerli', 'usdc');
+
+  // Deploy all Comet-related contracts
+  const deployed = await deployComet(deploymentManager, deploySpec);
+  const { comet } = deployed;
+
+  return { ...deployed, bulker, fauceteer, fxRoot };
+}
--- a/deployments/goerli/usdt/relations.ts
+++ b/deployments/goerli/usdt/relations.ts
@@ -1,0 +1,12 @@
+import baseRelationConfig from '../../relations';
+
+export default {
+  ...baseRelationConfig,
+  fxRoot: {
+    relations: {
+      stateSender: {
+        field: async fxRoot => fxRoot.stateSender()
+      }
+    }
+  },
+};
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -20,6 +20,7 @@ import './tasks/scenario/task.ts';
 import relationConfigMap from './deployments/relations';
 import goerliRelationConfigMap from './deployments/goerli/usdc/relations';
 import goerliWethRelationConfigMap from './deployments/goerli/weth/relations';
+import goerliUsdtRelationConfigMap from './deployments/goerli/usdt/relations';
 import mumbaiRelationConfigMap from './deployments/mumbai/usdc/relations';
 import mainnetRelationConfigMap from './deployments/mainnet/usdc/relations';
 import mainnetWethRelationConfigMap from './deployments/mainnet/weth/relations';
@@ -269,7 +270,8 @@ const config: HardhatUserConfig = {
     networks: {
       goerli: {
         usdc: goerliRelationConfigMap,
-        weth: goerliWethRelationConfigMap
+        weth: goerliWethRelationConfigMap, 
+        usdt: goerliUsdtRelationConfigMap,
       },
       mumbai: {
         usdc: mumbaiRelationConfigMap
@@ -330,6 +332,11 @@ const config: HardhatUserConfig = {
         name: 'goerli-weth',
         network: 'goerli',
         deployment: 'weth',
+      },
+      {
+        name: 'goerli-usdt',
+        network: 'goerli',
+        deployment: 'usdt',
       },
       {
         name: 'mumbai',


cwang25:
The scenario will fail as we encountered back in Arbitrum native usdc deploy. 
Since this is deploying another comet on the chain with existing governor contract, without proper setFactory, setConfigurator on migration script, the scenario tests will run into errors. 

But this can be confirmed in migration PR, which with migration script the scenario run will succeed.

contracts/Comet.sol
cwang25, line 1:
```
 ···································|··············|·················
 |  CometFactory                    ·      24.044  ·                │
 ···································|··············|·················
 |  CometHarness                    ·      20.606  ·                │
 ···································|··············|·················
 |  CometMath                       ·       0.062  ·                │
 ···································|··············|·················
 |  CometModified                   ·      18.977  ·                │
 ···································|··············|·················
 |  CometModifiedFactory            ·      24.182  ·                │
 ···································|··············|·················
```

The contract size is approaching to limit, compiler will complain with `Warning: 2 contracts exceed the size limit for mainnet deployment.`
But the actual size limit is 24.576 kb, so we should be fine :P

contracts/Comet.sol
```diff
@@ -760,18 +760,49 @@ contract Comet is CometMainInterface {
...
+                case 32 {                      // This is a compliant ERC-20
+                    returndatacopy(0, 0, 32)
+                    success := mload(0)        // Set `success = returndata` of override external call
+                }
+                default {                      // This is an excessively non-compliant ERC-20, revert.
+                    revert(0, 0)
+                }
+        }
         if (!success) revert TransferInFailed();
+        return IERC20NonStandard(asset).balanceOf(address(this)) - preTransferBalance;
```
kevincheng96, line 784: This return value doesn't seem to be used anywhere right now, meaning fee tokens are not properly handled. You'll want to use the returned amount for accounting.

contracts/Comet.sol
```diff
@@ -760,18 +760,49 @@ contract Comet is CometMainInterface {
...
     }
 
     /**
      * @dev Safe ERC20 transfer out
      */
```
kevincheng96, line 789: Can you also add the dev note here:

`     * @dev Note: Safely handles non-standard ERC-20 tokens that do not return a value. See here: https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca
`

scenario/SupplyScenario.ts
```diff
@@ -136,6 +137,42 @@ scenario(
...
+  {
+    tokenBalances: {
+      albert: { $base: 1000 }, // in units of asset, not wei
+    },
+    filter: async (ctx) => matchesDeployment(ctx, [{ deployment: 'usdt' }])
+  },
+  async ({ comet, actors }, context, world) => {
+    // Set fees for USDT for testing
+    const USDTAdmin = await world.deploymentManager.getSigner();
+    const USDT = await world.deploymentManager.existing('USDT', await comet.baseToken(), 'goerli');
```
kevincheng96, line 151: You can use `world.base.network` to programmatically fetch the network instead of hardcoding it.

scenario/SupplyScenario.ts
```diff
@@ -136,6 +137,42 @@ scenario(
...
+    await USDT.connect(USDTAdmin).setParams(10, 10);
+
+    const { albert } = actors;
+    const baseAssetAddress = await comet.baseToken();
+    const baseAsset = context.getAssetByAddress(baseAssetAddress);
+    const scale = (await comet.baseScale()).toBigInt();
+
+    expect(await baseAsset.balanceOf(albert.address)).to.be.equal(1000n * scale);
+
+    // Albert supplies 100 units of base to Comet
```
kevincheng96, line 162:
```suggestion
    // Albert supplies 1000 units of base to Comet
```

scenario/SupplyScenario.ts
```diff
@@ -167,6 +204,45 @@ scenario(
...
+    },
+    cometBalances: {
+      albert: { $base: -1000 } // in units of asset, not wei
+    },
+    filter: async (ctx) => matchesDeployment(ctx, [{ deployment: 'usdt' }]),
+  },
+  async ({ comet, actors }, context, world) => {
+    // Set fees for USDT for testing
+    const USDTAdmin = await world.deploymentManager.getSigner();
+    const USDT = await world.deploymentManager.existing('USDT', await comet.baseToken(), 'goerli');
```
kevincheng96, line 221: Same here

scenario/SupplyScenario.ts
```diff
@@ -167,6 +204,45 @@ scenario(
...
+    const { albert } = actors;
+    const baseAssetAddress = await comet.baseToken();
+    const baseAsset = context.getAssetByAddress(baseAssetAddress);
+    const scale = (await comet.baseScale()).toBigInt();
+    const utilization = await comet.getUtilization();
+    const borrowRate = (await comet.getBorrowRate(utilization)).toBigInt();
+
+    expectApproximately(await albert.getCometBaseBalance(), -1000n * scale, getInterest(1000n * scale, borrowRate, 1n) + 1n);
+
+    // Albert repays 100 units of base borrow
```
kevincheng96, line 234:
```suggestion
    // Albert repays 1000 units of base borrow
```

scenario/SupplyScenario.ts
```diff
@@ -167,6 +204,45 @@ scenario(
...
+    await baseAsset.approve(albert, comet.address);
+    const txn = await albert.supplyAsset({ asset: baseAsset.address, amount: 1000n * scale });
+
+    // XXX all these timings are crazy
+    // Expect to have -1000000, due to token fee, alber only repay 999 USDT instead of 1000 USDT, thus alber still owe 1 USDT which is 1000000
+    expectApproximately(await albert.getCometBaseBalance(), -1000000n, getInterest(1000n * scale, borrowRate, 4n) + 2n);
```
kevincheng96, line 240: Can you use `exp(1, 6)` instead? The number of 0s is hard to read

kevincheng96, line 240: Do you need `expectApproximately`? Does exact expect work?

scenario/SupplyScenario.ts
```diff
@@ -200,6 +276,46 @@ scenario(
...
+    const baseAsset = context.getAssetByAddress(baseAssetAddress);
+    const scale = (await comet.baseScale()).toBigInt();
+
+    expect(await baseAsset.balanceOf(albert.address)).to.be.equal(1000n * scale);
+    expect(await comet.balanceOf(betty.address)).to.be.equal(0n);
+
+    await baseAsset.approve(albert, comet.address);
+    await albert.allow(betty, true);
+
+    // Betty supplies 100 units of base from Albert
```
kevincheng96, line 305:
```suggestion
    // Betty supplies 1000 units of base from Albert
```

kevincheng96 (commented):
Looks great! Could you also add unit tests for `supply base`, `supply collateral`, and `buy collateral`? I notice there is no scenarios for `buy collateral`, so would be good to get some unit test coverage at least.

scenario/SupplyScenario.ts
cwang25: Oh i see. Thank you so much! 😄

cwang25: Replaced all with world.base.network

scenario/SupplyScenario.ts
cwang25: Changed to exp(1,6) but since it's negative so I added  `-1n *`

cwang25: I think in test cases without fee token also has to use expectApproximately() 🤔 
I tried use exact will get 2wei diff.

contracts/Comet.sol
cwang25: added to doTransferOut

deployments/goerli/usdt/configuration.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "Compound USDT",
+  "symbol": "cUSDTv3",
+  "baseToken": "USDT",
+  "baseTokenPriceFeed": "0xAb5c49580294Aff77670F839ea425f5b78ab3Ae7",
+  "borrowMin": "10e6",
```
kevincheng96, line 6: Let's make this 1e0

deployments/goerli/usdt/configuration.json
```diff
@@ -0,0 +1,54 @@
...
+    "borrowSlopeLow": 0.035,
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "rewardToken": "COMP",
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000402083333333e15",
+    "baseBorrowSpeed": "0.000402083333333e15",
+    "baseMinForRewards": "10000e6"
```
kevincheng96, line 26: Let's lower this to 500.

deployments/goerli/usdt/configuration.json
```diff
@@ -0,0 +1,54 @@
...
+    "borrowKink": 0.8,
+    "borrowSlopeLow": 0.035,
+    "borrowSlopeHigh": 0.25,
+    "borrowBase": 0.015
+  },
+  "rewardToken": "COMP",
+  "tracking": {
+    "indexScale": "1e15",
+    "baseSupplySpeed": "0.000402083333333e15",
+    "baseBorrowSpeed": "0.000402083333333e15",
```
kevincheng96, line 25: Should be 0 right?

deployments/goerli/usdt/deploy.ts
```diff
@@ -0,0 +1,51 @@
...
+  // Send some forked USDT to timelock
+  await deploymentManager.idempotent(
+    async () => await USDT.connect(signer).balanceOf(timelock.address) == 0,
+    async () => {
+      trace(`Sending USDC to timelock`);
+      await USDT.connect(signer).transfer(
+        timelock.address,
+        exp(50_000_000, 6),
+      );
+      trace(`Sent USDC to timelock completed`);
```
kevincheng96, line 43: Should say USDT instead of USDC

deployments/goerli/usdt/deploy.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  // Import shared contracts from cUSDCv3
+  const cometAdmin = await deploymentManager.fromDep('cometAdmin', 'goerli', 'usdc');
+  // Purposely don't use the factory because Comet implementation changed.
+  // const cometFactory = await deploymentManager.fromDep('cometFactory', 'goerli', 'usdc');
+  const $configuratorImpl = await deploymentManager.fromDep('configurator:implementation', 'goerli', 'usdc');
+  const configurator = await deploymentManager.fromDep('configurator', 'goerli', 'usdc');
+  const rewards = await deploymentManager.fromDep('rewards', 'goerli', 'usdc');
+  const fauceteer = await deploymentManager.fromDep('fauceteer', 'goerli', 'usdc');
+  const fxRoot = await deploymentManager.fromDep('fxRoot', 'goerli', 'usdc');
```
kevincheng96, line 29: I don't think you need this, it's only for Polygon

deployments/goerli/usdt/relations.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export default {
+  ...baseRelationConfig,
+  fxRoot: {
+    relations: {
+      stateSender: {
+        field: async fxRoot => fxRoot.stateSender()
+      }
+    }
+  },
```
kevincheng96, line 11: Same here

deployments/goerli/usdt/deploy.ts
cwang25: Fixed.

deployments/goerli/usdt/deploy.ts
cwang25: removed fxRoot.

deployments/goerli/usdt/relations.ts
cwang25: removed :)

deployments/goerli/usdt/configuration.json
cwang25: set to 0. :)

deployments/goerli/usdt/configuration.json
cwang25: 500e6 👌

deployments/goerli/usdt/configuration.json
cwang25: set to 1e0 1wei.

cwang25:
> Looks great! Could you also add unit tests for `supply base`, `supply collateral`, and `buy collateral`? I notice there is no scenarios for `buy collateral`, so would be good to get some unit test coverage at least.

I have pushed the unit test for all 3 cases :)

contracts/test/NonStandardFaucetFeeToken.sol
```diff
@@ -0,0 +1,102 @@
...
+ *  See https://github.com/ethereum/EIPs/issues/20
+ * @dev With USDT fee token mechanism
+ * @dev Note: `transfer` and `transferFrom` do not return a boolean
+ */
+contract NonStandardFeeToken {
+    string public name;
+    string public symbol;
+    uint8 public decimals;
+    uint256 public totalSupply;
+    mapping (address => mapping (address => uint256)) public allowance;
```
kevincheng96, line 16:
```suggestion
    mapping(address => mapping (address => uint256)) public allowance;
```

contracts/test/NonStandardFaucetFeeToken.sol
```diff
@@ -0,0 +1,102 @@
...
+pragma solidity 0.8.15;
+
+/**
+ * @title Non-standard ERC20 token
+ * @dev Implementation of the basic standard token.
+ *  See https://github.com/ethereum/EIPs/issues/20
+ * @dev With USDT fee token mechanism
+ * @dev Note: `transfer` and `transferFrom` do not return a boolean
+ */
+contract NonStandardFeeToken {
```
kevincheng96, line 11: Can you make this extend the `IERC20NonStandard` just to make sure the interfaces match?

contracts/test/NonStandardFaucetFeeToken.sol
```diff
@@ -0,0 +1,102 @@
...
+            balanceOf[address(this)] = balanceOf[address(this)] + fee;
+        }
+
+        allowance[src][msg.sender] = allowance[src][msg.sender] - amount;
+        balanceOf[src] = balanceOf[src] - amount;
+        balanceOf[dst] = balanceOf[dst] + sendAmount;
+        emit Transfer(src, dst, sendAmount);
+    }
+
+    function approve(address _spender, uint256 amount) external returns (bool) {
```
kevincheng96, line 72: This returns a `bool` even though the non-standard version doesn't.

test/supply-test.ts
```diff
@@ -405,8 +405,127 @@ describe('supplyTo', function () {
...
+
+  it('supplies collateral the correct amount in a fee-like situation', async () => {
+    const assets = defaultAssets();
+    // Add FeeToken Collateral to assets on top of default assets
+    assets['FeeToken'] = {
+      initial: 1e8,
+      decimals: 18,
+      factory: (await ethers.getContractFactory('NonStandardFaucetFeeToken')) as NonStandardFaucetFeeToken__factory,
+    };
+
```
kevincheng96, line 480: nit: extra line

contracts/Comet.sol
kevincheng96, line 1: For withdraw, should the `Withdraw` event emit the original amount or the amount after fees?

kevincheng96: I think probably just the original amount.

test/buy-collateral-test.ts
```diff
@@ -323,8 +323,89 @@ describe('buyCollateral', function () {
...
+        from: alice.address,
+        to: comet.address,
+        amount: exp(49.5, 6),
+      }
+    });
+    expect(event(txn, 1)).to.be.deep.equal({
+      Transfer: {
+        from: comet.address,
+        to: alice.address,
+        amount: 54450000000000000000n,
```
kevincheng96, line 398: You can use `exp` here.

test/buy-collateral-test.ts
```diff
@@ -323,8 +323,89 @@ describe('buyCollateral', function () {
...
+      }
+    });
+    expect(event(txn, 2)).to.be.deep.equal({
+      BuyCollateral: {
+        buyer: alice.address,
+        asset: COMP.address,
+        baseAmount: exp(49.5, 6),
+        collateralAmount: 55000000000000000000n,
```
kevincheng96, line 406: Same here.

test/buy-collateral-test.ts
```diff
@@ -323,8 +323,89 @@ describe('buyCollateral', function () {
...
+    const txn = await wait(cometAsA.buyCollateral(COMP.address, exp(50, 18), 50e6, alice.address));
+    const p1 = await portfolio(protocol, alice.address);
+    const r1 = await comet.getReserves();
+
+    expect(r0).to.be.equal(0n);
+    expect(r0).to.be.lt(await comet.targetReserves());
+    expect(p0.internal).to.be.deep.equal({ USDT: 0n, COMP: 0n });
+    expect(p0.external).to.be.deep.equal({ USDT: exp(100, 6), COMP: 0n });
+    expect(p1.internal).to.be.deep.equal({ USDT: 0n, COMP: 0n });
+    expect(p1.external).to.be.deep.equal({ USDT: exp(50, 6), COMP: 54450000000000000000n });
```
kevincheng96, line 385: `exp` here as well

kevincheng96 (commented):
Nice work on the tests! They look great 👍

contracts/test/NonStandardFaucetFeeToken.sol
cwang25: np 👍

contracts/test/NonStandardFaucetFeeToken.sol
cwang25: oh yeah,  I have now removed the return bool statement :)

test/supply-test.ts
cwang25: removed 👍

contracts/Comet.sol
cwang25: Yeah this probably boils down to UX.
If 1% fee, user deposit 1000 token to contract, user will see balance show as 999.  (In a sense contract receives only 999 token).  If user try to withdraw 999 from contract, user only receive 998.1 🤔  (In a sense contract gives up 999 token)
If based on contract's perspective, then I think original amount make more sense to me.

Also, I just tried to add another balance compare in transferout to return calculated amount. 
(contract price might go up to... 24.218 and  24.355 still lower than the absolute max limit)

contracts/test/NonStandardFaucetFeeToken.sol
```diff
@@ -1,14 +1,15 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.15;
 
+import "../IERC20NonStandard.sol";
```
kevincheng96, line 4: nit: can you add an extra line below

kevincheng96 (approved):
Think this should be ready for audit with OZ now!

Could you update the PR description to say this introduces a new implementation of Comet that supports non-standard ERC20s and fee tokens? 

Separately, I wonder if we should add this logic in a new `Comet` file, like `NonStandardComet` or `FeeTokenComet`?

kevincheng96:
Looking at the re-entrancy test fixes, I think the better fix is actually just to have the `transfer` and `transferFrom` functions in the `EvilToken` do actual transfer accounting while also calling `performAttack()`.

cwang25:
Synced with ⛑️  offline. 
For the re-entrancy issue that Comet is not immune to the tokens that's got exploited. (i.e. EvilToken.sol) 
We decided to not adding additional layer to prevent that in Comet as this vulnerability will require governance to be exploited that governance has to vote and add the suspicious token into Comet's contract. If suspicious token contract is not added to Comet's market, this should not be a concern. 


For reference, if import re-entrancy guard from openzeppelin which can prevent exploited token contract to execute re-entrancy attack, but it will also increase the contract size to: 
```
 ···································|··············|·················
 |  CometFactory                    ·      24.173  ·                │
 ···································|··············|·················
 |  CometModifiedFactory            ·      24.311  ·                │
 ···································|··············|·················
 ```

File: contracts/ReentrancyGuard.sol [hunk 23]
```diff
@@ -0,0 +1,77 @@
...
+
+    // The values being non-zero value makes deployment a bit more expensive,
+    // but in exchange the refund on every call to nonReentrant will be lower in
+    // amount. Since refunds are capped to a percentage of the total
+    // transaction's gas, it is best to keep them low in cases like this one, to
+    // increase the likelihood of the full refund coming into effect.
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = 2;
+
+    uint256 private _status;
```
kevincheng96, line 37: Does this clash with existing storage variables if we exist an existing Comet to it? We need to be very careful about introducing new storage slots

cwang25: Ah yea it does. :( 
I just removed this import. 
I added those field following the CometStorage.sol. 
```
Add const STATE append to the end of the consts in CometCore's constant field
Add state variable append to the end of the variables in CometStorage
Add customer error append to the end of the custom errors in CometMainInterface
Add modifier functions in Comet.sol
```

Hope this will resolve the clash issue.

cwang25: Discussed offline, we will not add the reentrancyguard to comet.

test/supply-test.ts
```diff
@@ -436,10 +554,11 @@ describe('supplyTo', function () {
     await EVIL.setAttack(attack);
 
     await comet.connect(alice).allow(EVIL.address, true);
-
+    await wait(EVIL.connect(alice).approve(comet.address, 75e6));
+    await EVIL.allocateTo(alice.address, 75e6);
     await expect(
       comet.connect(alice).supplyTo(bob.address, EVIL.address, 75e6)
-    ).to.be.revertedWith("custom error 'SupplyCapExceeded()'");
+    ).to.be.revertedWith("custom error 'TransferInFailed()'");
```
cwang25, line 561: Now the accounting in comet won't be inflated by re-entrancy supply reaching SupplyCapExceeded(). 
It will be blocked and throw TransferInFailed() error.

test/withdraw-test.ts
```diff
@@ -511,7 +511,7 @@ describe('withdraw', function () {
       // in callback, EVIL token calls transferFrom(alice.address, bob.address, 1e6)
       await expect(
         comet.connect(alice).withdraw(EVIL.address, 1e6)
-      ).to.be.revertedWith("custom error 'NotCollateralized()'");
+      ).to.be.revertedWith("custom error 'TransferOutFailed()'");
```
cwang25, line 514: Now the accounting in comet won't be mistaken by re-entrancy withdraw reaching NotColalteralized() error.
Instead, it will be blocked and throw TransferInFailed() error right away.

contracts/Comet.sol
```diff
@@ -201,6 +201,32 @@ contract Comet is CometMainInterface {
         (asset14_a, asset14_b) = getPackedAssetInternal(config.assetConfigs, 14);
     }
 
+    modifier nonReentrant() {
+        _nonReentrantBefore();
+        _;
+        _nonReentrantAfter();
```
kevincheng96, line 207: Let's inline these

contracts/Comet.sol
```diff
@@ -201,6 +201,32 @@ contract Comet is CometMainInterface {
...
+    }
+
+    function _nonReentrantBefore() private {
+        bytes32 slot = COMET_REENTRANCY_GUARD_FLAG_SLOT;
+        uint256 status;
+        assembly {
+            status := sload(slot)
+        }
+
+        if(status == COMET_REENTRANCY_GUARD_ENTERED) revert ReentrantCallBlocked();
```
kevincheng96, line 217: nit: space after `if`

contracts/CometCore.sol
```diff
@@ -56,6 +56,13 @@ abstract contract CometCore is CometConfiguration, CometStorage, CometMath {
     /// @dev The scale for factors
     uint64 internal constant FACTOR_SCALE = 1e18;
 
+    /// @dev The storage slot for reentrancy guard flags
+    bytes32 internal constant COMET_REENTRANCY_GUARD_FLAG_SLOT = bytes32(keccak256("comet.reentrancy.guard"));
+
+    /// @dev The reentrancy guard status
+    uint256 internal constant COMET_REENTRANCY_GUARD_NOT_ENTERED = 1;
+    uint256 internal constant COMET_REENTRANCY_GUARD_ENTERED = 2;
+
```
kevincheng96, line 65: Can just inline rather than creating constants. Also, let's use a default value of 0. Otherwise, you need to initialize the slot to 1.

contracts/Comet.sol
```diff
@@ -776,7 +802,7 @@ contract Comet is CometMainInterface {
      * @dev Safe ERC20 transfer out
      * @dev Note: Safely handles non-standard ERC-20 tokens that do not return a value. See here: https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca
      */
-    function doTransferOut(address asset, address to, uint amount) internal {
+    function doTransferOut(address asset, address to, uint amount) nonReentrant internal {
```
kevincheng96, line 805: Having `nonReentrant` on the transfer functions breaks `BuyCollateral` right?

test/buy-collateral-test.ts
```diff
@@ -543,8 +543,8 @@ describe('buyCollateral', function () {
       expect(normalTotalsBasic.trackingSupplyIndex).to.equal(evilTotalsBasic.trackingSupplyIndex);
       expect(normalTotalsBasic.trackingBorrowIndex).to.equal(evilTotalsBasic.trackingBorrowIndex);
       expect(normalTotalsBasic.totalSupplyBase).to.equal(1e6);
-      // EvilToken attack is using 3000 EVIL tokens, so totalSupplyBase should have 3000e6
-      expect(evilTotalsBasic.totalSupplyBase).to.equal(3000e6);
+      // EvilToken attack should be blocked
+      expect(evilTotalsBasic.totalSupplyBase).to.equal(0);
```
kevincheng96, line 547: Let's add an assertion to assert the re-entrancy error.

test/supply-test.ts
```diff
@@ -558,47 +558,7 @@ describe('supplyTo', function () {
...
-    ).to.be.revertedWith("custom error 'SupplyCapExceeded()'");
-  });
-
-  // This test is purposely test the edge case where comet internal accounting will be incorrect when EvilToken re-entrancy attack
-  // is performed. The attack will cause comet to have inflated supply amount of token. 
-  //
-  // We decided that this won't be a concern for Comet, as in order to pull off this attack, the governance has to propose and vote to 
-  // add suspicious token in to Comet's market. As long as governance doesn't add suspicious token contract or erc-777 token to market, the 
-  // Comet should not be vulnerable to this type of attack.
-  it('incorrect accounting via re-entrancy', async () => {
```
kevincheng96: Instead of removing this test, let's re-purpose it to show that re-entrancy is no longer possible.

test/withdraw-test.ts
```diff
@@ -558,7 +558,7 @@ describe('withdraw', function () {
       // in callback, EvilToken attempts to withdraw USDC to bob's address
       await expect(
         comet.connect(alice).withdraw(EVIL.address, 1e6)
-      ).to.be.revertedWith("custom error 'NotCollateralized()'");
+      ).to.be.revertedWith("custom error 'TransferOutFailed()'");
```
kevincheng96, line 561: Test name should change as well

test/withdraw-test.ts
kevincheng96: Shouldn't it throw the re-entrancy error instead?

contracts/Comet.sol
cwang25: ok will remove the modifier function, and inline the functions into TransferIn and TransferOut.

contracts/CometCore.sol
cwang25: I recalled seeing in openzeppelin's reentrancy guard contract saying that using 1, 2 is cheaper on running than setting it to 0.  (reference: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/ReentrancyGuard.sol) 

Since we're only checking on if status == COMET_REENTRANCY_GUARD_ENTERED.  It looks like the initialization value of 0 or 1 wouldn't affect the first run behavior, so I ended up changing to 1,2.  And each run it only checks on if the status is ENTERED(2) :P.   But yeah I agree the first initialization value having to be 0 instead of 1 is a bit confusing, and I can change it back to 0 if the gas saving gain is not enough to tradeoff with the confusion initialization values.

test/supply-test.ts
cwang25: Sure no problem. 
The reason I removed it, it's because I found other old tests (that assumed re-entrancy can be done by exceeding supply cap) sort of already re-purposed and covered the block of re-entrancy already.

test/withdraw-test.ts
cwang25: Yeah it was bit weird,  but since the error is thrown in the nested transfer() function, so it gets sort of "re-wrap" into TransferInFailed() as the code here doesn't re-raise the error (or should we make it re-raise the error?) 
```
uint256 preTransferBalance = ERC20(asset).balanceOf(address(this));
        (bool success, bytes memory returndata) = asset.call(abi.encodeWithSelector(ERC20.transferFrom.selector, from, address(this), amount));
        if (!success || !(returndata.length == 0 || abi.decode(returndata, (bool)))) {
            revert TransferInFailed();
        }
        return ERC20(asset).balanceOf(address(this)) - preTransferBalance;
```

From code we can see now the transferFrom revert won't reveirt right away with returndata, but instead when it fails, we throw `TransferInFailed()`.

PR #805, opened 2023-08-10
Skip liquidation bot scenarios

kevincheng96:
These scenarios are noisy and don't really give us anything actionable to do at this point. Monitoring on-chain liquidity should be done outside of scenarios.


scott-silver (approved):
👍

PR #807, opened 2023-08-14
fix: Spider fix

Arveymenon:
Ran all test cases.
Attaching spider related test cases results


kevincheng96:
Hi @Arveymenon, thanks for taking on this fix. The proposed fix doesn't make Spider case-insensitive. I think the proper fix is to convert both addresses to lowercase before matching them.

PR #817, opened 2023-10-27
Scroll: Remove basefee, ensure scenario tests all run, add migration file

isabellewei:


scenario/utils/relayScrollMessage.ts
```diff
@@ -49,7 +49,7 @@ export default async function relayScrollMessage(
 
     const aliasAccount = await impersonateAddress(
       bridgeDeploymentManager,
-      applyL1ToL2Alias(scrollMessenger.address)
+      "0xD69c917c7F1C0a724A51c189B4A8F4F8C8E8cA0a"
```
kevincheng96, line 52: What address is this? Ideally, this will be pulled in from the deployment manager rather than hard-coded, since this file will be used for Scroll mainnet as well.

isabellewei: For our Goerli testnet, we are using an EOA for this, which is why it's hardcoded, but for mainnet `applyL1ToL2Alias(scrollMessenger.address)` will be accurate

deployments/scroll-goerli/usdc/migrations/1694594095_configurate_and_ens.ts
```diff
@@ -0,0 +1,232 @@
...
+    const config = await rewards.rewardConfig(comet.address);
+    expect(config.token).to.be.equal(COMP.address);
+    expect(config.rescaleFactor).to.be.equal(exp(1, 12));
+    expect(config.shouldUpscale).to.be.equal(true);
+
+    // 2. & 3.
+    expect(await COMP.balanceOf(rewards.address)).to.be.equal(exp(1_000, 18));
+
+    // 4 & 5,
+    expect(await USDC.balanceOf(rewards.address)).to.be.equal(exp(100_000, 6));
```
kevincheng96, line 167: This should be removed. If you run scenarios locally, I believe it'll fail here.

deployments/scroll-goerli/usdc/migrations/1694594095_configurate_and_ens.ts
```diff
@@ -0,0 +1,232 @@
...
+
+    // 6.
```
kevincheng96, line 169: nit: 4

PR #818, opened 2023-11-08
Address OZ's feedback on USDT comet support

cwang25:


test/buy-collateral-test.ts
```diff
@@ -543,8 +543,8 @@ describe('buyCollateral', function () {
       expect(normalTotalsBasic.trackingSupplyIndex).to.equal(evilTotalsBasic.trackingSupplyIndex);
       expect(normalTotalsBasic.trackingBorrowIndex).to.equal(evilTotalsBasic.trackingBorrowIndex);
       expect(normalTotalsBasic.totalSupplyBase).to.equal(1e6);
-      // EvilToken attack is using 3000 EVIL tokens, so totalSupplyBase should have 3000e6
-      expect(evilTotalsBasic.totalSupplyBase).to.equal(3000e6);
+      // EvilToken attack should be blocked
+      expect(evilTotalsBasic.totalSupplyBase).to.equal(0);
```
cwang25, line 547: Try to expect on error thrown, but somehow seems tricky when the test has pauses Automining and resume after actions are appended 🤔 

Is there a way to catch the error on the failed actions when append during automining is pause, and after automining is resumed?

contracts/Comet.sol
```diff
@@ -765,22 +765,10 @@ contract Comet is CometMainInterface {
...
-                }
-                case 32 {                      // This is a compliant ERC-20
-                    returndatacopy(0, 0, 32)
-                    success := mload(0)        // Set `success = returndata` of override external call
-                }
-                default {                      // This is an excessively non-compliant ERC-20, revert.
-                    revert(0, 0)
-                }
+        (bool success, bytes memory returndata) = asset.call(abi.encodeWithSelector(IERC20NonStandard.transferFrom.selector, from, address(this), amount));
+        if (!success || !(returndata.length == 0 || abi.decode(returndata, (bool)))) {
```
kevincheng96, line 790: I think `(returndata.length != 0 && !abi.decode(returndata, (bool)))` is easier to understand than `!(returndata.length == 0 || abi.decode(returndata, (bool)))`. Same for `doTransferOut`

contracts/Comet.sol
```diff
@@ -1262,14 +1262,15 @@ contract Comet is CometMainInterface {
...
+     * @dev Note: For USDT, if there is non-zero prior allowance, it must be reset to 0 first before setting a new value in proposal
      * @param asset The asset that the manager will gain approval of
      * @param manager The account which will be allowed or disallowed
      * @param amount The amount of an asset to approve
      */
     function approveThis(address manager, address asset, uint amount) override external {
         if (msg.sender != governor) revert Unauthorized();
 
-        IERC20NonStandard(asset).approve(manager, amount);
+        ERC20(asset).approve(manager, amount);
```
kevincheng96, line 1297: Do we have tests for this? I feel this would revert right since we're not using a low-level call right?

contracts/Comet.sol
```diff
@@ -764,47 +784,27 @@ contract Comet is CometMainInterface {
...
-                    success := not(0)          // set success to true
-                }
-                case 32 {                      // This is a compliant ERC-20
-                    returndatacopy(0, 0, 32)
-                    success := mload(0)        // Set `success = returndata` of override external call
-                }
-                default {                      // This is an excessively non-compliant ERC-20, revert.
-                    revert(0, 0)
-                }
+        _nonReentrantBefore();
```
kevincheng96, line 787: Sorry, I meant inline the `_nonReentrant()` logic inside the `nonReentrant` modifier and then keep the modifier on these functions.

contracts/Comet.sol
kevincheng96: Looks like scenarios are not running properly in CI, but this test should probably fail: `Comet#approveThis > allows governor to authorize and rescind authorization for non-Comet ERC20`

contracts/CometCore.sol
```diff
@@ -56,6 +56,13 @@ abstract contract CometCore is CometConfiguration, CometStorage, CometMath {
     /// @dev The scale for factors
     uint64 internal constant FACTOR_SCALE = 1e18;
 
+    /// @dev The storage slot for reentrancy guard flags
+    bytes32 internal constant COMET_REENTRANCY_GUARD_FLAG_SLOT = bytes32(keccak256("comet.reentrancy.guard"));
+
+    /// @dev The reentrancy guard status
+    uint256 internal constant COMET_REENTRANCY_GUARD_NOT_ENTERED = 0;
+    uint256 internal constant COMET_REENTRANCY_GUARD_ENTERED = 2;
```
kevincheng96, line 64: nit: 1

kevincheng96, line 64: I wonder if defining these as constants rather than inlining has an effect on bytecode size. I think not, but would be good to verify.

contracts/IERC20NonStandard.sol
```diff
@@ -7,9 +7,37 @@ pragma solidity 0.8.15;
  *  See https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca
  */
 interface IERC20NonStandard {
```
kevincheng96, line 9: Looks like this is not used anymore. I did leave a comment suggesting we may need this again, but if not, we should remove this file.

test/buy-collateral-test.ts
kevincheng96: Hmm `expect.to.be.reverted` doesn't work? Does it throw an error?

test/buy-collateral-test.ts
```diff
@@ -409,7 +409,7 @@ describe('buyCollateral', function () {
   });
 
   describe('reentrancy', function() {
-    it('is not broken by reentrancy supply ', async () => {
+    it('block reentrancy supply ', async () => {
```
kevincheng96, line 412:
```suggestion
    it('is blocked during reentrant supply', async () => {
```

test/supply-test.ts
```diff
@@ -558,47 +558,7 @@ describe('supplyTo', function () {
...
-    });
-    await EVIL.setAttack(attack);
-
-    await comet.connect(alice).allow(EVIL.address, true);
-    await wait(EVIL.connect(alice).approve(comet.address, 75e6));
-    await EVIL.allocateTo(alice.address, 75e6);
-    await comet.connect(alice).supplyTo(bob.address, EVIL.address, 75e6);
-    // Re-entrancy attack 2 loops, expect to have 2X accounting balance 75*2 = 150
-    expect(await comet.collateralBalanceOf(bob.address, EVIL.address)).to.be.equal(150e6);
+    ).to.be.revertedWith("custom error 'TransferInFailed()'");
```
kevincheng96, line 561: Reminder for me to look closer at this and see why it doesn't return the re-entrancy error instead.

test/supply-test.ts
```diff
@@ -537,7 +537,7 @@ describe('supplyTo', function () {
           decimals: 6,
           initialPrice: 2,
           factory: await ethers.getContractFactory('EvilToken') as EvilToken__factory,
-          supplyCap: 100e6
+          supplyCap: 250e6
```
kevincheng96, line 540: Can you change this back to 100e6? This changes the behavior of the test

test/supply-test.ts
```diff
@@ -524,10 +524,10 @@ describe('supplyTo', function () {
     expect(q1.external).to.be.deep.equal({ USDC: 0n, COMP: 0n, WETH: 0n, WBTC: 0n, FeeToken: 0n });
     expect(t1.totalSupplyAsset).to.be.equal(t0.totalSupplyAsset.add(1998e8));
     // Fee Token logics will cost a bit more gas than standard ERC20 token with no fee calculation
-    expect(Number(s0.receipt.gasUsed)).to.be.lessThan(163000);
+    expect(Number(s0.receipt.gasUsed)).to.be.lessThan(186000);
   });
 
-  it('prevents exceeding the supply cap via re-entrancy', async () => {
+  it('re-entrancy block', async () => {
```
kevincheng96, line 530:
```suggestion
  it('blocks reentrancy from exceeding the supply cap', async () => {
```

test/supply-test.ts
```diff
@@ -702,4 +662,37 @@ describe('supplyFrom', function () {
...
+      amount: 75e6,
+      maxCalls: 1
+    });
+    await EVIL.setAttack(attack);
+
+    await comet.connect(alice).allow(EVIL.address, true);
+    await wait(EVIL.connect(alice).approve(comet.address, 75e6));
+    await EVIL.allocateTo(alice.address, 75e6);
+    await comet.connect(alice).supplyTo(bob.address, EVIL.address, 75e6);
+    expect(await comet.collateralBalanceOf(bob.address, EVIL.address)).to.be.revertedWith("custom error 'TransferInFailed()'");
```
kevincheng96, line 696: `collateralBalanceOf`? Doesn't seem right

test/supply-test.ts
```diff
@@ -702,4 +662,37 @@ describe('supplyFrom', function () {
     await wait(cometAsB.allow(charlie.address, true));
     await expect(cometAsC.supplyFrom(bob.address, alice.address, COMP.address, 7)).to.be.revertedWith("custom error 'Paused()'");
   });
+
+  it('block re-entrancy', async () => {
```
kevincheng96, line 666: This test seems to be an exact copy of the other reentrancy test.

test/withdraw-test.ts
```diff
@@ -475,7 +475,7 @@ describe('withdraw', function () {
   });
 
   describe('reentrancy', function () {
-    it('is not broken by malicious reentrancy transferFrom', async () => {
+    it('block malicious reentrancy transferFrom', async () => {
```
kevincheng96, line 478:
```suggestion
    it('blocks malicious reentrant transferFrom', async () => {
```

test/withdraw-test.ts
```diff
@@ -521,7 +521,7 @@ describe('withdraw', function () {
       expect(await USDC.balanceOf(bob.address)).to.eq(0);
     });
 
-    it('is not broken by malicious reentrancy withdrawFrom', async () => {
+    it('block malicious reentrancy withdrawFrom', async () => {
```
kevincheng96, line 524:
```suggestion
    it('blocks malicious reentrant withdrawFrom', async () => {
```

contracts/Comet.sol
```diff
@@ -764,47 +784,27 @@ contract Comet is CometMainInterface {
...
+        (bool success, bytes memory returndata) = asset.call(abi.encodeWithSelector(ERC20.transfer.selector, to, amount));
+        if (!success || !(returndata.length == 0 || abi.decode(returndata, (bool)))) {
+            revert TransferOutFailed();
```
kevincheng96, line 805: What about having these custom errors for `TransferIn/Out` have a `bytes` arg of the `returnData`. This should allow our tests to decode the `returnData` and check that it matches the reentrancy custom error.

contracts/Comet.sol
```diff
@@ -764,47 +784,27 @@ contract Comet is CometMainInterface {
...
-                    success := not(0)          // set success to true
-                }
-                case 32 {                      // This is a compliant ERC-20
-                    returndatacopy(0, 0, 32)
-                    success := mload(0)        // Set `success = returndata` of override external call
-                }
-                default {                      // This is an excessively non-compliant ERC-20, revert.
-                    revert(0, 0)
-                }
+        _nonReentrantBefore();
```
kevincheng96, line 802: I'm leaning on having the reentrancy guards be on the `external` functions like `supply` and `withdraw` rather than the internal `doTransferIn/Out` functions. It's easier to reason about which functions need the guard, and we also avoid having to set up the guards twice in functions that transfer multiple times like `buyCollateral`.

contracts/IERC20NonStandard.sol
cwang25: I think bulker seems still referencing to this file, right 🤔 ?

test/supply-test.ts
kevincheng96: This suggestion (https://github.com/compound-finance/comet/pull/818#discussion_r1387336851) should allow us to check for reentrancy error as well.

contracts/IERC20NonStandard.sol
kevincheng96: Ah you're right, it is

contracts/CometCore.sol
cwang25: hmmm interestingly that I tried both way the compiled bytescode size are identical 🤔

test/buy-collateral-test.ts
cwang25: I think only this one doesn't work as it called 
` await ethers.provider.send('evm_setAutomine', [false]);`
then after all transaction step is done, it called 
```
 await ethers.provider.send('evm_mine', [start + 1000]);
 await ethers.provider.send('evm_setAutomine', [true]);
```
To process all steps at once.

contracts/Comet.sol
cwang25: Oh I see 😅  Updated it.

test/supply-test.ts
cwang25: oops, I think this is what failed the CI. 
Fixed it with the l right function call.

test/supply-test.ts
cwang25: I think before it was removed as there were two tests one test to having exceed supply cap and one test to ensure the inflated accounting number.  Since now both tests will now be reverted right away. 
I think we can go ahead and remove one of them. Since both are identical and both got reverted right away.

contracts/Comet.sol
cwang25: I tried to add to  
`supplyInternal` and `withdrawInternal` and `transferInternal` and `buyCollateral` 

But unfortunately :( the compiled bytecode size will be somewhat 24.8  :(

contracts/Comet.sol
```diff
@@ -1286,14 +1279,18 @@
...
      * @param amount The amount of an asset to approve
      */
     function approveThis(address manager, address asset, uint amount) override external {
         if (msg.sender != governor) revert Unauthorized();
 
-        IERC20NonStandard(asset).approve(manager, amount);
+        (bool success, bytes memory returndata) = asset.call(abi.encodeWithSelector(ERC20.approve.selector, manager, amount));
+        if (!success || (returndata.length != 0 && !abi.decode(returndata, (bool)))) {
+            revert ApproveFailed();
+        }
```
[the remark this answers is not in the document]
cwang25: Shall we follow this suggestions?

contracts/Comet.sol
cwang25: updated to use low-level call as well. And checked on success and bool return.

test/supply-test.ts
cwang25: After adding nonReentranct to all external functions, I can succesullfy verify the inner Error to be Reentrant error, although hardhat makes it a bit harder to assert encoded nested custom error :P

contracts/Comet.sol
cwang25: With the latest changes on Custom error, and nonReentrant on all exeternal function the contract size became: 
```
|  CometFactory                    ·      24.745  ·        +0.200  │
 ···································|··············|·················
 |  CometModifiedFactory            ·      24.883  ·        +0.200  │

```

Both are above 24.576Kb :(

contracts/Comet.sol
```diff
@@ -201,6 +201,23 @@ contract Comet is CometMainInterface {
         (asset14_a, asset14_b) = getPackedAssetInternal(config.assetConfigs, 14);
     }
 
+    modifier nonReentrant() {
```
kevincheng96, line 204: I think one way to reduce contract size is to split these logic into two separate internal functions: `nonReentrantBefore` and `nonReentrantAfter`. It should save contract size because it uses a JUMP instruction instead of adding the bytecode to each function that uses the modifier.

contracts/Comet.sol
```diff
@@ -1286,13 +1303,13 @@ contract Comet is CometMainInterface {
...
      * @dev Note: Setting the `asset` as Comet's address will allow the manager
      * to withdraw from Comet's Comet balance
+     * @dev Note: For USDT, if there is non-zero prior allowance, it must be reset to 0 first before setting a new value in proposal
      * @param asset The asset that the manager will gain approval of
      * @param manager The account which will be allowed or disallowed
      * @param amount The amount of an asset to approve
      */
     function approveThis(address manager, address asset, uint amount) override external {
         if (msg.sender != governor) revert Unauthorized();
-
```
kevincheng96, line 1295: nit: let's keep the extra line

contracts/CometCore.sol
```diff
@@ -56,6 +56,13 @@ abstract contract CometCore is CometConfiguration, CometStorage, CometMath {
     /// @dev The scale for factors
     uint64 internal constant FACTOR_SCALE = 1e18;
 
+    /// @dev The storage slot for reentrancy guard flags
+    bytes32 internal constant COMET_REENTRANCY_GUARD_FLAG_SLOT = bytes32(keccak256("comet.reentrancy.guard"));
+
+    /// @dev The reentrancy guard status
```
kevincheng96, line 62:
```suggestion
    /// @dev The reentrancy guard statuses
```

contracts/Comet.sol
kevincheng96: Sorry, just realized we went back and forth on this since you had it in separate functions originally 😅

cwang25: No problem :) 
The contract size reduced about 0.225 KiB 
(24.490KiB -> 24.266KiB)

contracts/Comet.sol
cwang25: No problem 👍

contracts/CometCore.sol
```diff
@@ -56,6 +56,13 @@ abstract contract CometCore is CometConfiguration, CometStorage, CometMath {
...
+    uint256 internal constant COMET_REENTRANCY_GUARD_NOT_ENTERED = 0;
+    uint256 internal constant COMET_REENTRANCY_GUARD_ENTERED = 1;
```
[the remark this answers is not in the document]
cwang25: removed COMET

test/buy-collateral-test.ts
```diff
@@ -409,7 +409,7 @@ describe('buyCollateral', function () {
   });
 
   describe('reentrancy', function() {
-    it('is not broken by reentrancy supply ', async () => {
+    it('is blocked during reentrant supply', async () => {
```
kevincheng96, line 412: Can you assert that the call reverts with `ReentrantCallBlocked`?

kevincheng96 (approved):
Nice, just a few final comments. But this is looking almost ready now 👍

contracts/Comet.sol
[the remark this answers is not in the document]
cwang25: added natspec 👍

test/buy-collateral-test.ts
cwang25: Added another small test case in the next to test on revert.

contracts/Comet.sol
```diff
@@ -201,12 +201,18 @@ contract Comet is CometMainInterface {
         (asset14_a, asset14_b) = getPackedAssetInternal(config.assetConfigs, 14);
     }
 
+    /**
+     * @dev Non-reentrant modifier
```
kevincheng96, line 205:
```suggestion
     * @dev Prevents marked functions from being reentered 
```

contracts/Comet.sol
```diff
@@ -201,12 +201,18 @@ contract Comet is CometMainInterface {
...
+     */
     modifier nonReentrant() {
         nonReentrantBefore();
         _;
         nonReentrantAfter();
     }
 
+    /**
+     * @dev This will set the flag before the call and revert if it is already set
```
kevincheng96, line 214:
```suggestion
     * @dev Checks that the reentrancy flag is not set and then sets the flag
```

contracts/Comet.sol
```diff
@@ -220,6 +226,9 @@ contract Comet is CometMainInterface {
         }
     }
 
+    /**
+     * @dev This will unset the flag after the call
```
kevincheng96, line 230:
```suggestion
     * @dev Unsets the reentrancy flag
```

test/buy-collateral-test.ts
```diff
@@ -562,5 +562,68 @@ describe('buyCollateral', function () {
...
+      await evilWETH.allocateTo(evilComet.address, exp(100, 18));
+      await EVIL.allocateTo(evilAlice.address, exp(5000, 6));
+
+      // approve Comet to move funds
+      await EVIL.connect(evilAlice).approve(EVIL.address, exp(5000, 6));
+      await EVIL.connect(evilAlice).approve(evilComet.address, exp(5000, 6));
+
+      // authorize EVIL, since callback will originate from EVIL token address
+      await evilComet.connect(evilAlice).allow(EVIL.address, true);
+      // call buyCollateral; supplyFrom is called in in callback
```
kevincheng96, line 618:
```suggestion
      // call buyCollateral; supplyFrom is called in callback
```

test/buy-collateral-test.ts
```diff
@@ -562,5 +562,68 @@ describe('buyCollateral', function () {
       // EvilToken attack should be blocked, so totalSupplyBase should be 0
       expect(evilBobPortfolio.internal.EVIL).to.equal(0);
     });
+
+    it('reentrant supply is reverted', async () => {
```
kevincheng96, line 566: Should be buycollateral, not supply. same for comments in test

cwang25: Ok renamed to buyCollateral. :) 
Thought in the existing test is using 'supply', and decided to name it the same. 
(I guess that rationale with that was the retrant attack loop is happening in supplyFrom via buyCollateral)

kevincheng96: Oh I see. That means the test isn't testing the right function. It should be triggering reentrancy via `BuyCollateral`. I think you'll have to set it in the EvilToken.

deployments/goerli/usdt/migrations/1691022234_configurate_and_ens.ts
kevincheng96, line 1: Committed by accident?

contracts/test/EvilToken.sol
```diff
@@ -40,7 +42,8 @@ contract EvilToken is FaucetToken {
             destination: address(this),
             asset: address(this),
             amount: 1e6,
-            maxCalls: type(uint).max
+            maxCalls: type(uint).max,
+            collateralAsset: address(0)
```
kevincheng96, line 46: I think you can just use the existing `asset` parameter instead of introducing a new one.

deployments/goerli/usdt/migrations/1691022234_configurate_and_ens.ts
cwang25: oops... yeah :( let me remove it.

contracts/test/EvilToken.sol
cwang25: sure 👍

kevincheng96 (approved):
Great, looks ready to merge

PR #827, opened 2024-02-02
Add USDT market on ETH mainnet.

vincetiu8:
Based on https://github.com/compound-finance/comet/pull/818.


contracts/Comet.sol
```diff
@@ -194,6 +194,8 @@ contract Comet is CometMainInterface {
 
     /**
      * @dev Prevents marked functions from being reentered 
+     * Note: this restrict contracts from calling comet functions in their hooks.
```
hannespfeiffer, line 197: Nitpick: "restrict" should be "restricts"

PR #831, opened 2024-02-22
Add USDC market to Sepolia

kevincheng96:
This PR adds support for Sepolia and deploys a new USDC market there. More work will need to be done to migrate cross-chain governance from L2s to point to Sepolia instead of Goerli, but this should help to unblock developers that need access to a Comet on Sepolia.


deployments/sepolia/usdc/relations.ts
kevincheng96, line 1: Mostly copied from the Goerli `relations.ts`. We will eventually need these L2 bridge contracts for cross-chain governance.

The relations for the price feeds are a new addition. We need them because Chainlink price feeds on Sepolia are not verified on Etherscan.

hayesgm (approved):
lgtm

PR #838, opened 2024-03-23
Deploy Compound III USDC market on Optimism 

dmitriy-woof-software:
This PR introduces the deployment for the USDC market on the Optimism network as well as the proposal to initialize the market there.

Forum thread - https://www.comp.xyz/t/deploy-compound-iii-on-optimism/4975


deployments/optimism/usdc/configuration.json
```diff
@@ -0,0 +1,65 @@
...
+      "priceFeed": "0x0D276FC14719f9292D5C1eA2198673d1f4269246",
+      "decimals": "18",
+      "borrowCF": 0.65,
+      "liquidateCF": 0.7,
+      "liquidationFactor": 0.8,
+      "supplyCap": "0e18"
+    },
+    "wstETH": {
+      "address": "0x1F32b1c2345538c0c6f582fCB022739c4A194Ebb",
+      "priceFeed": "0x698B585CbC4407e2D54aa898B2600B53C68958f7",
```
kevincheng96, line 57: This is the market-based `wstETH / USD` price feed. Lido's recent proposals to list wstETH seem to go with the exchange-rate-based `wstETH / stETH` * `ETH / USD` custom price feed. Let's maybe wait for them to finalize on that decision and make the same decision here to be consistent.

plugins/import/import.ts
kevincheng96, line 1: Unrelated. Pls revert

scenario/BulkerScenario.ts
```diff
@@ -8,7 +8,7 @@ import { exp } from '../test/helpers';
 scenario(
   'Comet#bulker > (non-WETH base) all non-reward actions in one txn',
   {
-    filter: async (ctx) => await isBulkerSupported(ctx) && !matchesDeployment(ctx, [{deployment: 'weth'}, {network: 'mumbai'}, { network: 'linea-goerli' }]),
+    filter: async (ctx) => await isBulkerSupported(ctx) && !matchesDeployment(ctx, [{deployment: 'weth'}, {network: 'mumbai'}, { network: 'linea-goerli' }, { network: 'optimism' }]),
```
kevincheng96, line 11: So based on the comment on line 7, this is failing because `WETH` is defined as the first collateral asset in `configuration.json`. I think it would be good to keep this test coverage, so IMO it would be better to define another asset (e.g. `OP`) as the first collateral asset and remove this filter.

deployments/optimism/usdc/migrations/1707394874_configurate_and_ens.ts
```diff
@@ -0,0 +1,290 @@
...
+        ],
+        [
+          setConfigurationCalldata,
+          deployAndUpgradeToCalldata,
+          setRewardConfigCalldata,
+        ],
+      ]
+    );
+
+    const COMPAmountToBridge = exp(3_600, 18);
```
kevincheng96, line 110: Curious why this amount? Was this recommended by Gauntlet?

kevincheng96 (commented):
Great work on this so far! Left some comments.

deployments/optimism/usdc/migrations/1707394874_configurate_and_ens.ts
dmitriy-bergman-works: Yes. The Gauntlet team provided 3600 COMP for the 6 months.

plugins/import/import.ts
dmitriy-bergman-works: Should use regex with `"` because it cannot read contract https://etherscan.io/address/0x566511a1A09561e2896F8c0fD77E8544E59bFDB0#code

dmitriy-bergman-works: Снимок экрана 2024-03-25 в 22 49 22

dmitriy-bergman-works: As this contract has no transactions, it tries to parse it from html, but regexp with single quote does not match.

dmitriy-bergman-works: https://github.com/compound-finance/comet/pull/838/files#diff-4d03efd71c6d73abcd486164f95ba4a821f5d42f2068420121910e79983f43b5R91

deployments/optimism/usdc/configuration.json
dmitriy-bergman-works: Not clear. Could we not wait for Lido's decision?

kevincheng96: As you suggested, you could just exclude `wstETH` for now until that decision has been made.

PR #842, opened 2024-04-06
Proposal for Optimism USDC market

dmitriy-bergman-works:


kevincheng96 (approved):
Proposal is live on-chain: https://compound.finance/governance/proposals/238

PR #850, opened 2024-05-08
Deploy compound v3 USDT market on Polygon

EviLord032:


scenario/constraints/ProposalConstraint.ts
EviLord032, line 1: Proposals:
1. https://compound.finance/governance/proposals/246
2. https://compound.finance/governance/proposals/247
Are going to fail, thus we skip them to pass scenarios. These changes should be removed after deployment.

PR #909, opened 2024-08-14
Add weETH as collateral to WETH market on Optimism

MishaShWoof:


alexandfox (approved):
looks good, may want to check conflict with 905

PR #914, opened 2024-08-28
Add wrsETH as collateral to WETH market on Base

MishaShWoof:


dmitriy-woof-software:
Proposal - https://compound.finance/governance/proposals/319

PR #923, opened 2024-09-13
Add cbBTC as collateral to WETH market on Mainnet

MishaShWoof:


SMSDAO (commented):
Looking good for me !

alexandfox (approved):
👍🏼

PR #1012, opened 2025-08-14
Sync infra, tests, scenarios

dmitriy-woof-software:


scenario/utils/index.ts
```diff
@@ -701,6 +705,247 @@ async function mockRedstoneOracle(dm: DeploymentManager, feed: string) {
   await proxyAdmin.connect(owner).upgrade(feed, newImplementation.address);
 }
 
+export async function tenderlyExecute(
```
torreyatcitty, line 708: What does this function actually do? Looking at it, it appears it is simulating executing a proposal on chain? Not sure if there is a snippet read me somewhere on where/how this might be used?

VladyslavBudiukWOOF: The `tenderlyExecute` function votes on and executes the cached proposal, and if L2 execution is required, it also executes it on L2. Currently, it works only with a simulation of `enact`, but we are planning to modify it to support on-chain proposals.
For example:
 ```
If a proposal adds new collateral, this function simulates the execution of that proposal. The result would look like this:
========================== TENDERLY ==========================


Executing Tenderly simulation for proposal 479...
Tenderly simulation bundle size: 9
 >>> PROPOSAL EXECUTED  479 

Simulation 4a742b77-3914-4e18-910e-27ee2f310c6b done, status: true
Link: https://www.tdly.co/shared/simulation/4a742b77-3914-4e18-910e-27ee2f310c6b
Proposals relayed: 1
 >>> PROPOSAL RELAYED 479 

Simulation d15fd661-1426-439e-9e02-b842970fc72c done, status: true
Link: https://www.tdly.co/shared/simulation/d15fd661-1426-439e-9e02-b842970fc72c

================================================================
```
This example comes from [this GitHub Action run](https://github.com/woof-software/comet/actions/runs/17771704262).

File: scenario/utils/relayArbitrumMessage.ts [hunk 2]
```diff
@@ -6,10 +6,15 @@ import { Log } from '@ethersproject/abstract-provider';
 import { sourceTokens } from '../../plugins/scenario/utils/TokenSourcer';
 import { OpenBridgedProposal } from '../context/Gov';
 
+function isTenderlyLog(log: any): log is { raw: { topics: string[], data: string } } {
```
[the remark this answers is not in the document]
VladyslavBudiukWOOF: We will move this function and another one to some kind of local helper(maybe index)

PR #1099, opened 2026-03-23
[WIP] feat: Partial Liquidation

vlad-woof-software:
## Overview                                                                                                                        
                                                                                                                                     
  This PR introduces partial liquidation support to the Comet protocol.                                                              
  Instead of always seizing all collateral from an underwater account,                                                               
  the liquidator now brings the account's health factor up to a configured                                                           
  `targetHealthFactor` threshold - seizing only as much collateral as needed.
  This reduces unnecessary value extraction from borrowers and better aligns
  with the protocol RFC.

  ## What has been done

  ### Configuration
  - `targetHealthFactor` moved into `CometConfiguration.Configuration` struct - no longer a separate contract/interface, but a
  first-class config field                                                                                                           
  - `targetHealthFactor` initialized in `Comet` and `ScrollComet` constructors from config
  - `targetHealthFactor()` added to `CometMainInterface`                                                                             
  - `Configurator` updated to expose `setTargetHealthFactor` management

  ### Core logic
  - Partial liquidation algorithm implemented in `CometWithExtendedAssetList.absorbInternal`: iterates collateral assets and either
  partially seizes one asset (stopping at `targetHF`) or fully seizes it and moves to the next
  - Formula: `∆ = (targetHF × debt − totalCV) × FACTOR_SCALE / (LP × targetHF − CF)`
  - Falls back to original full-seizure behaviour when `targetHF == 0`

  ### Tests
  - Initial test suite added in `test/partial-liquidation-test.ts` covering basic scenarios: single-collateral partial seizure,
  multi-collateral mixed seizure, full-liquidation fallback                                                                          
  - Existing test suite fixed to account for new `targetHealthFactor` config field and updated `makeProtocol` setup
                                                                                                                                     
  ## What is still pending (reason for WIP)                                                                                          
   
  - Test coverage: edge cases and full flow not yet covered (e.g. `baseBorrowMin` boundary, dust positions, rounding behaviour)      
  - `baseBorrowMin` check: if remaining debt after partial seizure falls below `baseBorrowMin`, the protocol should fall through to
  full liquidation instead
  - Configurator-level validation that `LP × targetHF > CF` for all assets (prevents invalid configuration that would cause revert or
   division by zero at runtime)
  - Removal of remaining `console.log` debug calls from contracts
  - Internal review
  - Scenario / integration tests
  - Audit-readiness review


contracts/CometExtPartialLiquidation.sol
VladyslavBudiukWOOF, line 1: Remove this file too. This is a part of the CometWithPartialLiquidation

contracts/ConfiguratorPartialLiquidation.sol
VladyslavBudiukWOOF, line 1: Remove this file. This is a part of the CometWithPartialLiquidation

docs/partial-liquidation-analysis.md
VladyslavBudiukWOOF, line 1: This should be an internal doc. Remove

docs/partial-liquidation-example.md
VladyslavBudiukWOOF, line 1: This should be an internal doc. Remove

docs/partial-liquidation-fix-plan.md
VladyslavBudiukWOOF, line 1: This should be an internal doc. Remove

docs/partial-liquidation-flowchart.md
VladyslavBudiukWOOF, line 1: This should be an internal doc. Remove

docs/partial-liquidation-revert-cases.md
VladyslavBudiukWOOF, line 1: This should be an internal doc. Remove

contracts/CometConfiguration.sol
```diff
@@ -33,6 +33,7 @@ contract CometConfiguration {
         uint104 baseMinForRewards;
         uint104 baseBorrowMin;
         uint104 targetReserves;
+        uint64 targetHealthFactor;
```
VladyslavBudiukWOOF: Why was the variable in this exact place? Did you calculate the slots?

contracts/AssetList.sol
```diff
@@ -141,6 +142,9 @@ contract AssetList {
         if (assetConfig.borrowCollateralFactor >= assetConfig.liquidateCollateralFactor) revert CometMainInterface.BorrowCFTooLarge();
         if (assetConfig.liquidateCollateralFactor > MAX_COLLATERAL_FACTOR) revert CometMainInterface.LiquidateCFTooLarge();
 
+        // Ensure target health factor is valid relative to collateral factors
+        if ((assetConfig.liquidationFactor * targetHealthFactor / FACTOR_SCALE) <= assetConfig.borrowCollateralFactor) revert CometMainInterface.BadAssetHealthFactor();
```
pavlo-h: we have guaranteed:  borrowCollateralFactor < liquidateCollateralFactor (by check above)
and we have guaranteed relation: liquidateCollateralFactor <= liquidationFactor
so borrowCollateralFactor is always < than liquidationFactor

also targetHF >=1.05, so liquidationFactor < liquidationFactor * targetHF

that's why we can set a check:

if (assetConfig.liquidateCollateralFactor > assetConfig.liquidationFactor) revert CometMainInterface.LiquidateCFTooLarge();

action points:
- remove check on targetHF
- remove targetHF parameter added to the function interface
- add check instead: if (assetConfig.liquidateCollateralFactor > assetConfig.liquidationFactor) revert CometMainInterface.LiquidateCFTooLarge();

contracts/AssetList.sol
```diff
@@ -113,9 +113,10 @@ contract AssetList {
      *      and the supply cap in the next 64 bits
      * @param assetConfigs The asset configurations
      * @param i The index of the asset info to get
+     * @param targetHealthFactor The target health factor for the asset list, used for validation
      * @return The packed asset info
      */
-    function getPackedAssetInternal(CometConfiguration.AssetConfig[] memory assetConfigs, uint i) internal view returns (uint256, uint256) {
+    function getPackedAssetInternal(CometConfiguration.AssetConfig[] memory assetConfigs, uint i, uint targetHealthFactor) internal view returns (uint256, uint256) {
```
pavlo-h: parameter can be removed - see comment on the check for liquidationFactor

contracts/AssetList.sol
```diff
@@ -72,34 +72,34 @@ contract AssetList {
     /// @notice The number of assets this contract actually supports
     uint8 public immutable numAssets;
     
-    constructor(CometConfiguration.AssetConfig[] memory assetConfigs) {
+    constructor(CometConfiguration.AssetConfig[] memory assetConfigs, uint targetHealthFactor) {
```
pavlo-h: parameter can be removed - see comment on the check for liquidationFactor

contracts/AssetListFactory.sol
```diff
@@ -13,10 +13,11 @@ contract AssetListFactory {
     /**
      * @notice Create a new asset list
      * @param assetConfigs The asset configurations
+     * @param targetHealthFactor The target health factor for the asset list, used for validation
      * @return assetList The address of the new asset list
      */
-    function createAssetList(CometCore.AssetConfig[] memory assetConfigs) external returns (address assetList) {
-        assetList = address(new AssetList(assetConfigs));
+    function createAssetList(CometCore.AssetConfig[] memory assetConfigs, uint targetHealthFactor) external returns (address assetList) {
```
pavlo-h: parameter can be removed - see comment on the check for liquidationFactor

contracts/CometConfiguration.sol
pavlo-h: Add comments on slots, as currently it is not obvious why targetHF was added at that place. For example:

    struct Configuration {
        address governor;							// slot 1: 20 bytes
        address pauseGuardian;					// slot 2: 20 bytes
        address baseToken;
        address baseTokenPriceFeed;
        address extensionDelegate;

        uint64 supplyKink;
        uint64 supplyPerYearInterestRateSlopeLow;
        uint64 supplyPerYearInterestRateSlopeHigh;
        uint64 supplyPerYearInterestRateBase;		// slot 6: 32 bytes
        uint64 borrowKink;
        uint64 borrowPerYearInterestRateSlopeLow;
        uint64 borrowPerYearInterestRateSlopeHigh;
        uint64 borrowPerYearInterestRateBase;		// slot 7: 32 bytes
        uint64 storeFrontPriceFactor;
        uint64 trackingIndexScale;
        uint64 baseTrackingSupplySpeed;
        uint64 baseTrackingBorrowSpeed;				// slot 8: 32 bytes
        uint104 baseMinForRewards;
        uint104 baseBorrowMin;						// slot 9: 208 bits = 26 bytes
        uint104 targetReserves;
        uint64 targetHealthFactor;					// slot 10: 168 bits = 21 bytes

        AssetConfig[] assetConfigs;					// slot 11: 32 bytes
    }

or use any convenient compact comment format

contracts/IAssetListFactory.sol
```diff
@@ -10,7 +10,8 @@ interface IAssetListFactory {
     /**
      * @notice Create a new asset list
      * @param assetConfigs The asset configurations
+     * @param targetHealthFactor The target health factor for the asset list, used for validation
      * @return assetList The address of the new asset list
      */
-    function createAssetList(CometCore.AssetConfig[] memory assetConfigs) external returns (address assetList);
+    function createAssetList(CometCore.AssetConfig[] memory assetConfigs, uint targetHealthFactor) external returns (address assetList);
```
pavlo-h: parameter can be remove - see comment on the liquidationFactor check

contracts/Configurator.sol
```diff
@@ -293,6 +300,10 @@ contract Configurator is ConfiguratorStorage {
 
     /** Other helpers **/
 
+    function targetHealthFactors(address cometProxy) external view returns (uint64) {
```
pavlo-h: remove this function - it is not used and not necessary. You have getConfiguration() which returns all necessary parameters

contracts/CometWithExtendedAssetList.sol
```diff
@@ -94,6 +94,9 @@ contract CometWithExtendedAssetList is CometMainInterface {
     /// @notice The minimum base token reserves which must be held before collateral is hodled
     uint public override immutable targetReserves;
 
+    /// @notice The target health factor for partial liquidation (0 = full liquidation)
+    uint public override immutable targetHealthFactor;
```
pavlo-h: uint256 - always use strict type convention

contracts/CometWithExtendedAssetList.sol
```diff
@@ -106,8 +109,12 @@ contract CometWithExtendedAssetList is CometMainInterface {
     /// @notice The address of the asset list
     address immutable public assetList;
 
+    /// @notice Maximum number of assets supported when using an external asset list
```
pavlo-h: comment duplciates the name of the constant, remove it

contracts/CometWithExtendedAssetList.sol
```diff
@@ -106,8 +109,12 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
     uint8 internal constant MAX_ASSETS_FOR_ASSET_LIST = 24;
 
+    /// @notice The minimum allowed value for targetHealthFactor (1.05 in factor scale)
```
pavlo-h: comment duplicates the name of the constant - remove it

contracts/CometWithExtendedAssetList.sol
```diff
@@ -159,7 +166,7 @@ contract CometWithExtendedAssetList is CometMainInterface {
         // Set asset info
         numAssets = uint8(config.assetConfigs.length);
 
-        assetList = IAssetListFactory(IAssetListFactoryHolder(extensionDelegate).assetListFactory()).createAssetList(config.assetConfigs);
+        assetList = IAssetListFactory(IAssetListFactoryHolder(extensionDelegate).assetListFactory()).createAssetList(config.assetConfigs, targetHealthFactor);
```
pavlo-h: remove parameter - see commet on liquidationFactor check in assetList

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
         liquidatorPoints[absorber] = points;
     }
 
+    /// @dev Collateral-factor-weighted sum of collateral USD value for the account.
+    ///      When `liquidation=false` uses borrowCollateralFactor (used by absorbInternal to
+    ///      initialise totalCollateralizedValue for the partial-liquidation formula).
+    ///      When `liquidation=true` uses liquidateCollateralFactor (used by isLiquidatable).
+    function _getLiquidity(address account, bool liquidation) internal view returns (uint256) {
```
pavlo-h: sense of this function was to optimize the number of pricefeed calls in functions isLiquidatable(), absorbInternal() and isColalterized(). Currently the function does not fulfill its purpose and only encapsulates a small piece of logic. Review again how the function should be used and optimize price feeds calls and number of times user's collaterals are checked.

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+            if (isInAsset(assetsIn, i, _reserved)) {
+                AssetInfo memory asset = getAssetInfo(i);
+                uint256 newAmount = mulPrice(
+                    userCollateral[account][asset.asset].balance,
+                    getPrice(asset.priceFeed),
+                    asset.scale
+                );
+                liquidity += mulFactor(
+                    newAmount,
+                    liquidation ? asset.liquidateCollateralFactor : asset.borrowCollateralFactor
```
pavlo-h, line 1284: this place signals that the function was supposed to be used in isCollaterized as well (to decrease the number of duplicated logic). However it is currently unused for that. Use the function for its purpose.

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
         liquidatorPoints[absorber] = points;
     }
 
+    /// @dev Collateral-factor-weighted sum of collateral USD value for the account.
```
pavlo-h: why does comment not follow natspec standard? with params and return value?

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+    ///      When `liquidation=false` uses borrowCollateralFactor (used by absorbInternal to
+    ///      initialise totalCollateralizedValue for the partial-liquidation formula).
+    ///      When `liquidation=true` uses liquidateCollateralFactor (used by isLiquidatable).
```
pavlo-h: currently is not used in isLiquidatable

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+     *          is converted to base tokens and credited to the account.
+     *        - If a full-liquidation path was taken (no asset was able to reach
+     *          `targetHF`) any residual negative balance is zeroed out so the
+     *          protocol absorbs the remaining bad debt.
+     *
+     * @param absorber The address that called `absorb` and receives liquidator-points credit
+     * @param account  The underwater account whose collateral and debt are being absorbed
      */
     function absorbInternal(address absorber, address account) internal {
         if (!isLiquidatable(account)) revert NotLiquidatable();
```
pavlo-h: review comments on _getLiquidity() and use the function within the isLiquidatable() function. _getLiquidity() returns liquidity which can be checked against 0 within the isLiquidatable()

for example
```
function isLiquidatable(address account) override public view returns (bool) {
        int104 principal = userBasic[account].principal;
        if (principal >= 0) return false;
        int debt = signedMulPrice(
            presentValue(principal),
            getPrice(baseTokenPriceFeed),
            uint64(baseScale)
        );
        return (debt + int(_getLiquidity(account, true)) < 0);
}
```

pavlo-h: same for isCollaterized()

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+            }
+            unchecked { i++; }
+        }
+        return liquidity;
+    }
+
     /**
      * @dev Transfer user's collateral and debt to the protocol itself.
+     *      Iterates over all collateral assets the account holds and seizes them
+     *      one by one until the account's health factor reaches `targetHealthFactor`.
```
pavlo-h: ... until the account's health factor reaches `targetHealthFactor` or until the whole debt value is covered.

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+     *
+     *      Partial-liquidation logic (applied per asset):
```
pavlo-h: why comment just duplicates the codebase? it should work as summary of the function, not its exact copy or should add details for unclear logic. Current comment does not follow the best practice.

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
-                uint128 seizeAmount = userCollateral[account][asset].balance;
-                userCollateral[account][asset].balance = 0;
-                totalsCollateral[asset].totalSupplyAsset -= seizeAmount;
+        uint256 deltaValue; // Accumulated liqFactor-discounted USD value of all collateral seized so far
+        uint256 targetHF = targetHealthFactor;
+
+        // Note: targetHealthFactor is always ≥ MIN_TARGET_HEALTH_FACTOR (enforced in constructor),
+        //       so the seizure formula below is always well-defined.
+        // debt is negative (USD) when the account has a borrow; used to compute debtRemaining per asset
+        int debt = signedMulPrice(presentValue(oldPrincipal), basePrice, uint64(baseScale));
```
pavlo-h: int256 - use strict types

contracts/CometCore.sol
```diff
@@ -17,6 +17,13 @@ abstract contract CometCore is CometConfiguration, CometStorage, CometMath {
         uint128 supplyCap;
     }
 
+    struct LiquidationData {
```
pavlo-h: why do you need a struct which is used only within 1 internal function and never leaves the function?

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
-                address asset = assetInfo.asset;
-                uint128 seizeAmount = userCollateral[account][asset].balance;
-                userCollateral[account][asset].balance = 0;
-                totalsCollateral[asset].totalSupplyAsset -= seizeAmount;
+        uint256 deltaValue; // Accumulated liqFactor-discounted USD value of all collateral seized so far
+        uint256 targetHF = targetHealthFactor;
+
+        // Note: targetHealthFactor is always ≥ MIN_TARGET_HEALTH_FACTOR (enforced in constructor),
+        //       so the seizure formula below is always well-defined.
+        // debt is negative (USD) when the account has a borrow; used to compute debtRemaining per asset
```
pavlo-h: useless autogenerated comment, remove

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
-            if (isInAsset(assetsIn, i, _reserved)) {
-                AssetInfo memory assetInfo = getAssetInfo(i);
-                address asset = assetInfo.asset;
-                uint128 seizeAmount = userCollateral[account][asset].balance;
-                userCollateral[account][asset].balance = 0;
-                totalsCollateral[asset].totalSupplyAsset -= seizeAmount;
+        uint256 deltaValue; // Accumulated liqFactor-discounted USD value of all collateral seized so far
+        uint256 targetHF = targetHealthFactor;
+
+        // Note: targetHealthFactor is always ≥ MIN_TARGET_HEALTH_FACTOR (enforced in constructor),
```
pavlo-h: note is not attached to the code, remove

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+        //       so the seizure formula below is always well-defined.
+        // debt is negative (USD) when the account has a borrow; used to compute debtRemaining per asset
+        int debt = signedMulPrice(presentValue(oldPrincipal), basePrice, uint64(baseScale));
+
+        LiquidationData memory liquidationData;
+        // BorrowCF-weighted collateral USD value; decremented as fully-seized assets leave the portfolio
```
pavlo-h: please remove useless AI-generated comments

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+        AssetInfo memory assetInfo;
+
+        for (uint8 i; i < numAssets; ) {
+            if (isInAsset(accountUser.assetsIn, i, accountUser._reserved)) {
+                assetInfo = getAssetInfo(i);
+                uint256 price = getPrice(assetInfo.priceFeed);
+                uint256 fullBalance = userCollateral[account][assetInfo.asset].balance;
+                uint256 availableUSD = mulPrice(fullBalance, price, assetInfo.scale);
+
+                // Gap-closure rate: how much (debtRemaining × targetHF − totalCollaterizedValue) shrinks
```
pavlo-h: please remove useless AI-generated comments which duplicate the code
comments should be meaningfull, add missing details or context, or serve as visual representation of the code

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                // per 1 USD of this asset seized — derived from the targetHF equation:
+                //   (collateral − x·CF) = targetHF × (debt − x·LF)  =>  denom = LF·targetHF − CF
+                // Safety: denom > 0 is guaranteed by the AssetList constructor check
+                //   `LF * targetHF / FACTOR_SCALE > CF`  (AssetList.sol line 146),
+                // so this subtraction never underflows.
+                uint256 denom = mulFactor(assetInfo.liquidationFactor, targetHF) - assetInfo.borrowCollateralFactor;
+
+                // Remaining USD debt not yet offset by collateral seized in earlier loop iterations
```
pavlo-h: please remove useless AI-generated comments which duplicate the code

contracts/CometWithExtendedAssetList.sol
pavlo-h: you already have oldBalance, you can omit extra presentValue conversion

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+
+        LiquidationData memory liquidationData;
+        // BorrowCF-weighted collateral USD value; decremented as fully-seized assets leave the portfolio
+        liquidationData.totalCollaterizedValue = _getLiquidity(account, false);
+        AssetInfo memory assetInfo;
+
+        for (uint8 i; i < numAssets; ) {
+            if (isInAsset(accountUser.assetsIn, i, accountUser._reserved)) {
+                assetInfo = getAssetInfo(i);
+                uint256 price = getPrice(assetInfo.priceFeed);
```
pavlo-h: you can move memory vars creation out of the cycle

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                uint256 fullBalance = userCollateral[account][assetInfo.asset].balance;
```
pavlo-h: please, let's call it collateralBalance, as fullBalance is confusing

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                uint256 availableUSD = mulPrice(fullBalance, price, assetInfo.scale);
```
pavlo-h: please, let's call it collateralValue, as availableUSD is confusing

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+
+                // Gap-closure rate: how much (debtRemaining × targetHF − totalCollaterizedValue) shrinks
+                // per 1 USD of this asset seized — derived from the targetHF equation:
+                //   (collateral − x·CF) = targetHF × (debt − x·LF)  =>  denom = LF·targetHF − CF
+                // Safety: denom > 0 is guaranteed by the AssetList constructor check
+                //   `LF * targetHF / FACTOR_SCALE > CF`  (AssetList.sol line 146),
+                // so this subtraction never underflows.
```
pavlo-h: meaningfull comment will look like smth as:
LF * targetHF > LF > LCF > BCF (checked in Configurator)

or smth like that. Less == more

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                uint256 denom = mulFactor(assetInfo.liquidationFactor, targetHF) - assetInfo.borrowCollateralFactor;
```
pavlo-h: denom is used only once - so put it back into the formula. Its separation is confusing

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+
+                // Remaining USD debt not yet offset by collateral seized in earlier loop iterations
+                uint256 debtRemaining = uint256(-debt) - deltaValue;
+                // USD collateral to seize so that the account lands exactly at targetHF:
+                //   rawCollateralUSD = (debtRemaining × targetHF - totalCollaterizedValue) / denom
```
pavlo-h: as for the comments - again there is a comment which duplicates a code and is factually useless

and as for the code - we have a formula which seems to come from nowhere. So let's put formula deriving into the comment

for example
```
// current HF = total collateral value / debt
// note: total collateral value is a sum scaled by CF, so it is collateral value 1 * CF 1 + collateral value 2 * CF 2 ...
// we calculate expected HF after 1 step of seizing cycle - when we seize the 1st collateral
// expected HF = (total collateral value - siezed collateral value * CF) / (debt - delta debt)
//
// delta debt = siezed collateral * LF (as we penalize the value of the seized collataral
//
// we want expected HF to be targetHF, so make substitutions:
// targetHF = (total collateral value - siezed collateral value * CF) / (debt - siezed collateral value * LF)
// targetHF * (debt - siezed collateral value * LF) = (total collateral value - siezed collateral value * CF)
// targetHF * debt - targetHF * siezed collateral value * LF = total collateral value - siezed collateral value * CF
// targetHF * siezed collateral value * LF - siezed collateral value * CF = targetHF * debt - total collateral value
// siezed collateral value * (targetHF * LF - CF) = targetHF * debt - total collateral value
// siezed collateral value = (targetHF * debt - total collateral value) / (targetHF * LF - CF)
```

but of course use more compact form - there is no need to show all steps. just essential ones

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                uint256 rawCollateralUSD = (mulFactor(debtRemaining, targetHF) - liquidationData.totalCollaterizedValue)
```
pavlo-h: also, let's call a variable wantedCollateralValue or seizedCollateralValue smth like that

rawCollateral gives no information what does it mean, and USD suffix is not unified with balance/value suffixes we have

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                    * FACTOR_SCALE / denom;
+
```
pavlo-h: also good place for a comment which describes the context. At this point the reader has no idea what is rawCollateral (so address the comment above to describe why we need that) and now has no idea why we compare it with the collateral value (if renaming comment is addressed). Good place to describe why we do that

for example:
`                // so, we want to sieze a collateral of value calculated above. Further choice is simple - if user has more collateral than we want, we seize only calculated value
                // if user has less collateral value than we want - we seize what we can and move to the next cycle step to the next collateral`

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                    uint256 debtReduction = mulFactor(rawCollateralUSD, assetInfo.liquidationFactor);
+                    uint256 debtAfterPartial = debtRemaining - debtReduction;
+                    uint256 baseDebtAfterPartial = divPrice(debtAfterPartial, basePrice, uint64(baseScale));
+
+                    if (baseDebtAfterPartial >= baseBorrowMin) {
+                        // Partial seizure: exactly reaches targetHF
+                        liquidationData.seizeAmount = divPrice(rawCollateralUSD, price, assetInfo.scale);
+                        liquidationData.seizedValue = debtReduction;
+                        liquidationData.currentHF = targetHF;
+                    } else {
```
pavlo-h: this branch is incorrect - you should not seize the whole collateral, but only the value of the debt if applicable

e.g. 
seizeAmount += divPrice(baseDebtAfterPartial, price, assetInfo.scale);
and if this amount is less than current collateral amount than re-calculate the seizeValue

there may be a possibility to go to another collateral, but also may be not (e.g. dust debt is 1$ and we have 100$ of collateral left)

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                        // Remaining debt would be a dust position — fall through to full liquidation
+                        liquidationData.seizeAmount = fullBalance;
+                        liquidationData.seizedValue = mulFactor(availableUSD, assetInfo.liquidationFactor);
+                        liquidationData.currentHF = 0;
```
pavlo-h: another incorrect flow - in case if minDebt we need to recalculate if current collateral covers the dust leftover - as we may have enough collateral to cover it. But you continue to the next cycle, though we may have enough to break the cycle. So, recalculate amount and value if current collateral covers it.

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1050,62 +1057,169 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+                    // Asset insufficient — seize fully, continue to next
+                    liquidationData.seizeAmount = fullBalance;
+                    liquidationData.seizedValue = mulFactor(availableUSD, assetInfo.liquidationFactor);
+                    liquidationData.currentHF = 0;
+                }
 
-                uint256 value = mulPrice(seizeAmount, getPrice(assetInfo.priceFeed), assetInfo.scale);
-                deltaValue += mulFactor(value, assetInfo.liquidationFactor);
+                deltaValue += liquidationData.seizedValue;
+                if (liquidationData.currentHF < targetHF) {
```
pavlo-h: why did you break if-else for HF comparison?

use the intended order:
```
deltaValue +=
emit event

userCollateral -=
totalsCollateral -=

if (currentHF < targetHF) {
    totalCollaterizedValue  -=
}
else break;
```

you will save space and gas on 1 condition less

contracts/AssetList.sol
vlad-woof-software: ++

contracts/AssetList.sol
vlad-woof-software: ++

contracts/AssetList.sol
vlad-woof-software: ++

contracts/AssetListFactory.sol
vlad-woof-software: ++

contracts/CometConfiguration.sol
vlad-woof-software: ++

contracts/IAssetListFactory.sol
vlad-woof-software: ++

contracts/Configurator.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/AssetList.sol
```diff
@@ -139,7 +139,10 @@ contract AssetList {
 
         // Ensure collateral factors are within range
         if (assetConfig.borrowCollateralFactor >= assetConfig.liquidateCollateralFactor) revert CometMainInterface.BorrowCFTooLarge();
-        if (assetConfig.liquidateCollateralFactor > MAX_COLLATERAL_FACTOR) revert CometMainInterface.LiquidateCFTooLarge();
+        if (
+            assetConfig.liquidateCollateralFactor > MAX_COLLATERAL_FACTOR &&
```
pavlo-h: here should be OR, not AND

contracts/CometConfiguration.sol
pavlo-h: target health factor is now no incorrect place, you moved it from 10th slot (after targetHealthFactor) to 9th slot, thus corrupting the storage

before:
uint104 baseMinForRewards;
uint104 baseBorrowMin;
uint104 targetReserves;
uint64 targetHealthFactor
104 + 104 = 208, 9th slot, so targetReserves is in the next slot

now:
uint64 targetHealthFactor;
uint104 baseMinForRewards;
uint104 baseBorrowMin;
uint104 targetReserves; 

64 + 104 = 168, 9th slot. Though baseMinForRewards is now moved 64 bits right and corrupted.
and baseBorrowMin is moved from 9th slot to 10th slot and corrupted

contracts/AssetList.sol
pavlo-h: ++

contracts/Configurator.sol
pavlo-h: ++

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1051,61 +1008,156 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+
+                assetInfo = getAssetInfo(i);
+                price = getPrice(assetInfo.priceFeed);
+                collateralBalance = userCollateral[account][assetInfo.asset].balance;
+                collateralValue = mulPrice(collateralBalance, price, assetInfo.scale);
+
+                debtRemaining = uint256(-debt) - deltaValue;
+                // The denominator will always be greater than zero, since:
+                // LF * targetHF > LF > LCF > BCF (checked in Configurator)
+                wantedCollateralValue = (mulFactor(debtRemaining, targetHF) - liquidationData.totalCollaterizedValue)
```
pavlo-h: formula is still taken from a thin air - good place to put comments

contracts/AssetList.sol
vlad-woof-software: ++

contracts/CometWithExtendedAssetList.sol
vlad-woof-software: ++

contracts/Configurator.sol
```diff
@@ -238,6 +239,12 @@ contract Configurator is ConfiguratorStorage {
         emit SetTargetReserves(cometProxy, oldTargetReserves, newTargetReserves);
     }
 
+    function setTargetHealthFactor(address cometProxy, uint64 newTargetHealthFactor) external governorOrMarketAdmin {
```
Im-Vishwjeet, line 242: `setTargetHealthFactor` skips the `MIN_TARGET_HEALTH_FACTOR` check that the constructor enforces.

`CometWithExtendedAssetList.sol:130` guards deployment with:

```solidity
if (config.targetHealthFactor < MIN_TARGET_HEALTH_FACTOR) revert BadHealthFactor();
where MIN_TARGET_HEALTH_FACTOR = 105e16. But this setter writes any uint64 — including 0 — directly into configuratorParams[cometProxy].targetHealthFactor. Because governorOrMarketAdmin is broader than onlyGovernor, any market admin can quietly stage an out-of-range value.

Impact path:

Setter succeeds, event emits, getter reads back the invalid value — no revert.
The next deploy() call reverts in the Comet constructor with BadHealthFactor(), potentially after a governance proposal has already queued.
The phase-1 denominator mulFactor(LF, targetHF) - BCF becomes -BCF if targetHF = 0, which underflows uint256. Enforcing the bound at write time makes that state unreachable from storage.
Suggested:


function setTargetHealthFactor(address cometProxy, uint64 newTargetHealthFactor) external governorOrMarketAdmin {
    if (newTargetHealthFactor < MIN_TARGET_HEALTH_FACTOR) revert BadHealthFactor();
    uint64 oldTargetHealthFactor = configuratorParams[cometProxy].targetHealthFactor;
    configuratorParams[cometProxy].targetHealthFactor = newTargetHealthFactor;
    emit SetTargetHealthFactor(cometProxy, oldTargetHealthFactor, newTargetHealthFactor);
}
One wrinkle: MIN_TARGET_HEALTH_FACTOR currently lives as an internal constant inside CometWithExtendedAssetList.sol. To reference it from Configurator.sol without duplicating the literal, it probably wants to move to CometConfiguration.sol (or a shared constants file).

File: contracts/Comet.sol [hunk 40]
```diff
@@ -91,6 +91,9 @@ contract Comet is CometMainInterface {
     /// @notice The minimum base token reserves which must be held before collateral is hodled
     uint public override immutable targetReserves;
 
+    /// @notice The target health factor for partial liquidation
+    uint public override immutable targetHealthFactor;
```
Im-Vishwjeet, line 95: Is `targetHealthFactor` intentionally a no-op in `Comet.sol / ScrollComet.sol`? I see it gets stored in the constructors but isn't referenced in their `absorbInternal`. Asking because the constant `MIN_TARGET_HEALTH_FACTOR` and the corresponding revert check only exist in `CometWithExtendedAssetList.sol` — wanted to confirm that's by design (interface uniformity) rather than an oversight before reading further.

contracts/CometWithExtendedAssetList.sol
```diff
@@ -1051,66 +1008,287 @@ contract CometWithExtendedAssetList is CometMainInterface {
...
+        // Total borrowCollateralFactor-weighted USD value of the borrower's collateral,
+        // used as the baseline for the targetHF formula in the first iteration.
+        uint256 totalCollateralizedValue = _getLiquidity(account, false);
+        uint256 collateralValue;
+
+        AssetInfo memory assetInfo;
+        // numAssets lives in storage and would be re-read on every iteration - cache it once.
+        uint8 _numAssets = numAssets;
+
+        for (uint8 i; i < _numAssets;) {
```
Im-Vishwjeet: Since it breaks on the first asset that satisfies phase-1, the asset at the lowest `assetsIn` index is the one whose balance gets reduced. Under full liquidation the order didn't matter; under partial it does.

Example: index 0 is Asset A (LF 0.85), index 1 is Asset B (LF 1.05), both $1k. `wantedCollateralValue` is smaller for B (higher LF in the denominator), but A gets hit first because of index order. So governance's `addAsset` order effectively decides which collateral the borrower loses.

Is list order intended to be part of the spec here? If yes, worth a line in SPEC.md + a comment above `absorbInternal` so integrators know. If not, sorting `assetsIn` by LF desc before the loop would make it deterministic on economics instead of history.

siposova-304 (commented):
https://revolut.me/r/eZ37KPjLkY

siposova-304:
Balance tranfer account https://revolut.me/r/eZ37KPjLkY

PR #1113, opened 2026-04-22
Update rsETH price feeds on WETH and wstETH markets on Mainnet

MishaShWoof:


deployments/base/aero/migrations/1761125221_upgrade_to_capo_price_feeds.ts
```diff
@@ -0,0 +1,160 @@
+import { expect } from 'chai';
```
torreyatcitty, line 1: We only need the migrations for the 2 new price feeds correct?

PR #1140, opened 2026-07-11
fix(spider): normalize address keys to lowercase for case-insensitive lookups

### LINKED ISSUE #765: Addresses in `relations.ts` should be case-insensitive
Spider currently expects addresses in `relations.ts` to be all lower case. We should make it case-insensitive.

e.g.
```
  '0xDbcd5BafBAA8c1B326f14EC0c8B125DB57A5cC4c': {
    artifact: 'contracts/ERC20.sol:ERC20',
  },
  ```
will not work right now

dev-khimesh:
## Description

The `crawl()` function looks up addresses in the relation config map via `relations[address.toLowerCase()]`, but the map keys may be checksummed (e.g. `0xAbCd...`) instead of lowercase (`0xabcd...`). When a config author uses a checksummed address as a key, the spider silently misses the contract configuration for that address.

## Fix

Adds `normalizeRelationKeys()` that lowercases only Ethereum address keys (matching `/^0x[a-fA-F0-9]{40}$/`) at `spider()` entry. Non-address keys like contract names (`contracts/Dog.sol:Dog`) and alias templates (`comet`) are preserved as-is — those lookups use the original casing and would break if lowercased unconditionally.

## Test Plan

- [ ] Existing spider tests should pass (address lookups remain case-insensitive as they were already using `address.toLowerCase()`)
- [ ] Configs with checksummed address keys should now resolve correctly
- [ ] Configs with lowercase address keys continue to work (no regression)
- [ ] Configs with non-address keys (contract names, alias templates) are unaffected

Closes #765


dev-khimesh:
This PR also fixes issue #765 (addresses case-insensitivity in relations.ts).

